Issue #5, opened 2021-03-31
Better names for jobs


spirali:
- [x] Add argument ``--name=XXX`` to ``hq submit`` that allows to set an arbitrary name
- [x] Make checks that name is not extra long or does not contain newlines, etc that may break outputs
- [x] When `hq submit /a/b/c.sh` is submited then make default name only "c.sh"

spirali:
Implemented in #29

PR #29, opened 2021-05-10
Addition of --name argument for hq submit

vyomkeshj:
Fixes: #5


src/bin/hq.rs
```diff
@@ -135,6 +135,8 @@ struct JobDetailOpts {
 #[derive(Clap)]
 #[clap(setting = clap::AppSettings::ColoredHelp)]
 struct SubmitOpts {
+    #[clap(long)]
```
Kobzol, line 138: Please add help for the option. It's enough to add a doc comment above the attribute (see other `*Opts` structs.

src/client/commands/submit.rs
```diff
@@ -2,24 +2,35 @@ use tako::messages::common::ProgramDefinition;
...
+                    PathBuf::from(t)
+                        .file_name()
+                        .and_then(|t| t.to_str().map(|s| s.to_string()))
+                })
+                .unwrap_or_else(|| "job".to_string())
+        }
+
+        Some(name) => {
+            validate_and_sanitize_name(name)?
+        }
```
Kobzol, line 31: `Some(name) => validate_and_sanitize_name(name)?`

src/client/commands/submit.rs
```diff
@@ -37,3 +48,13 @@ pub async fn submit_computation(
     print_job_detail(gsettings, response.job);
     Ok(())
 }
+
+ fn validate_and_sanitize_name(mut name:String) ->anyhow::Result<String, Error> {
+     if name.contains('\n') || name.contains('\t'){
+         error(format!("name cannot have a newline or a tab"))
```
Kobzol, line 54: You can also just use the `anyhow::anyhow!` macro.

Kobzol (changes requested):
Looks great :) There is an unresolved import of an error.

When there are upstream changes (to master), better rebase instead of merging, to make the commit history more readable (we would squash the commits before merge anyway, but it's easier for review).

src/client/commands/submit.rs
```diff
@@ -2,34 +2,33 @@ use tako::messages::common::ProgramDefinition;
...
-                    PathBuf::from(t)
-                        .file_name()
-                        .and_then(|t| t.to_str().map(|s| s.to_string()))
-                })
-                .unwrap_or_else(|| "job".to_string())
-        }
 
-        Some(name) => {
-            validate_and_sanitize_name(name)?
+    let name = name.map(|name| {
```
Kobzol, line 19: I was also thinking about how to make this part of the code nicer, but I think that it was actually OK before. It is defnitely better to let the user know that there was an error in a proper way, instead of panicking (`unwrap`). So I suggest to return to the previous version with the `match`.

Kobzol (changes requested):

tests/test_job.py
```diff
@@ -45,6 +45,25 @@ def test_job_submit(hq_env: HqEnv):
...
+
+    hq_env.command(["submit", "sleep", "1", "--name=sleep_prog"])
+    time.sleep(0.2)
+    table = hq_env.command("jobs", as_table=True)
+    assert len(table) == 2
+    assert table[1][:3] == ["1", "sleep_prog", "WAITING"]
+    failed = False
+    try:
+        hq_env.command(["submit", "sleep", "1", "--name=second_sleep \t"])
+        hq_env.command(["submit", "sleep", "1", "--name=second_sleep \n"])
```
Kobzol, line 59: This should be put into another `try`/`except` block, otherwise the second command won't ever be executed.

Kobzol (changes requested):

tests/test_job.py
vyomkeshj: I was waiting for a comment, I didn't like how having two try blocks looked 🙃

vyomkeshj:
yes, it just looked nicer. I think I should have added a newline after the last match =>.
Is it okay?

Kobzol:
Just run `cargo fmt` and you don't have to think about these things :)

spirali:
In general definitely run cargo fmt. But please do not run it in this particular case because of I need to do some heavy rebasing.

spirali:
Rebased and merged as 7539b60844bc809cb2d289199214745606cada73

Thank you

PR #32, opened 2021-05-17
Support of array jobs

spirali:
Note: Formatting was postponed to avoid conflicts in other branch.


src/server/client.rs
```diff
@@ -189,66 +202,77 @@ async fn handle_job_cancel(
...
+            JobType::Array(a) => a.task_count(),
+        };
+        let tako_base_id = state.new_task_id(task_count);
+        let mut tako_task_id = tako_base_id;
 
-        let body = rmp_serde::to_vec_named(&program_def).unwrap();
-        let job = Job::new(task_id, message.name.clone(), program_def.clone());
+        let task_defs = match &message.job_type {
+            JobType::Simple => vec![make_task(job_id, 0, tako_task_id)],
+            JobType::Array(a) => a.iter().map(|tt_id| {
```
Kobzol, line 261:
```rust
a.iter().zip(0..a.task_count()).map(|(tt_id, index)|
```
but that's probably less readable.

src/server/job.rs
```diff
@@ -1,53 +1,211 @@
...
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct JobTaskInfo {
+    pub state: JobTaskState,
+    pub task_id: JobTaskId,
+}
+
+pub enum JobState {
+    SingleTask(JobTaskState),
```
Kobzol, line 24: Why is it needed to distinguish these two states? Couldn't `SingleTask` just be a `Map` of size one?

src/transfer/messages.rs
```diff
@@ -24,7 +33,13 @@ pub struct CancelRequest {
 pub struct JobInfoRequest {
     // If None then all jobs are turned
     pub job_ids: Option<Vec<JobId>>,
-    pub include_program_def: bool,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct JobDetailRequest {
+    // If None then all jobs are turned
```
Kobzol, line 40: This comment is not relevant anymore.

Kobzol (approved):

src/common/arrayparser.rs
```diff
@@ -0,0 +1,59 @@
...
+use nom::IResult;
+use nom::error::VerboseError;
+use nom::character::complete::digit1;
+use nom::combinator::all_consuming;
+use crate::common::arraydef::{TaskIdRange, ArrayDef};
+use nom::bytes::complete::tag;
+use nom::sequence::{preceded, tuple};
+use anyhow::anyhow;
+
+pub type NomResult<'a, Ret> = IResult<&'a str, Ret, VerboseError<&'a str>>;
```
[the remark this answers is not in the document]
spirali: Of course for two numbers it is overkill. The syntax will be more complex in the future. I wanted to make it easier for another "onboard" tasks, so I already started with a more robust solution

src/server/job.rs
spirali: This is kind of premature optimization. I expect that some part of users will submit lots of jobs (instead of submiting job arrays) so I wanted to avoid allocation (and I did not find any smallvec alternatives to map).

src/transfer/messages.rs
spirali: Good catch. Fixed

src/common/arrayparser.rs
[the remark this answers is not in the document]
Kobzol: Good idea :)

src/client/commands/submit.rs
```diff
@@ -4,36 +4,57 @@ use crate::client::globalsettings::GlobalSettings;
...
-        })
-        .unwrap_or_else(|| "job".to_string());
+                .unwrap_or_else(|| "job".to_string());
+
+    let mut args = opts.args;
+    args.insert(0, opts.command);
+
+    let job_type = opts.array.map(JobType::Array).unwrap_or(JobType::Simple);
+
+    let submit_cwd = std::env::current_dir().unwrap();
```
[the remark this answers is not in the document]
spirali: I had the feeling that when this fails, then something is really wring and it is probably better just quit right away :)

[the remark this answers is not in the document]
Kobzol: Sure :D A bigger question is where should the stdout/stderr live. But that's not for this PR.

src/server/client.rs
spirali: You are right, just it has to be modified to different start, as it is not counted from 0, but from tako_base_id

Btw: There is also .enumerate() so you suggestion can be also even more simplified.

Kobzol: Nice, I forgot about that.

Issue #37, opened 2021-05-18
Filter online/offline workers


spirali:
Command ``hq workers list`` shows all workers. Add option ``--online`` to show only online workers.

Kobzol:
Implemented in https://github.com/It4innovations/hyperqueue/pull/38

PR #38, opened 2021-05-21
Filter online offline workers

fredykoci:
Fixes: https://github.com/It4innovations/hyperqueue/issues/37

### DIFF
--- a/src/bin/hq.rs
+++ b/src/bin/hq.rs
@@ -221,13 +221,13 @@ async fn command_worker_stop(
 }
 fn be_printed(worker:&WorkerInfo,option:&WorkerListOpts) -> bool{
     let mut output_state = false;
-    if(option.common.running){
-        if (worker.ended_at == None) {
+    if option.common.running{
+        if worker.ended_at == None {
             output_state = true;
         }
     }
-    if(option.common.offline){
-        if (worker.ended_at != None) {
+    if option.common.offline{
+        if worker.ended_at != None {
             output_state = true;
         }
     }
--- a/src/bin/hq.rs
+++ b/src/bin/hq.rs
@@ -239,22 +239,13 @@ async fn command_worker_list(
 ) -> anyhow::Result<()> {
     let mut connection = get_client_connection(&gsettings.server_directory()).await?;
     let mut workers = get_worker_list(&mut connection, &gsettings)
-        .await.unwrap();
-    let mut print_workers = Vec::new();
+        .await?;
+
     if _opts.common.running != false || _opts.common.offline != false{
-        for worker in workers{
-            if be_printed(&worker,&_opts){
-                print_workers.push(worker)
-            }
-        }
-    }
-    else{
-        print_workers = workers;
+        workers.retain(|x| be_printed(&x,&_opts));
     }
 
-
-
-    print_worker_info(print_workers, &gsettings);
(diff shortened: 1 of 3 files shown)


src/bin/hq.rs
```diff
@@ -201,15 +219,38 @@ async fn command_worker_stop(
...
-        .map_err(|e| e.into())
+    let mut workers = get_worker_list(&mut connection, &gsettings)
+        .await.unwrap();
+    let mut print_workers = Vec::new();
+
+    for worker in workers{
+        if worker_state(&worker,&_opts){
+            print_workers.push(worker)
+        }
+    }
```
Kobzol, line 249:
```rust
let workers_to_print: Vec<_> = workers
    .iter()
    .filter(|worker| worker_state(worker, &opts)
    .collect();
```

src/bin/hq.rs
```diff
@@ -201,15 +219,38 @@ async fn command_worker_stop(
         .await
         .map_err(|e| e.into())
 }
-
+fn worker_state(worker:&WorkerInfo,option:&WorkerListOpts) -> bool{
```
Kobzol, line 222: I would rename this to something like `should_be_printed`.

Kobzol (changes requested):
With this implementation, if the user passes neither `running` nor `offline`, no workers will be shown.
Maybe it would be better to only implement a single flag (`--online`) with the following behaviour:
1) If nothing is passed, all workers will be displayed
2) If `--online` is passed, only online workers will be displayed

src/bin/hq.rs
```diff
@@ -201,15 +219,43 @@ async fn command_worker_stop(
...
 async fn command_worker_list(
     gsettings: GlobalSettings,
     _opts: WorkerListOpts,
 ) -> anyhow::Result<()> {
     let mut connection = get_client_connection(&gsettings.server_directory()).await?;
-    get_worker_list(&mut connection, &gsettings)
-        .await
-        .map_err(|e| e.into())
+    let mut workers = get_worker_list(&mut connection, &gsettings)
+        .await.unwrap();
```
spirali, line 242: Do not use unwrap but propagate the error (via ``?`` operator)

src/client/commands/worker.rs
```diff
@@ -1,14 +1,13 @@
...
 use crate::transfer::connection::ClientConnection;
-use crate::transfer::messages::{FromClientMessage, StopWorkerMessage, ToClientMessage};
+use crate::transfer::messages::{FromClientMessage, StopWorkerMessage, ToClientMessage, WorkerInfo};
 use crate::WorkerId;
 
 pub async fn get_worker_list(
     connection: &mut ClientConnection,
     gsettings: &GlobalSettings,
-) -> crate::Result<()> {
+) -> crate::Result<(Vec<WorkerInfo>)> {
```
spirali, line 10: Superfluous parenthesis

spirali (changes requested):

spirali:
Please include also Python integration tests in your code (directory "tests").

spirali:
I have rebased and squashed your commits as c9e6523d6806a36ef461db5efa1c5f5d23db719b.
I have also added a small refactoring 7144980369a5a747b029ff5710e304dab59e719e, if you have any question to my changes, do not hesitate to ask.

Thank you for your work.

Issue #34, opened 2021-05-17
Fail policy


spirali:
In the current version, when a task in task array failed, the computation continues. Implement fail policy as follows:

``hq submit --array=1-10 --max-fails=X``

Cancel remaining task in a job when there are more failed tasks then X. If --max-fails is not defined then use the current behaior.

PR #41, opened 2021-06-07
Isssue #34 

fredykoci:


src/client/commands/submit.rs
```diff
@@ -12,6 +12,10 @@ use crate::transfer::connection::ClientConnection;
 use crate::transfer::messages::{FromClientMessage, JobType, SubmitRequest, ToClientMessage};
 use anyhow::anyhow;
 
+pub struct SubmitCommon {
```
spirali, line 15: This is not needed

src/client/commands/submit.rs
```diff
@@ -59,6 +68,7 @@ pub async fn submit_computation(
             stderr: Some(stderr),
             cwd: None,
         },
+        fail_no: opts.max_fails.unwrap(),
```
spirali, line 71: * Please name the attributed and variables that hold this value as "max_fails"
* Do not call "unwrap" on data that came from the user. User should never cause "panic".
* If max fails is not defined, it should behave as the current version

src/server/job.rs
```diff
@@ -45,13 +46,14 @@ impl JobTaskCounters {
 pub struct Job {
     pub job_id: JobId,
     pub base_task_id: TakoTaskId,
-
+    pub n_to_cancel_job: u32,
```
spirali, line 49: Name it please max_fails

src/server/state.rs
```diff
@@ -109,8 +109,19 @@ impl State {
...
+        if job.counters.n_failed_tasks >= job.n_to_cancel_job {
+            let test = job.make_job_detail(true);
+            for x in test.tasks.iter() {
+                let (_, state) = job.get_task_state_mut(x.task_id.into());
+                if matches!(state ,JobTaskState::Running) || matches!(state, JobTaskState::Waiting) {
+                    job.set_cancel_state(x.task_id.into());
+                }
+            }
+        } else {
+            job.set_failed_state(msg.id, msg.info.message);
```
spirali, line 123: * ``job.set_failed_state(msg.id, msg.info.message);`` should be always called even max_fail was reached

spirali (changes requested):

src/client/commands/submit.rs
```diff
@@ -23,6 +24,10 @@ pub struct SubmitOpts {
 
     #[clap(long)]
     name: Option<String>,
+
+    #[clap(long, requires = "array")]
```
spirali, line 28: Why is there requires = "array"?

src/server/state.rs
```diff
@@ -40,6 +42,39 @@ pub type StateRef = WrappedRcRefCell<State>;
...
+    job_id: JobId,
+    tasks: Vec<TakoTaskId>,
+) {
+    if tasks.is_empty() {
+        return;
+    }
+    let tako_ref = tako_ref.clone();
+    let state_ref = state_ref.clone();
+    tokio::task::spawn_local(async move {
+        let message = FromGatewayMessage::CancelTasks(CancelTasks { tasks });
```
Kobzol, line 57: `rpc_call!`?

spirali: ``rpc_call`` is for client messages, this is a hq <-> tako message.

Kobzol: Ah. So, should we introduce `tako_call!`? :)

spirali: Good idea, but for another PR:)

PR #43, opened 2021-06-08
CPU management for HQ

spirali:


src/client/commands/submit.rs
```diff
@@ -6,11 +6,26 @@ use tako::messages::common::ProgramDefinition;
...
+use crate::client::resources::parse_cpu_request;
 use crate::common::arraydef::ArrayDef;
 use crate::rpc_call;
 use crate::transfer::connection::ClientConnection;
 use crate::transfer::messages::{FromClientMessage, JobType, SubmitRequest, ToClientMessage};
 use anyhow::anyhow;
+use std::str::FromStr;
+use tako::common::resources::{CpuRequest, ResourceRequest};
+
+// I am wrapping CpuRequest + implementing FromStr as I am not able to provide
```
Kobzol, line 18: I also didn't find any other way currently.

src/client/resources.rs
```diff
@@ -0,0 +1,78 @@
...
+            tuple((
+                p_uint,
+                opt(preceded(
+                    multispace1,
+                    alt((tag("compact!"), tag("compact"), tag("scatter"))),
+                )),
+            )),
+            |(count, policy)| {
+                if count == 0 {
+                    return Err(nom::Err::Error("xxx"));
```
Kobzol, line 23: xxx? :)

src/worker/start.rs
```diff
@@ -57,6 +61,21 @@ pub struct WorkerStartOpts {
     manager: ManagerOpts,
 }
 
+fn launcher_setup(task: &Task, def: LauncherDefinition) -> tako::Result<ProgramDefinition> {
+    let allocation = task.resource_allocation().unwrap();
```
Kobzol, line 65: `?`

Kobzol (approved):

src/client/resources.rs
spirali: Oops :)

src/worker/start.rs
spirali: launcher setup should be executed only on a running task and such that it has an allocation. It is a fatal internal error if allocation is not there, so I decided for unwrap.

Kobzol: Ok. I suggest to always use `expect` in such cases so that we have documentation about why it should not fail.

PR #45, opened 2021-06-09
CPU detection

spirali:


Kobzol (approved):

src/worker/hwdetect.rs
```diff
@@ -1,9 +1,70 @@
...
+            anyhow::bail!("Invalid format of range");
+        }
+        Ok((ids[0].parse()?, ids[1].parse()?))
+    } else {
+        let id: CpuId = line.parse()?;
+        Ok((id, id))
+    }
+}
+
+pub fn read_linux_numa() -> anyhow::Result<Vec<Vec<CpuId>>> {
```
Kobzol, line 17: Maybe we could create a type alias for `Vec<Vec<CpuId>>`? It's used in multiple places and it took me a while to "parse" what it represents.

Kobzol (approved):

src/worker/hwdetect.rs
spirali: Good point. I have added alias in Tako and I will chage this with the next update of Tako

PR #51, opened 2021-06-15
--host configuration for server start

spirali:


src/server/bootstrap.rs
```diff
@@ -73,9 +76,11 @@ async fn get_server_status(server_directory: &Path) -> crate::Result<ServerStatu
 }
 
 async fn initialize_server(
-    server_directory: &Path,
+    gsettings: &GlobalSettings,
     end_flag: Arc<Notify>,
+    host: Option<String>,
 ) -> anyhow::Result<impl Future<Output = anyhow::Result<()>>> {
+    let server_directory = gsettings.server_directory();
     let client_listener = TcpListener::bind("0.0.0.0:0")
```
Kobzol, line 84: Maybe the host should be used here instead of `0.0.0.0`? The user might want to make HQ only accessible from the given hostname. But I'm not sure if it wouldn't break some networking scenarios on e.g. IT4I clusters.

Kobzol (approved):

src/server/bootstrap.rs
spirali: As far I understand it, bind has to be IP address, not hostname. (Hostname is property of machine, but bind wants to select a a network interface). To expose it, we should probably expose some additional --bind parameter. (Just to distinguish that I want to set host (=visible name) to "login123" and "0.0.0.0" for bind).

Issue #24, opened 2021-05-01
More detailed program specification


spirali:
* [x] Specify working directory of task through ``hq submit --cwd=/a/path ...``
* [x] Specify specify stdout/stderr files ``hq submit --stdout=/a/path/file ...`` / ``hq submit --stderr=/a/path/file ...``
* [x] Substitution of "{JOB_ID}" to actual job id in paths ``hq submit --stdout=/a/path/file-{JOB_ID} ...``
* [x] Allow to specify extra env variables: ``hq submit --env PYTHONPATH=xxx --env MYVAR=yyy ...``
* [ ] Inheriting all current env variables ``hq submit --inherit-env ...``

Kobzol:
Oh, sorry if you wanted to work on env. variables @vyomkeshj :laughing:. I assumed that the assignment was only for working directory.

PR #53, opened 2021-06-16
#24, added --cwd

vyomkeshj:
Tests coming up

### DIFF
--- a/src/client/commands/submit.rs
+++ b/src/client/commands/submit.rs
@@ -73,6 +73,10 @@ pub struct SubmitOpts {
     #[clap(long)]
     pin: bool,
 
+    //Working dir for the submitted job, defaults to the current dir
+    #[clap(long)]
+    cwd: Option<PathBuf>,
+
     /// Path where the standard output of the job will be stored
     #[clap(long)]
     stdout: Option<String>,
@@ -145,7 +149,17 @@ pub async fn submit_computation(
         .collect();
     args.insert(0, opts.command.into());
 
-    let submit_cwd = std::env::current_dir().unwrap();
+    let submit_cwd = opts.cwd.unwrap_or_else(|| std::env::current_dir().unwrap());
+    if !fs::metadata(&submit_cwd)
+        .map(|m| m.is_dir())
+        .unwrap_or(false)
+    {
+        log::warn!(
+            "{:?} is not a valid directory on the current node",
+            submit_cwd
+        )
+    }
+
     let stdout = if opts.stdout.as_deref() == Some("none") {
         None
     } else {
@@ -186,13 +200,12 @@ pub async fn submit_computation(
     let message = FromClientMessage::Submit(SubmitRequest {
         job_type,
         name,
-        submit_cwd,
         spec: ProgramDefinition {
             args,
             env,
             stdout,
             stderr,
-            cwd: None,
+            cwd: Some(submit_cwd),
         },
         resources,
         pin: opts.pin,
--- a/tests/test_job.py
+++ b/tests/test_job.py
@@ -79,6 +79,21 @@ def test_custom_name(hq_env: HqEnv, tmp_path):
     assert len(table) == 2
 
 
+def test_custom_working_dir(hq_env: HqEnv, tmp_path):
+    offset = 10
+    hq_env.start_server()
+    cwd_custom_tbl = hq_env.command(
+        [
+            "submit",
+            "sleep 1",
+            "--cwd=/test/directory/",
+        ], as_table=True
+    )
+    time.sleep(0.2)
+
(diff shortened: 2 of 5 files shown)


src/client/commands/submit.rs
```diff
@@ -157,6 +163,15 @@ fn validate_name(name: String) -> anyhow::Result<String, anyhow::Error> {
     }
 }
 
+fn get_validated_path(cwd: String) -> anyhow::Result<PathBuf, anyhow::Error> {
+    let working_dir = PathBuf::from(cwd);
+    match fs::metadata(&working_dir) {
+        Ok(cwd) if cwd.is_dir() => Ok(working_dir),
+        Err(error)  => Err(anyhow!("Error setting cwd, {}", error)),
```
Kobzol, line 170: I would remove this line and just keep the fallback ("The path supplied is not a directory").

vyomkeshj: I was thinking about possible errors due to permissions issue, removed

Kobzol: We can't solve all the issues for the user, let him also do some debugging too :D

Btw @spirali it might happen that the passed cwd is not valid on the node of the submitter, but it will be valid on the node of the worker. Do we want to support this case?

spirali: Definitely, it should be not be an error that directory does not exists in submitter machine (as it may exists only on remote side). Or at least it should not be a hard error and only a warning. It is probably not a problem now, but it will be a problem in cloud partition.

Kobzol: Let's keep it as a warning with a message like `<cwd> is not a valid directory on the current node.`.

src/client/commands/submit.rs
```diff
@@ -157,6 +161,25 @@ fn validate_name(name: String) -> anyhow::Result<String, anyhow::Error> {
...
+                Ok(cwd) if cwd.is_dir() => working_dir_path,
+                _ => {
+                    warn!(
+                        "{} is not a valid directory on the current node",
+                        working_dir
+                    );
+                    working_dir_path
+                }
+            }
+        }
```
Kobzol, line 178: (just FYI, you don't have to change it). This could be written a bit shorter with `if let`:
```rust
if let Err(_) = fs::metadata(&working_dir) {
    warn!(
        "{} is not a valid directory on the current node",
        working_dir
    );
}
working_dir.into()
```
It's a shorter alternative if you only really need one branch of a `match`.

vyomkeshj: Thank you :D I'll change it because this is exactly what I've been obsessing over :D

src/client/commands/submit.rs
```diff
@@ -45,6 +46,9 @@ pub struct SubmitOpts {
     #[clap(long)]
     name: Option<String>,
 
+    #[clap(long)]
+    cwd: Option<String>,
```
spirali, line 50: It should be not String but PathBuf directly (I am sorry, I gave you a bad example in stdout/stderr, that should be changed too)

tests/test_job.py
```diff
@@ -75,6 +75,20 @@ def test_custom_name(hq_env: HqEnv, tmp_path):
...
+    cwd_custom_tbl = hq_env.command(
+        [
+            "submit",
+            "sleep 1",
+            "--cwd=/test/directory/",
+        ], as_table=True
+    )
+    time.sleep(0.2)
+
+    assert (cwd_custom_tbl[9][1] == "/test/directory/")
```
vyomkeshj, line 89: I am not sure how to test the default case here without the --cwd, how do I execute echo $PWD?

spirali: submit command ``bash -c "echo $PWD"``

Kobzol: In terminal, you can run ` ./target/debug/hq submit --cwd /tmp 'echo $PWD'`. This will not unroll `$PWD`. In tests it should not be unrolled automatically I think.

BTW, when I run e.g. ` ./target/debug/hq submit --cwd /tmp echo hello`, the job fails and nothing gets printed to stdout/stderr. Can you please check what is happening?

src/client/commands/submit.rs
vyomkeshj: I'm gonna change it all :D

Kobzol:
FYI: to resolve conflicts, you should rebase over `main` and then force-push (merging `main` won't help). It would be a good idea to squash first, too :)
Something like:
```
# squash
$ git reset --soft `git merge-base origin/main main`
$ git commit -m "Add --cwd"
# rebase
$ git rebase origin/main
# force push
$ git push --force
```

Kobzol:
Thanks @vyomkeshj :)

PR #54, opened 2021-06-16
hq worker info <worker-id>

User3574:
Closes: #22

### DIFF
--- a/src/bin/hq.rs
+++ b/src/bin/hq.rs
@@ -7,7 +7,7 @@ use cli_table::ColorChoice;
 use hyperqueue::client::commands::jobs::{cancel_job, output_job_detail, output_job_list};
 use hyperqueue::client::commands::stop::stop_server;
 use hyperqueue::client::commands::submit::{submit_computation, SubmitOpts};
-use hyperqueue::client::commands::worker::{get_worker_list, stop_worker};
+use hyperqueue::client::commands::worker::{get_worker_info, get_worker_list, stop_worker};
 use hyperqueue::client::globalsettings::GlobalSettings;
 use hyperqueue::client::job::Status;
 use hyperqueue::client::worker::print_worker_info;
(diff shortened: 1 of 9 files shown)


src/client/worker.rs
```diff
@@ -45,3 +46,52 @@ pub fn print_worker_info(workers: Vec<WorkerInfo>, gsettings: &GlobalSettings) {
         ]);
     assert!(print_stdout(table).is_ok());
 }
+
+pub fn print_worker_details(worker: WorkerDetail, gsettings: &GlobalSettings) {
```
spirali, line 50: Please do not duplicate print_worker_configuration. If you need to modify it, try to refactor by introducing a new function that is shared by both functions. However as it seems it may be just a one function now

src/client/worker.rs
```diff
@@ -1,8 +1,9 @@
 use cli_table::format::Justify;
-use cli_table::{print_stdout, Cell, Color, Style, Table};
+use cli_table::{print_stdout, Cell, Color, Style, Table, Row};
```
spirali, line 2: Please remove all warnings.

src/bin/hq.rs
```diff
@@ -123,6 +129,8 @@ enum WorkerCommand {
     List(WorkerListOpts),
     /// Hwdetect
     Hwdetect,
+    /// Display details about worker
+    Details(WorkerDetailsOpts),
```
spirali, line 133: I would still prefer "hq worker info <xxx>" as it is more compatible with the rest of the terminology. However, if there is a good reason why call it "detail", lets discuss it.

spirali (changes requested):

Kobzol:
See here: https://github.com/It4innovations/hyperqueue/pull/53#issuecomment-864843008 on how to merge changes from `main` into PR branches (rebase + force push).

spirali:
@User3574  Thank you

Issue #49, opened 2021-06-15
Progresbar


spirali:
Show the following information in a form progress bar:

![image](https://user-images.githubusercontent.com/263349/122027108-1c319a00-cdcb-11eb-9d64-2911d657c1fd.png)

PR #60, opened 2021-06-17
Add colored progress bar that shows progress of job's tasks

Kobzol:
Showcase (using `watch --color`):
progress-bar

I looked at existing progress bar libraries and I haven't found any that would allow me to output the progress bar to a string in a simple way. And for this we probably don't need a special dependency anyway. I included `colored` to color the bar though, maybe we can use `termcolor`, which is already used by `cli_table`.

Closes: https://github.com/It4innovations/hyperqueue/issues/49

### DIFF
--- a/src/bin/hq.rs
+++ b/src/bin/hq.rs
@@ -4,7 +4,7 @@ use std::str::FromStr;
 use clap::{Clap, ValueHint};
 use cli_table::ColorChoice;
 
-use hyperqueue::client::commands::jobs::{cancel_job, get_job_detail, get_job_list};
+use hyperqueue::client::commands::jobs::{cancel_job, output_job_detail, output_job_list};
 use hyperqueue::client::commands::stop::stop_server;
 use hyperqueue::client::commands::submit::{submit_computation, SubmitOpts};
 use hyperqueue::client::commands::worker::{get_worker_list, stop_worker};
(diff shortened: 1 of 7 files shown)


src/client/job.rs
```diff
@@ -66,6 +68,40 @@ fn status_cell(status: Status) -> CellStruct {
     }
 }
 
+/// Draws a colored progress bar that depicts counts of tasks with individual states
+fn job_progress_bar(info: &JobInfo) -> String {
+    let mut buffer = String::from("[");
+
+    let width: usize = 40;
+    let parts = vec![
```
spirali, line 76: * Do you need to construct a Vec, isnt just an array sufficient?
* Color settings should respect gsettings.color_policy

src/client/job.rs
```diff
@@ -76,7 +112,7 @@ fn job_status_with_counts_cells(info: &JobInfo) -> Vec<CellStruct> {
             );
         }
     };
-    let mut result: Vec<CellStruct> = Vec::new();
+    let mut result: Vec<CellStruct> = vec![job_progress_bar(info).cell()];
```
spirali, line 115: Maybe disable progress when it is not a task array? (I do not know, just thinking)

src/client/job.rs
Kobzol: - Array will be possible to use here from Rust 1.53, which will be released.. in about an hour :)
- Color settings are set [here](https://github.com/It4innovations/hyperqueue/pull/60/files#diff-68808a0be41327ce142404d37bbf8f747794caa49909229052f8fbf9a284df09R284).

src/client/job.rs
Kobzol: I would keep it also for jobs with a single task, for consistency and fixed table layout (mainly for tests).

src/bin/hq.rs
```diff
@@ -281,12 +281,21 @@ fn make_global_settings(opts: CommonOpts) -> GlobalSettings {
     GlobalSettings::new(server_dir, color_policy)
 }
 
+fn set_colored_settings(settings: &GlobalSettings) {
```
spirali, line 284: Maybe pattern match directly on ColorPolicy? (opts.colors?)

spirali (approved):

src/bin/hq.rs
Kobzol: I would keep it like this, you can in theory create global settings out of any input interface, and it's our own struct :)

spirali: But ColorChoice is not our enum and your "_" may fail if they add something into it. But it is just an micro issue:)

Kobzol: Oh, I see. Well, then I think `GlobalSettings` should contain `ColorPolicy` instead of `ColorChoice`, and `ColorChoice` should use `#[non_exhaustive]` :)

Anyway, thanks to `_` it will compile even if `ColorChoice` changes.

spirali: I meant "fail" in semantics. But actually as we create the value ourselves so it is ok. However, you are right, in general, gsettings should probably hold our structure and create ColorChoice on-fly.

PR #64, opened 2021-06-18
Idle timeout

spirali:


src/bin/hq.rs
```diff
@@ -70,6 +71,9 @@ struct ServerStartOpts {
     /// Hostname/IP of the machine under which is visible to others, default: hostname
     #[clap(long)]
     host: Option<String>,
+
+    #[clap(long)]
+    idle_timeout: Option<ArgDuration>,
```
Kobzol, line 76: Hmm, I would expect this to be a parameter of the worker, not the server :thinking:

tests/test_worker.py
```diff
@@ -101,3 +100,18 @@ def test_worker_list_resources(hq_env: HqEnv):
...
     assert table[1][3] == "1x10 cpus"
     assert table[2][3] == "4x5 cpus"
+
+
+def test_idle_timeout(hq_env: HqEnv):
+    hq_env.start_server(args=["--idle-timeout", "1s"])
+    w = hq_env.start_worker(args=["--heartbeat", "500ms"])
+    time.sleep(0.5)
+    hq_env.command(["submit", "--", "sleep", "1"])
+    time.sleep(1.0)
```
Kobzol, line 110: `wait_for_job_state`?

Kobzol (approved):

src/bin/hq.rs
spirali: I was thinking about it as kind of "session" parameter where session is a single start of server. But maybe we can have both server setting (that provides a default) and worker setting that may override the default

tests/test_worker.py
spirali: I am testing here that worker is not turned off too soon. So putting here an wait_for_job state could be placed here, but still some sleep is required

Kobzol: Yeah just the `time.sleep(1.0)` should be replaced with `wait_for_job_state(hq_env, 1, "FINISHED")`.

Issue #42, opened 2021-06-07
Handle a large ammont of stdout/stderr files


spirali:
In case of array jobs with many tasks produce many stdout/stderr file on shared fs. Introduce an option that writes these files into a ramdisk (or /tmp) and only if a task fails its stdout/stderr are written to a final destination

Kobzol:
When should the generated files be removed? It's unclear whether following jobs/taks will need to use them. They will not be accessible from another node anyway.

Kobzol:
Fixed by https://github.com/It4innovations/hyperqueue/pull/100

PR #65, opened 2021-06-18
HQ script for backing up stdout/stderr files

Kobzol:
The name `script` is bikesheddable, of course :)

Closes: https://github.com/It4innovations/hyperqueue/issues/42


src/client/commands/submit.rs
```diff
@@ -183,6 +184,13 @@ pub async fn submit_computation(
         )
     }
 
+    if let Some(ref stdout) = stdout {
```
Kobzol, line 187: This is incorrect, these environment variables need to be created by Tako, not HQ, because they can contain placeholders. @spirali right?

spirali: Yes, if we allow to insert dynamic placeholdres (like DATE / WORKDIR) then it have to be done in launcher setup

Issue #57, opened 2021-06-17
Implement new variables in --stdout/--stderr


spirali:
Implement variables in -stdout/--stderr in defined in #55 + ``%{CWD}`` that contains the working directory of job.

Issue #55, opened 2021-06-17
Variables in --cwd definitions and --create-cwd


spirali:
Implement replacing the following strings in path provided in --cwd

- ``%{JOB_ID}%``
- ``%{TASK_ID}%``
- ``%{SUBMIT_DIR}%``
- ``%{DATE}``
- ``%{SERVER_UUID}`` -- contact @spirali for this

- Implement --create-cwd that is equivalent to --cwd but creates a directory if it exists (it will create the directory on worker!).
(add a boolean flag in submit definitions that cwd should created)

Kobzol:
Just a note: working directory should be resolved with placeholders first, then it can be used as a placeholder in stdout/stderr.
e.g.
cwd: `/home/user/jobs/%{DATE}` -> turns into
cwd: `/home/user/jobs/123456789`

stdout: `%{CWD}/stdout` -> turns into
stdout: `/home/user/jobs/123456789/stdout`

PR #67, opened 2021-06-21
Replace placeholders in stderr, stdout and cwd

Kobzol:
@spirali Let's change the defaults after https://github.com/It4innovations/hyperqueue/issues/55 is implemented (I'll work on it next).

Closes: https://github.com/It4innovations/hyperqueue/issues/55
Closes: https://github.com/It4innovations/hyperqueue/issues/57


src/client/commands/submit.rs
```diff
@@ -149,7 +167,7 @@ pub async fn submit_computation(
         .collect();
     args.insert(0, opts.command.into());
 
-    let submit_cwd = opts.cwd.unwrap_or_else(|| std::env::current_dir().unwrap());
+    let submit_cwd = opts.cwd;
```
spirali, line 170: This should be resolve "." to absolute path. Or is there still a discussion what should be default cwd?

Kobzol: I think that it should be resolved in the worker, not the submitter, but this is just a temporary placeholder so that tests pass. Let's discuss the defaults after we properly implement placeholder replacing during submit.

spirali: Ok, but then lets change at least the default stdout to %{SUBMIT_DIR}/stdout....

Kobzol: There's no point in doing that until we replace SUBMIT_DIR :) Actually, I will implement the replacement in this PR, there's no reason to divide the PRs.

spirali (approved):

spirali:
LGTM :+1:, please update user documentation.

PR #72, opened 2021-06-23
Set job state to waiting if there are still uncomputed tasks and no task is running

Kobzol:


spirali (commented):
LGTM, please update the user doc

PR #78, opened 2021-06-30
Get PBS job ETA time issue 

fredykoci:


src/worker/pbs_info.rs
```diff
@@ -0,0 +1,53 @@
+use std::process::Command;
+use std::{str, env};
+use regex::Regex;
+use std::time::Duration;
+
+type TimeNumberFormat = u64;
+
+fn get_value_from_qstat_string(command_output: &String, property_name: &mut String, data_pattern: &String) -> String {
```
Kobzol, line 8: What is the format of the `qstat` output? Please add a documentation comment to the function so that we know how the parser behaves and what should its input/output look like.

src/worker/pbs_info.rs
```diff
@@ -0,0 +1,53 @@
...
+        let mut walltime = re.find(&command_output).unwrap();
+        re = Regex::new(data_pattern.as_str()).ok().unwrap();
+        walltime = re.find(walltime.as_str()).unwrap();
+        String::from(walltime.as_str())
+    } else {
+        return String::new();
+    }
+}
+
+fn process_time_format(raw_time: String) -> TimeNumberFormat {
```
Kobzol, line 21: Why not return `Duration` directly?

src/worker/pbs_info.rs
```diff
@@ -0,0 +1,53 @@
...
+        _ => String::new(),
+    };
+    let job_id = if job_id.is_empty() == false { job_id.as_str() } else { panic!("no jobID"); };
+
+    let command_output = match Command::new("qstat")
+        .args(&["-f", job_id])
+        .output() {
+        Result::Ok(x) => match String::from_utf8(x.stdout) {
+            Ok(x) => x,
+            Err(e) => panic!("{}", e),
```
Kobzol, line 42: We definitely do not want to panic here, the function should return `Result<Duration>` and handle errors properly.

Kobzol (changes requested):

Kobzol:
Could you try to use the JSON output (`qstat -f -F json`) instead of using regexes? You can use `serde` to map the output of PBS to some structure, to avoid manual parsing of the JSON output.

Kobzol:
The branch was not properly rebased, there are still extra commits from `main`. You can remove them e.g. using an interactive rebase.
```
$ git rebase -i main
```
Interactive rebase lets you change history of your branch interactively. Run it and then use `d` (drop) command for the commits before `77db5ff` (function returning PBS ...). Then exit the editor, which will finish the rebase process. There should be no conflicts.

After that, you can force push.

fredykoci:
> The branch was not properly rebased, there are still extra commits from `main` and also some merge commits. You can remove them e.g. using an interactive rebase.
> 
> ```
> $ git rebase -i main
> ```
> 
> Interactive rebase lets you change history of your branch interactively. Run it and then use `d` (drop) command for the commits before `77db5ff` (function returning PBS ...). Then exit the editor, which will finish the rebase process. There should be no conflicts.
> 
> After that, you can force push.

now you have to have same main history

Cargo.toml
```diff
@@ -17,6 +17,7 @@ opt-level = 3
 panic = "abort"
 
 [dependencies]
+regex = "1"
```
Kobzol, line 20: This is not longer needed.

src/worker/pbs_info.rs
```diff
@@ -0,0 +1,52 @@
+use std::process::Command;
+use std::{str, env};
+use std::time::Duration;
+use serde_json::Value;
+
+type TimeNumberFormat = u64;
+
+
+fn process_time_format(raw_time: String) -> TimeNumberFormat {
```
Kobzol, line 9: You can just receive `&str` instead of `String`. `String` is an owned type which allocates on the heap, it shouldn't be needed here.

src/worker/pbs_info.rs
```diff
@@ -0,0 +1,52 @@
...
+    let used = process_time_format(String::from(data_json["Jobs"][job_id]["resources_used"]["walltime"].as_str().unwrap()));
+
+    Ok(Duration::from_secs(walltime - used))
+}
+
+
+fn get_remaining_time() -> serde_json::Result<Duration> {
+    let job_id = match env::var("PBS_JOBID") {
+        Result::Ok(x) => x,
+        _ => String::new(),
```
Kobzol, line 32: Using an empty job ID is not a good idea, as `qstat` will then return all of the jobs. We should probably instead return something like `serde_json::Result<Option<Duration>>` from the function and if pbs job id is missing, return `None`. Or the job id can be a parameter of this function, since it should only be called within PBS context anyway.

src/worker/pbs_info.rs
```diff
@@ -0,0 +1,52 @@
...
+    let data_json: Value = serde_json::from_str(data.as_str())?;
+
+    let walltime = process_time_format(String::from(data_json["Jobs"][job_id]["Resource_List"]["walltime"].as_str().unwrap()));
+    let used = process_time_format(String::from(data_json["Jobs"][job_id]["resources_used"]["walltime"].as_str().unwrap()));
+
+    Ok(Duration::from_secs(walltime - used))
+}
+
+
+fn get_remaining_time() -> serde_json::Result<Duration> {
```
Kobzol, line 29: You can return `anyhow::Result` so that you can replace the panics with `?`. For now it's probably enough.

src/worker/start.rs
```diff
@@ -20,7 +20,9 @@ use crate::common::timeutils::ArgDuration;
 use crate::worker::hwdetect::detect_resource;
 use crate::worker::output::print_worker_configuration;
 use crate::worker::parser::parse_cpu_definition;
+use crate::worker::pbs_info::get_valltime;
```
Kobzol, line 23: This function no longer exists.

Kobzol (changes requested):
You have accidentally removed `Cargo.lock`, restore it please :)

Kobzol:
I refactored the code in preparation for self-contained PBS/Slurm modules, which will be also needed for automatic job allocation. I also added basic infrastructure for mocking external programs, so that we can test the PBS manager which now uses `qstat`.

tests/utils/mock.py
```diff
@@ -0,0 +1,32 @@
...
+class ProgramMock:
+    def __init__(self, directory: str):
+        self.directory = str(os.path.abspath(directory))
+        os.makedirs(self.directory, exist_ok=True)
+
+    def update_env(self, env: Dict[str, str]):
+        path = ""
+        if "PATH" in env:
+            path = f"{env['PATH']}:"
+        path += self.directory
```
spirali, line 17: Isnt it better to put mocked directory at the beginning of the path? What is qsub is actually installed on machine?

src/common/manager/pbs.rs
```diff
@@ -0,0 +1,40 @@
...
+    let seconds: u64 = numbers[0].parse::<u64>().unwrap() * 3600
+        + numbers[1].parse::<u64>().unwrap() * 60
+        + numbers[2].parse::<u64>().unwrap();
+    Duration::from_secs(seconds)
+}
+
+fn get_time(job_id: &str, data: &str) -> Option<Duration> {
+    let data_json: Value = serde_json::from_str(data).ok()?;
+
+    let walltime = parse_duration(data_json["Jobs"][job_id]["Resource_List"]["walltime"].as_str()?);
```
spirali, line 17: Is this is panic-free if Jobs is not in the directory?

spirali (changes requested):

src/common/manager/pbs.rs
```diff
@@ -0,0 +1,40 @@
...
+use std::process::Command;
+use std::str;
+use std::time::Duration;
+
+fn parse_duration(raw_time: &str) -> Duration {
+    let numbers: Vec<&str> = raw_time.split(":").collect();
+    let seconds: u64 = numbers[0].parse::<u64>().unwrap() * 3600
+        + numbers[1].parse::<u64>().unwrap() * 60
+        + numbers[2].parse::<u64>().unwrap();
+    Duration::from_secs(seconds)
```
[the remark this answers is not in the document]
Kobzol: There is no `Duration::from_hours` :(

src/worker/start.rs
```diff
@@ -345,13 +346,18 @@ fn try_get_pbs_info() -> anyhow::Result<Map<String, String>> {
...
 
-    let mut result = Map::with_capacity(2);
+    let mut result = Map::with_capacity(3);
     result.insert("MANAGER".to_string(), "PBS".to_string());
-    result.insert("MANAGER_JOB_ID".to_string(), manager_job_id);
-
-    // TODO: Run "qstat -f -F json $PBS_JOBID" to get walltime
+    result.insert("MANAGER_JOB_ID".to_string(), manager_job_id.clone());
+    result.insert(
+        "MANAGER_WALLTIME".to_string(),
```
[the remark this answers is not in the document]
Kobzol: Ok. I would also like to propose to create a typed structure for manager information and not put it into an opaque dictionary.

src/common/manager/pbs.rs
Kobzol: Yes, it will just return `Value::Null`.

src/common/manager/pbs.rs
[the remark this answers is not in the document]
spirali: You can use https://docs.rs/chrono/0.4.19/chrono/struct.Duration.html, HQ is already using chrono library. It has conversion from minutes/hours

src/worker/start.rs
```diff
@@ -372,30 +370,29 @@ fn try_get_slurm_info() -> anyhow::Result<Map<String, String>> {
...
+            try_get_pbs_info().or_else(|_| try_get_slurm_info()).ok()
         }
         ManagerOpts::None => {
             log::debug!("Manager detection disabled");
-            Ok(Map::new())
+            None
         }
-        ManagerOpts::Pbs => try_get_pbs_info(),
-        ManagerOpts::Slurm => try_get_slurm_info(),
+        ManagerOpts::Pbs => Some(try_get_pbs_info().expect("Failed to get PBS info")),
```
spirali, line 394: No panic please

spirali (changes requested):

src/worker/start.rs
Kobzol: But the program should still crash, right?

spirali: It should terminated with an error.

Kobzol:
I squashed the changes and rebased them over `main`. I think that we can merge this PR now to avoid it getting too big.

Issue #36, opened 2021-05-17
Complex array definitions


spirali:
Implement more complex array definitions:

The following should work:

``hq submit --array=1,2,7-10,25-40:5,110 ...``

That is equivalent to run array on ids: 1,2,7,8,9,10, 25, 30, 35, 40, 110

* Include a test that all ranges are non-overlapping.

Kobzol:
Related to https://github.com/It4innovations/hyperqueue/issues/52.

Issue #33, opened 2021-05-17
``hq resubmit`` command


spirali:
Implement ``hq resubmit <job_id>`` then resubmits again a previous non-running job as a new job.

``hq resubmit <job_id> --failed`` resubmits only failed tasks from the original job
``hq resubmit <job_id> --canceled`` resubmits only canceled tasks from the original job

PR #79, opened 2021-07-02
hq resubmit

User3574:
Fixes: #33 
Fixes: #36

### DIFF
--- a/src/common/arrayparser.rs
+++ b/src/common/arrayparser.rs
@@ -1,31 +1,58 @@
 use anyhow::anyhow;
 use nom::bytes::complete::tag;
 use nom::combinator::all_consuming;
-use nom::combinator::{map, map_res, opt};
+use nom::combinator::{map_res, opt};
+use nom::multi::separated_list1;
 use nom::sequence::{preceded, tuple};
 
 use crate::common::arraydef::{ArrayDef, TaskIdRange};
-use crate::common::parser::{format_parse_error, p_uint, NomResult};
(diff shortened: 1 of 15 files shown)


src/common/arrayparser.rs
```diff
@@ -1,25 +1,46 @@
...
             _ => Err("Invalid range"),
         },
     )(input)
 }
 
 fn p_array_def(input: &str) -> NomResult<ArrayDef> {
-    map(p_task_id_range, ArrayDef::new)(input)
+    map_res(p_task_id_ranges, |r| match r {
+        res if !is_overlapping(res.clone()) => Ok(ArrayDef::new(res)),
+        _ => Err("Ranges overlap"),
```
Kobzol, line 31: @spirali Why should overlapping ranges cause an error? Maybe a warning could be printed, but it seems to me that an error is too strict.

spirali: I prefer the error. I do not see a usecase.

Kobzol: It seems to me that overlap is totally harmless here.

It's an artificial example, but if I want to e.g. resubmit every even task and also tasks from some range, I could just do
`0-1000:2, 500-600`. With this check I would need to split the intervals manually.

spirali:
I have did some squashing and small refactoring. It is ready to be merged. Can you please (@Kobzol @User3574) rebase it on main? I am getting some conflicts wrt. to Nom error messages)

Kobzol:
@User3574 Please check of the rebased version is correct.

PR #85, opened 2021-07-08
store worker id into job Running state 

fredykoci:


fredykoci:
require my tako merge request

src/server/state.rs
```diff
@@ -170,7 +170,7 @@ impl State {
         log::debug!("Task id={} updated {:?}", msg.id, msg.state);
         let job = self.get_job_mut_by_tako_task_id(msg.id).unwrap();
         match msg.state {
-            TaskState::Running => job.set_running_state(msg.id),
+            TaskState::Running(worker) => job.set_running_state(msg.id,worker),
```
vyomkeshj, line 173: Let's keep it worker_id just for clarity :)

vyomkeshj: plus running cargo_fmt should automatically add the missing whitespace after the comma

src/client/job.rs
```diff
@@ -57,7 +57,7 @@ pub fn job_status(info: &JobInfo) -> Status {
 fn task_status(status: &JobTaskState) -> Status {
     match status {
         JobTaskState::Waiting => Status::Waiting,
-        JobTaskState::Running => Status::Running,
+        JobTaskState::Running(_) => Status::Running,
```
vyomkeshj, line 60: @Kobzol Don't we want to propagate it to status too within this PR?

vyomkeshj: I wonder if there is a possible test here where we propagate this to the job_list table, create two tasks, spawn two workers,  see if the tasks are running on different workers

Kobzol: Yeah this PR is not too big, we can just directly propagate and display the worker ID in the corresponding tables.

src/server/job.rs
```diff
@@ -2,14 +2,16 @@ use serde::{Deserialize, Serialize};
 use tako::messages::common::ProgramDefinition;
 
 use crate::transfer::messages::{JobDetail, JobInfo, JobType};
-use crate::{JobId, JobTaskCount, JobTaskId, Map, TakoTaskId};
+use crate::{JobId, JobTaskCount, JobTaskId, Map, TakoTaskId, WorkerId};
 use tako::common::resources::ResourceRequest;
 
+//type WorkerId = u64;
```
Kobzol, line 8: This can be deleted.

src/server/job.rs
```diff
@@ -206,45 +209,58 @@ impl Job {
...
-        self.counters.n_running_tasks -= 1;
-        self.counters.n_finished_tasks += 1;
+
+        match state {
+            JobTaskState::Running(worker_id) => {
+                *state = JobTaskState::Finished(*worker_id);
+                self.counters.n_running_tasks -= 1;
+                self.counters.n_finished_tasks += 1;
+            }
+            _ => assert!(matches!(state, JobTaskState::Running(_))),
```
Kobzol, line 229: `panic!("Invalid worker state, expected Running, got {:?}", state)`

The same in set_failed_state.

src/server/job.rs
```diff
@@ -206,45 +209,58 @@ impl Job {
...
-        *state = JobTaskState::Failed(error);
-        self.counters.n_running_tasks -= 1;
-        self.counters.n_failed_tasks += 1;
+
+        match state {
+            JobTaskState::Running(worker_id) => {
+                let mut my_error = error.to_owned();
+                my_error.push_str(" on Worker ");
+                my_error.push_str(worker_id.to_string().as_str());
+                *state = JobTaskState::Failed(my_error);
```
Kobzol, line 248: `let my_error = format!("{} on Worker {}", error, worker_id);`

src/client/job.rs
```diff
@@ -57,20 +57,39 @@ pub fn job_status(info: &JobInfo) -> Status {
...
+}
+
+fn status_cell(status: Status, task_state: Option<&JobTaskState>) -> CellStruct {
+    match task_state {
+        Some(state) => match state {
+            JobTaskState::Waiting => "WAITING".cell().foreground_color(Some(Color::Cyan)),
+            JobTaskState::Finished(worker_id) => {
+                worker_state_messages("FINNISHED ON ", worker_id, Color::Green)
+            }
+            JobTaskState::Failed(error) => "FAILED".cell().foreground_color(Some(Color::Red)),
```
Kobzol, line 80: `JobTaskState::Failed(_)`

src/client/job.rs
```diff
@@ -57,20 +57,39 @@ pub fn job_status(info: &JobInfo) -> Status {
...
+    message.push_str(worker_id.to_string().as_str());
+    message.cell().foreground_color(Some(color))
+}
+
+fn status_cell(status: Status, task_state: Option<&JobTaskState>) -> CellStruct {
+    match task_state {
+        Some(state) => match state {
+            JobTaskState::Waiting => "WAITING".cell().foreground_color(Some(Color::Cyan)),
+            JobTaskState::Finished(worker_id) => {
+                worker_state_messages("FINNISHED ON ", worker_id, Color::Green)
```
Kobzol, line 78: You can just use the `format!` macro.

`format!("FINISHED ON {}", worker_id)`

Also I would switch the matches (first match `state`, then inside choose specific message based on task_state, to avoid duplicating the colors.

Kobzol (changes requested):

src/client/job.rs
```diff
@@ -57,20 +57,47 @@ pub fn job_status(info: &JobInfo) -> Status {
...
+        StatusColors::Failed => Color::Red,
+        StatusColors::Running => Color::Green,
+        StatusColors::Canceled => Color::Magenta
+    }
+}
+
+fn status_cell(status: Status, task_state: Option<&JobTaskState>) -> CellStruct {
+    match task_state {
+        Some(state) => match state {
+            JobTaskState::Waiting => "WAITING".cell().foreground_color(Some(get_color(StatusColors::Waiting))),
```
spirali, line 88: * Status should be always only one of the keywords "RUNNING", "CANCELED", without any extra data (i.e. *not* RUNNING ON ...). 
Information about the worker should be another row/column in the table. There will be more information about worker in the future (e.g. CPU allocations) and we cannot have a super long status.

* If the task failed I also want to know on which worker the task was running.

spirali (changes requested):

Kobzol:
What's left to do:
- Display information about the worker in the job/task table (include it as a separate row)
- If there are too many workers (e.g. more than 4), display only some of them and then show `…`
- Display also the hostname of the workers, instead of just ID. The hostnames of workers should be fetched from the server.

tests/test_job.py
```diff
@@ -500,3 +500,22 @@ def test_job_resubmit_all(hq_env: HqEnv):
...
+def test_job_tasks_table(hq_env:HqEnv):
+    hq_env.start_server()
+    hq_env.command(["submit", "echo", "test"])
+    table = hq_env.command(["job", "1" , "--tasks"],as_table=True)
+    assert table[11][1] == "WAITING"
+    assert table[11][2] == ""
+    hq_env.start_worker()
+    table = hq_env.command(["job", "1" , "--tasks"],as_table=True)
+    assert table[11][1] == "FINISHED"
+    assert table[11][2] != ""
```
Kobzol, line 513: You can get the hostname in Python using e.g. `socket.gethostname()`.

tests/test_job.py
```diff
@@ -500,3 +500,22 @@ def test_job_resubmit_all(hq_env: HqEnv):
...
     assert table[3] == ["Tasks", "3; Ids: 2, 7, 9"]
+
+def test_job_tasks_table(hq_env:HqEnv):
+    hq_env.start_server()
+    hq_env.command(["submit", "echo", "test"])
+    table = hq_env.command(["job", "1" , "--tasks"],as_table=True)
+    assert table[11][1] == "WAITING"
+    assert table[11][2] == ""
+    hq_env.start_worker()
+    table = hq_env.command(["job", "1" , "--tasks"],as_table=True)
```
Kobzol, line 511: Use `wait_for_job_state` to make sure that the test will not fail if the task execution is too slow.

src/client/commands/submit.rs
```diff
@@ -206,7 +206,7 @@ pub async fn submit_computation(
         submit_dir: std::env::current_dir().unwrap().to_str().unwrap().into(),
     });
     let response = rpc_call!(connection, message, ToClientMessage::SubmitResponse(r) => r).await?;
-    print_job_detail(gsettings, response.job, true, true);
+    print_job_detail(gsettings, response.job, true, false);
```
Kobzol, line 209: Why was this changed to `false`?

Kobzol (approved):

src/client/commands/submit.rs
fredykoci: it were in older commits

Kobzol:
Thanks :)

PR #86, opened 2021-07-08
Naive cancel implementation without tests

User3574:


src/bin/hq.rs
```diff
@@ -238,6 +239,7 @@ async fn command_job_detail(gsettings: GlobalSettings, opts: JobDetailOpts) -> a
                 }
             }
         }
+        JobSpecifier::All => return Err(Error::msg("Specifier all is not implemented for job details, did you mean: job list?"))
```
Kobzol, line 242: It's enough to use `bail!("Error message")` or `return Err(anyhow!("Error message"))`.

src/client/commands/jobs.rs
```diff
@@ -62,28 +75,33 @@ pub async fn output_job_detail(
 pub async fn cancel_job(
     _gsettings: &GlobalSettings,
     connection: &mut ClientConnection,
-    job_id: JobId,
+    job_ids: Vec<JobId>,
 ) -> crate::Result<()> {
-    let response =
-        rpc_call!(connection, FromClientMessage::Cancel(CancelRequest { job_id }), ToClientMessage::CancelJobResponse(r) => r).await?;
+    let mut sorted_ids = job_ids.clone();
```
Kobzol, line 80: Why do we need to sort the IDs? If we do, I think that there's no need to clone them.

src/transfer/messages.rs
```diff
@@ -68,8 +68,8 @@ pub enum FromClientMessage {
 
 #[derive(Serialize, Deserialize, Debug)]
 pub enum CancelJobResponse {
-    Canceled(Vec<JobTaskId>, JobTaskCount),
-    InvalidJob,
+    Canceled(JobId, Vec<JobTaskId>, JobTaskCount),
```
Kobzol, line 71: Maybe the `JobId` isn't needed here if the order between `CancelRequest::job_ids` and `CancelJobResponse.0` stays the same?

src/transfer/messages.rs
```diff
@@ -28,7 +28,7 @@ pub struct SubmitRequest {
 
 #[derive(Serialize, Deserialize, Debug)]
 pub struct CancelRequest {
-    pub job_id: JobId,
+    pub job_ids: Vec<JobId>,
```
Kobzol, line 31: Please use `JobSelector` here (see `JobInfoRequest`). A function to convert `JobSpecifier` to `JobSelector` will also probably be needed. We can also rename `JobSpecifier` to `JobSelectorArg` or something like that.

spirali:
Looks good except last thing: It tries to cancel also jobs that is already fully finished / canceled. So it produce lots of unnecessary visual errors. "cancel all" should affect only jobs where is something to cancel

spirali:
(I have rebased and do some small changes, the last problem with canceling already canceled is still there)

src/transfer/messages.rs
```diff
@@ -36,7 +36,7 @@ pub struct CancelRequest {
 #[derive(Serialize, Deserialize, Debug)]
 pub enum JobSelector {
     All,
-    LastN(usize),
+    LastN(JobId),
```
Kobzol, line 39: It feels weird that this is `JobId` :)

spirali: Ok, we can alias is as JobCounts, but technically it should be the same type.

Kobzol: I don't think that it should be the same type necessarily, you just want N last jobs, which corresponds to `usize`. The fact that you implement this interface with subtraction of `JobId`s shouldn't affect the interface :)

src/server/client.rs
```diff
@@ -165,7 +166,16 @@ async fn handle_job_cancel(
     let mut responses: Vec<(JobId, CancelJobResponse)> = Vec::new();
 
     let job_ids: Vec<JobId> = match selector {
-        JobSelector::All => state.jobs().map(|j| j.job_id).collect(),
+        JobSelector::All => state
+            .jobs()
+            .map(|job| job.make_job_info())
+            .filter(|job_info| {
+                vec![Status::Running, Status::Waiting]
```
Kobzol, line 173: I would prefer a simpler check like

```rust
let status = job_status(&job_info);
status == Status::Running || status == Status::Waiting
```

Maybe the compiler could inline all this, but I think that making it simpler is also more readable :)

We could also introduce something like

```rust
impl Status {
    fn is_active(&self) -> bool {
         matches!(self, Self::Running | Self::Waiting)
    }
}
```

src/transfer/messages.rs
spirali: I strongly disagree. usize has platform dependent-size, and you definitely do not want to have a protocol which shape depends on a platform. "Please send 32b if your server is running on platform X or 64b if on platform Y, btw: you do not know what platform is you peer" is worse then Dask protocol:)

Kobzol: Oh, ok, I mainly meant "an unsigned integer", `usize` is indeed impractical. So, let's say `u64`. Still, there should be no relation to `JobId` :)

src/server/client.rs
spirali: I agree, creating vec over and over is too much. matches! is much better

Issue #83, opened 2021-07-03
Command ``hq worker stop all``


spirali:
Stop all workers (without stopping the server)

PR #88, opened 2021-07-15
hq worker stop all

User3574:
Fixes: #83


src/client/commands/worker.rs
```diff
@@ -47,8 +48,23 @@ pub async fn get_worker_info(
...
-    rpc_call!(connection, message, ToClientMessage::StopWorkerResponse).await
+    let message = FromClientMessage::StopWorker(StopWorkerMessage { selector });
+    let mut responses =
+        rpc_call!(connection, message, ToClientMessage::StopWorkerResponse(r) => r).await?;
+
+    responses.sort_unstable_by_key(|x| x.0);
+    for (id, response) in responses {
+        match response {
+            StopWorkerResponse::Failed(e) => {
+                println!("Worker: {}, failed to stop, error: {}", id, e.to_string())
```
Kobzol, line 61: This should probably be returned as an error? But we would have to combine all of the error messages into a single `Result` to avoid losing data. Or at least use `eprintln` so that it is printed to stderr instead of stdout.

src/server/client.rs
```diff
@@ -114,19 +115,56 @@ pub async fn client_rpc_loop<
...
+            .map(|(_, worker)| worker.worker_id())
+            .collect(),
+    };
+
+    for worker_id in worker_ids {
+        if let Some(worker) = state.get_worker(worker_id) {
+            if worker.make_info().ended.is_some() {
+                responses.push((
+                    worker_id,
+                    StopWorkerResponse::Failed("worker is already stopped".to_string()),
```
Kobzol, line 140: IMO this shouldn't be an error. At least it should be renamed to something like `AlreadyStopped`.

src/server/client.rs
```diff
@@ -114,19 +115,56 @@ pub async fn client_rpc_loop<
...
+                ));
+            } else {
+                match tako_ref
+                    .send_message(FromGatewayMessage::StopWorker(StopWorkerRequest {
+                        worker_id,
+                    }))
+                    .await
+                    .unwrap()
```
Kobzol, line 148: Try to remove the `unwrap` and use results please.

tests/test_worker.py
```diff
@@ -58,6 +58,25 @@ def test_worker_stop(hq_env: HqEnv):
...
+
+
+def test_worker_stop_all(hq_env: HqEnv):
+    hq_env.start_server()
+    process_w_1, process_w_2 = hq_env.start_worker(), hq_env.start_worker()
+
+    wait_for_worker_state(hq_env, 1, "RUNNING")
+    wait_for_worker_state(hq_env, 2, "RUNNING")
+    hq_env.command(["worker", "stop", "all"])
+    wait_for_worker_state(hq_env, 1, "STOPPED")
```
Kobzol, line 74: You can also use `wait_for_worker_state(hq_env, [1, 2], "STOPPED")` to wait for multiple workers.

Kobzol (changes requested):

src/server/client.rs
User3574: What should the result message be in the case of Error?

src/client/commands/worker.rs
Kobzol: @spirali Maybe just logging to error would be enough here? To avoid complex combination of multiple errors into a single Result.

src/server/client.rs
Kobzol: You can just propagate the error from Tako using `?`. Maybe add some context to it (see `with_context` method), like `Unable to stop worker`.

src/bin/hq.rs
```diff
@@ -283,7 +299,13 @@ async fn command_worker_stop(
...
-    stop_worker(&mut connection, opts.worker_id)
+
+    let selector: WorkerSelector = match opts.selector {
+        WorkerSelectorArg::Id(job_id) => WorkerSelector::Specific(vec![job_id]),
+        WorkerSelectorArg::All => WorkerSelector::All,
+    };
+
+    stop_worker(&mut connection, selector)
         .await
         .map_err(|e| e.into())
```
Kobzol, line 310: `.await?` should be the same as `.await.map_err(|e| e.into())`. `?` call `into`.

src/bin/hq.rs
```diff
@@ -283,7 +299,13 @@ async fn command_worker_stop(
     opts: WorkerStopOpts,
 ) -> anyhow::Result<()> {
     let mut connection = get_client_connection(&gsettings.server_directory()).await?;
-    stop_worker(&mut connection, opts.worker_id)
+
+    let selector: WorkerSelector = match opts.selector {
```
Kobzol, line 303: Let's implement this as a `From` trait impl, to make it usable on more places.

```rust
impl From<WorkerSelectorArg> for WorkerSelector
```

src/client/commands/worker.rs
```diff
@@ -47,8 +48,29 @@ pub async fn get_worker_info(
...
+    for (id, response) in responses {
+        match response {
+            StopWorkerResponse::Failed(e) => {
+                log::error!("Stopping worker {} failed; {}", id, e.to_string());
+            }
+            StopWorkerResponse::InvalidWorker => {
+                log::error!("Stopping worker {} failed; worker not found", id);
+            }
+            StopWorkerResponse::AlreadyStopped => {
+                log::error!("Stopping worker {} failed; worker is already stopped", id);
```
Kobzol, line 67: I would display this as a warning, not an error.

src/client/commands/worker.rs
```diff
@@ -47,8 +48,29 @@ pub async fn get_worker_info(
...
+            }
+            StopWorkerResponse::Stopped => {
+                //log::info!("Worker {} stopped", id)
```
Kobzol, line 70: Let's uncomment this.

src/server/client.rs
```diff
@@ -114,19 +115,64 @@ pub async fn client_rpc_loop<
...
+                                responses.push((worker_id, StopWorkerResponse::Failed(error.message)))
+                            }
+                            msg => panic!(
+                                "Received invalid response to worker: {} stop: {:?}",
+                                worker_id, msg
+                            ),
+                        }
+                    }
+                    Err(err) => {
+                        panic!("Unable to stop worker: {} error: {:?}", worker_id, err)
```
Kobzol, line 166: I would use `responses.push((worker_id, StopWorkerResponse::Failed(error.message)))` for this case, maybe with an additional `log::error` log.

Kobzol (changes requested):

src/bin/hq.rs
```diff
@@ -283,9 +308,8 @@ async fn command_worker_stop(
     opts: WorkerStopOpts,
 ) -> anyhow::Result<()> {
     let mut connection = get_client_connection(&gsettings.server_directory()).await?;
-    stop_worker(&mut connection, opts.worker_id)
-        .await
-        .map_err(|e| e.into())
+    stop_worker(&mut connection, WorkerSelector::from(opts.selector)).await?;
```
Kobzol, line 311: It's more common to use `opts.selector.into()` than `WorkerSelector::from(opts.selector)`, but that's a detail.

src/client/commands/worker.rs
```diff
@@ -47,8 +48,29 @@ pub async fn get_worker_info(
...
+        rpc_call!(connection, message, ToClientMessage::StopWorkerResponse(r) => r).await?;
+
+    responses.sort_unstable_by_key(|x| x.0);
+    for (id, response) in responses {
+        match response {
+            StopWorkerResponse::Failed(e) => {
+                log::error!("Stopping worker {} failed; {}", id, e.to_string());
+            }
+            StopWorkerResponse::InvalidWorker => {
+                log::warn!("Stopping worker {} failed; worker not found", id);
```
Kobzol, line 64: This one should be an error in my opinion :D

PR #92, opened 2021-07-21
Parser for hwdetect

### LINKED ISSUE #90: Rewrite `worker::hwdetect::parse_range` using NOM

User3574:
Fixes #90

### DIFF
--- a/src/worker/hwdetect.rs
+++ b/src/worker/hwdetect.rs
@@ -1,24 +1,32 @@
+use crate::common::parser::{NomResult, p_uint, format_parse_error};
+use anyhow::anyhow;
+
 use tako::common::resources::{CpuId, NumOfCpus, ResourceDescriptor};
 
-pub fn parse_range(line: &str) -> anyhow::Result<Vec<CpuId>> {
-    let line = line.trim();
-    let mut result = Vec::new();
-    for part in line.split(',') {
-        let part = part.trim();
-        if part.contains('-') {
-            let ids: Vec<&str> = part.split('-').collect();
-            if ids.len() != 2 {
-                anyhow::bail!("Invalid format of range");
-            }
-            for i in ids[0].parse()?..=ids[1].parse()? {
-                result.push(i);
-            }
-        } else {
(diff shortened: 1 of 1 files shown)


Kobzol:
Looks good, but `all_consuming` is a problem, since the input can have whitespace. Either strip the whitespace beforehand or allow whitespace in NOM (the first should be simpler).

spirali:
The code does not cover all cases, newlines and "," delimiters. See tests.

src/worker/hwdetect.rs
```diff
@@ -1,24 +1,44 @@
...
+}
+
+fn p_cpu_ranges(input: &str) -> NomResult<Vec<CpuId>> {
+    separated_list1(tag(","), p_cpu_range)(input)
+        .map(|(a, b)| (a, b.into_iter().flatten().collect()))
+}
+
+pub fn parse_range(input: &str) -> anyhow::Result<Vec<CpuId>> {
+    let f_input = remove_whitespace(input);
+    let res = all_consuming(p_cpu_ranges)(&*f_input)
```
Kobzol, line 38: `f_input.as_str()`
and you can also remove the `let res` and just return the expression without `;`

Kobzol (approved):

src/worker/hwdetect.rs
```diff
@@ -1,24 +1,44 @@
...
-        }
-    }
-    Ok(result)
+use nom::bytes::complete::tag;
+use nom::combinator::all_consuming;
+use nom::combinator::{map_res, opt};
+use nom::multi::separated_list1;
+use nom::sequence::{preceded, tuple};
+
+fn remove_whitespace(s: &str) -> String {
```
spirali, line 12: Do not use this kind of method as it blindly removes spaces and makes the combinators uncomposable.
Use space1 from nom. Something like: terminated(xxx, space1) in places where space is allowed

User3574: Updated at my local repo [ https://github.com/User3574/hyperqueue/tree/hwdetect_parser ].

Issue #94, opened 2021-07-23
Submit ``--wait``


spirali:
Implement ``hq submit --wait ...``
that submits a job and then immediately waits for it. It should be equivalent to:

``hq submit ...``
``hq wait <JOB_ID_of_submit_above>``

(See #93)

spirali:
Open question: What should happen when process is terminated? (Probably the task should be canceled?)

Issue #93, opened 2021-07-23
``hq wait`` command


spirali:
Implement HQ command:

``hq wait <job_id``

that blocks waits until all tasks in job does not end in a final state (finished/failed/canceled).

spirali:
Open question: What should happen when process is terminated? (Probably the task should be canceled?)

PR #97, opened 2021-07-23
hq wait

User3574:
Fixes: #93
Fixes: #94

### DIFF
--- a/src/client/commands/wait.rs
+++ b/src/client/commands/wait.rs
@@ -12,9 +12,8 @@ pub async fn wait_on_job(
     connection: &mut ClientConnection,
     selector: JobSelector,
 ) -> anyhow::Result<()> {
-    let conn_ref = &mut *connection;
     let response = rpc_call!(
-        conn_ref,
+        connection,
         FromClientMessage::JobInfo(JobInfoRequest {
             selector,
         }),
@@ -37,10 +36,9 @@ pub async fn wait_on_job(
         let mut non_terminated_ids: Set<JobId> = job_ids.into_iter().collect();
 
         while !non_terminated_ids.is_empty() {
-            let conn_ref = &mut *connection;
             let ids_ref = &mut non_terminated_ids;
             let response = rpc_call!(
-                conn_ref,
+                connection,
                 FromClientMessage::JobInfo(JobInfoRequest {
                     selector: JobSelector::Specific(ids_ref.iter().copied().collect()),
                 }),
--- a/src/client/commands/wait.rs
+++ b/src/client/commands/wait.rs
@@ -5,8 +5,8 @@ use crate::transfer::connection::ClientConnection;
 use crate::transfer::messages::{FromClientMessage, JobInfoRequest, JobSelector, ToClientMessage};
 use crate::{rpc_call, JobId, Set};
 
-use std::thread::sleep;
 use std::time::Duration;
+use tokio::time::sleep;
 
 pub async fn wait_on_job(
     connection: &mut ClientConnection,
@@ -35,7 +35,7 @@ pub async fn wait_on_job(
         let bar = ProgressBar::new(job_ids.len() as u64);
         let mut non_terminated_ids: Set<JobId> = job_ids.into_iter().collect();
 
-        while !non_terminated_ids.is_empty() {
+        loop {
             let ids_ref = &mut non_terminated_ids;
             let response = rpc_call!(
                 connection,
@@ -52,11 +52,12 @@ pub async fn wait_on_job(
                     bar.inc(1);
                 }
             }
-
-            sleep(Duration::from_secs(1));
+            if non_terminated_ids.is_empty() {
+                break;
+            }
+            sleep(Duration::from_secs(1)).await;
         }
         bar.finish();
     }
-
     Ok(())
 }
--- a/src/client/commands/wait.rs
+++ b/src/client/commands/wait.rs
@@ -1,0 +1,63 @@
+use indicatif::ProgressBar;
+
+use crate::client::status::is_terminated;
+use crate::transfer::connection::ClientConnection;
+use crate::transfer::messages::{FromClientMessage, JobInfoRequest, JobSelector, ToClientMessage};
+use crate::{rpc_call, JobId, Set};
+
+use std::thread::sleep;
+use std::time::Duration;
+
+pub async fn wait_on_job(
+    mut connection: ClientConnection,
+    selector: JobSelector,
+) -> anyhow::Result<()> {
+    let conn_ref = &mut connection;
+    let response = rpc_call!(
+        conn_ref,
+        FromClientMessage::JobInfo(JobInfoRequest {
+            selector,
+        }),
+        ToClientMessage::JobInfoResponse(r) => r
+    )
+    .await?;
+    let job_ids: Vec<JobId> = response
+        .jobs
+        .iter()
+        .filter(|info| !is_terminated(info))
+        .map(|info| info.id)
+        .collect();
+
+    if job_ids.is_empty() {
+        log::warn!("There are no jobs to wait for");
+    } else {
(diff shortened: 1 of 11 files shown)


User3574:
Is there a way to replace  multiple get_client_connection calls inside loop by one connection call outside?

spirali:
I do not understand, you can create an connection and hold it for more requests

src/client/commands/wait.rs
```diff
@@ -0,0 +1,68 @@
...
+use crate::transfer::messages::{
+    FromClientMessage, JobInfo, JobInfoRequest, JobSelector, ToClientMessage,
+};
+use std::thread::sleep;
+use std::time::Duration;
+
+#[derive(Clap, Copy, Clone)]
+#[clap(setting = clap::AppSettings::ColoredHelp)]
+pub struct WaitOpts {
+    job_id: JobId,
```
Kobzol, line 18: I suppose that we can leverage job id range parsing and wait for a range of jobs?

src/client/commands/wait.rs
```diff
@@ -0,0 +1,68 @@
...
+}
+
+pub async fn wait_on_job(gsettings: &GlobalSettings, opts: WaitOpts) -> anyhow::Result<()> {
+    let mut response = rpc_call!(
+        get_client_connection(&gsettings.server_directory()).await?,
```
Kobzol, line 23:
```rust
```
and then just use `connection`.

src/client/commands/wait.rs
```diff
@@ -0,0 +1,68 @@
...
+        FromClientMessage::JobInfo(JobInfoRequest {
+            selector: JobSelector::Specific(vec![opts.job_id]),
+        }),
+        ToClientMessage::JobInfoResponse(r) => r
+    )
+    .await?;
+    let mut info: &JobInfo = response.jobs.get(0).unwrap();
```
Kobzol, line 30: First you should check if it's empty, otherwise the unwrap could crash the program.

src/client/commands/wait.rs
```diff
@@ -0,0 +1,75 @@
...
+            selector,
+        }),
+        ToClientMessage::JobInfoResponse(r) => r
+    )
+    .await?;
+    let job_ids: Vec<JobId> = response
+        .jobs
+        .iter()
+        .filter(|job_info| {
+            vec![Status::Running, Status::Waiting]
```
Kobzol, line 27: Please use `matches!` as we talked about before.

src/client/commands/wait.rs
```diff
@@ -0,0 +1,75 @@
...
+                    sleep(Duration::from_secs(1));
+                } else {
+                    is_running = false;
+                    log::error!("Job {} does not exist", job_id);
+                }
+            }
+            log::info!("Job {} has finished", job_id);
+
+            current_jobs += 1;
+            println!("Jobs done: {} / {}", current_jobs, total_jobs);
```
Kobzol, line 70: It would be nice to display some kind of progress bar :) Could you try to take a look at https://docs.rs/indicatif/0.16.2/indicatif/?

src/client/commands/wait.rs
```diff
@@ -0,0 +1,75 @@
...
+        .map(|info| info.id)
+        .collect();
+
+    if job_ids.is_empty() {
+        log::warn!("There are no jobs to wait for");
+    } else {
+        let mut current_jobs = 0;
+        let total_jobs = job_ids.len();
+
+        for job_id in job_ids {
```
Kobzol, line 40: Waiting one by one could in theory generate a lot of requests to the server, if there is a large number of jobs (i.e. million jobs = million requests).
@spirali is this OK? Or should the code instead request the status of a set of tasks at once?

spirali: Yes you are right, we should expect that user may wait on relatively a large number of jobs

src/client/commands/wait.rs
spirali: Maybe indicatif is too much, but the progress bar as you made in Job info would be fine

src/client/commands/wait.rs
Kobzol: That one was for a single use-case, I would rather not reinvent the wheel. But it's your call.

src/client/commands/wait.rs
```diff
@@ -0,0 +1,63 @@
...
+        .map(|info| info.id)
+        .collect();
+
+    if job_ids.is_empty() {
+        log::warn!("There are no jobs to wait for");
+    } else {
+        let bar = ProgressBar::new(job_ids.len() as u64);
+        let mut non_terminated_ids: Set<JobId> = job_ids.iter().copied().collect();
+        let mut is_empty: bool = false;
+        while !is_empty {
```
Kobzol, line 37:
```suggestion
        let mut non_terminated_ids: Set<JobId> = job_ids.into_iter().collect();
        while !non_terminated_ids.is_empty() {
```

src/client/commands/wait.rs
```diff
@@ -0,0 +1,63 @@
...
+    let job_ids: Vec<JobId> = response
+        .jobs
+        .iter()
+        .filter(|info| !is_terminated(info))
+        .map(|info| info.id)
+        .collect();
+
+    if job_ids.is_empty() {
+        log::warn!("There are no jobs to wait for");
+    } else {
```
Kobzol, line 33:
```suggestion
    } else {
         log::info!("Waiting for {} job(s)", job_ids.len());
```

Kobzol (approved):

Kobzol (approved):

src/client/commands/wait.rs
Kobzol: That one was for a single use-case, I would rather not reinvent the wheel. But it's your call.

tests/test_job.py
```diff
@@ -500,3 +500,20 @@ def test_job_resubmit_all(hq_env: HqEnv):
...
     table = hq_env.command(["resubmit", "1"], as_table=True)
     assert table[3] == ["Tasks", "3; Ids: 2, 7, 9"]
+
+
+def test_job_wait(hq_env: HqEnv):
+    hq_env.start_server()
+    hq_env.start_worker()
+    hq_env.command(["submit", "sleep", "1"])
+    r = hq_env.command(["wait", "1"])
+    assert "Waiting for 1 job(s)" in r
```
Kobzol, line 510: Please add a check that the job has actually finished after the wait command ended.

tests/test_job.py
```diff
@@ -500,3 +500,20 @@ def test_job_resubmit_all(hq_env: HqEnv):
...
+    r = hq_env.command(["wait", "all"])
+    assert "There are no jobs to wait for" in r
+
+
+def test_job_submit_wait(hq_env: HqEnv):
+    hq_env.start_server()
+    hq_env.start_worker()
+    r = hq_env.command(["submit", "sleep", "1", "--wait"])
```
Kobzol, line 518: The same here.

Kobzol (changes requested):

tests/test_job.py
```diff
@@ -500,3 +500,24 @@ def test_job_resubmit_all(hq_env: HqEnv):
...
+
+
+def test_job_wait(hq_env: HqEnv):
+    hq_env.start_server()
+    hq_env.start_worker()
+    hq_env.command(["submit", "sleep", "1"])
+    r = hq_env.command(["wait", "1"])
+    assert "Waiting for 1 job(s)" in r
+    table = hq_env.command(["job", "1"], as_table=True)
+    assert table[2][1] == "FINISHED"
```
Kobzol, line 512: There is a new API for reading the table outputs. Using numeric indices is annoying, since after each table change we have to change a lot of tests. Please use this: https://github.com/It4innovations/hyperqueue/blob/00d595f3659ae7ec326f214c8fc6797704c4dab8/tests/utils.py#L31 :innocent:

Kobzol (changes requested):

PR #98, opened 2021-07-26
Hyperqueue Dashboard skeleton

vyomkeshj:
Initial commits to start the discussion :)


src/client/commands/dashboard/top.rs
```diff
@@ -0,0 +1,100 @@
...
+                ("B8", 9),
+                ("B9", 14),
+                ("B10", 15),
+                ("B11", 1),
+            ],
+        }
+    }
+
+    fn update(&mut self) {
+        let value = self.data_vec.pop().unwrap();
```
Kobzol, line 41: What is this supposed to do?

Kobzol (commented):
The code doesn't compile, so I can't try the dashboard.

src/client/commands/dashboard/top.rs
vyomkeshj: It should compile now, the whole thing was in playground stage. I'll get it to work now that the tako update seems solidified

src/server/client.rs
```diff
@@ -102,6 +102,14 @@ pub async fn client_rpc_loop<
                         compute_job_detail(&state_ref, msg.job_id, msg.include_tasks)
                     }
                     FromClientMessage::Stats => compose_server_stats(&state_ref, &tako_ref).await,
+                    FromClientMessage::StartCliDashboard => {
```
Kobzol, line 105: I don't think that we want to run a dashboard loop on the server. It should run on the client instead. @spirali?

vyomkeshj: That was a mistake :) 
I've pushed my latest experiment where I was trying to get the data from the workers and bring it to the UI. It will be cleaned and done properly but I wanted to make something to start with.

Kobzol:
It seems to me that you have started to implement the dashboard "with bells and whistles" (seeing the layout code). This PR's getting quite big, do you think that we could first merge some very basic form that will just include the `dashboard` command and show some hello world screen?

vyomkeshj:
@Kobzol, yes. The plan is to just have one screen with a list of workers and their hw usage at the end of this PR. There is some unused code in this commit that I have inherited from kdash, I'll remove what I don't end up using. Is it okay?

vyomkeshj:
There are also a lot of unwanted changes in this PR right now that I'll remove

Kobzol:
> Is it okay?

Ok :) Let us know when it ready for review.

vyomkeshj:
https://user-images.githubusercontent.com/9003687/129660522-d3989b96-893e-404e-990c-1eac70bb85ff.mov

All workers are on the same machine in this one

Kobzol:
Looks pretty cool! :)

src/worker/start.rs
```diff
@@ -467,9 +467,9 @@ fn gather_configuration(opts: WorkerStartOpts) -> anyhow::Result<WorkerConfigura
         work_dir,
         log_dir,
         heartbeat_interval: opts.heartbeat.into_duration(),
+        hw_state_poll_interval: Some(Duration::from_millis(1000)),
```
Kobzol, line 470: Could you make this configurable from the command line (in another PR?). Just so that we don't forget about it.

src/transfer/messages.rs
```diff
@@ -86,6 +87,7 @@ pub enum FromClientMessage {
     WorkerInfo(WorkerInfoRequest),
     Stats,
     StopWorker(StopWorkerMessage),
+    GetCollectedOverview(OverviewRequest),
```
Kobzol, line 90: The name doesn't correspond to the other names. `Overview` should be enough.

src/bin/hq.rs
```diff
@@ -59,6 +60,10 @@ struct Opts {
 }
 
 #[allow(clippy::large_enum_variant)]
+#[derive(Clap)]
+#[clap(about = "HyperQueue Dashboard")]
```
Kobzol, line 64: You can just use a documentation comment (`///`).

src/dashboard/models.rs
```diff
@@ -0,0 +1,67 @@
+use tui::widgets::TableState;
+
+pub trait Scrollable {
```
Kobzol, line 3: Scrolling is currently not implemented, let's leave this for another PR.

src/dashboard/dashboard_state.rs
```diff
@@ -0,0 +1,26 @@
+pub struct DashboardState {
```
Kobzol, line 1: The files `state`, `manager` and `events` are already inside the `dashboard` module, so let's remove the `dashboard` prefix from them.

src/dashboard/dashboard_state.rs
```diff
@@ -0,0 +1,26 @@
...
+    pub config: Option<DashboardConfig>, //todo: this has to be some general structure for dashboard configuration.
+    pub ui_state: CurrentActiveUi,
+}
+
+/**
+ * What is currently being actively drawn on the dashboard
+ * fixme: Ui directly polls for data for now
+ */
+pub enum CurrentActiveUi {
```
Kobzol, line 10: Let's rename this to `DashboardScreen`.

src/dashboard/dashboard_state.rs
```diff
@@ -0,0 +1,26 @@
...
+    WorkerHwMonitorScreen,
+}
+
+#[derive(Debug, Default)]
+pub struct DashboardConfig {
```
Kobzol, line 15: Unused for now, let's remove it.

src/dashboard/ui/start.rs
```diff
@@ -0,0 +1,59 @@
...
+
+            DashboardEvent::ChangeUIStateEvent(_new_state) => {
+                //todo: change what is being drawn on the dashboard by changing the ui state!
+            }
+        }
+    }
+}
+
+async fn get_hw_overview(global_settings: &GlobalSettings) -> Result<CollectedOverview, HqError> {
+    let mut connection = get_client_connection(global_settings.server_directory()).await?;
```
Kobzol, line 48: Let's keep this connection persistent across the whole lifetime of the dashboard.

src/dashboard/ui/start.rs
```diff
@@ -0,0 +1,59 @@
...
+use crate::rpc_call;
+use crate::server::bootstrap::get_client_connection;
+use crate::transfer::messages::{FromClientMessage, ToClientMessage};
+
+pub async fn start_ui_loop(
+    painter: WrappedRcRefCell<DashboardPainter>,
+    event_handler: &DashboardEventHandler,
+    state: WrappedRcRefCell<DashboardState>,
+    global_settings: &GlobalSettings,
+) -> Result<(), anyhow::Error> {
```
Kobzol, line 19: `-> anyhow::Result<()>`

src/dashboard/ui/start.rs
```diff
@@ -0,0 +1,59 @@
...
+        match event_handler.receive_event()? {
+            DashboardEvent::KeyPressEvent(input) => {
+                if input == Key::Char('q') {
+                    // Quits the dashboard
+                    break Ok(());
+                }
+            }
+
+            DashboardEvent::Tick => {
+                let overview = get_hw_overview(global_settings).await?;
```
Kobzol, line 30: Only get the overview when the active screen is `WorkerHwMonitorScreen`.

src/dashboard/dashboard_manager.rs
```diff
@@ -0,0 +1,60 @@
...
+use crate::dashboard::dashboard_state::DashboardState;
+use crate::dashboard::ui::dashboard_ui::DashboardPainter;
+use crate::dashboard::ui::start::start_ui_loop;
+
+/// Allows drawing to the terminal
+pub type THandler = Terminal<TermionBackend<AlternateScreen<MouseTerminal<RawTerminal<Stdout>>>>>;
+
+/// The dashboard ui core structure
+pub struct DashboardManager {
+    pub dashboard_state: WrappedRcRefCell<DashboardState>,
```
Kobzol, line 21: This probably doesn't need to be `WrappedRcRefCell`.

src/dashboard/dashboard_manager.rs
```diff
@@ -0,0 +1,60 @@
...
+    global_settings: GlobalSettings,
+    terminal: WrappedRcRefCell<DashboardPainter>,
```
Kobzol, line 23: The same here.

src/dashboard/dashboard_manager.rs
```diff
@@ -0,0 +1,60 @@
...
+    event_handler: DashboardEventHandler,
+}
+
+impl DashboardManager {
+    /// Initialises with the connection to the terminal and provides an event_handler to
+    /// update the UI
+    pub fn new(global_settings: GlobalSettings) -> Result<Self, io::Error> {
```
Kobzol, line 30: Maybe the `DashboardManager` is useless for now? You can just pass these arguments directly to `start_ui_loop`.

src/dashboard/dashboard_events.rs
```diff
@@ -0,0 +1,76 @@
...
+    })
+}
+
+///Handles key press events when the dashboard_ui is active
+fn key_event_listener(tx: Sender<DashboardEvent>) -> thread::JoinHandle<()> {
+    thread::spawn(move || {
+        let stdin = io::stdin();
+        for evt in stdin.keys() {
+            if let Ok(key) = evt {
+                //todo: instead of sending KeyPressEvent, resolve here to the correct next state
```
Kobzol, line 66: I would handle it elsewhere, let's keep this simple and only make it send the events.

src/dashboard/dashboard_events.rs
```diff
@@ -0,0 +1,76 @@
...
+        if let Err(err) = tx.send(DashboardEvent::Tick) {
+            eprintln!("{}", err);
+            break;
+        }
+        thread::sleep(Duration::from_millis(250));
+    })
+}
+
+///Handles key press events when the dashboard_ui is active
+fn key_event_listener(tx: Sender<DashboardEvent>) -> thread::JoinHandle<()> {
```
Kobzol, line 61: Let's use `tokio::sync::mpsc::UnboundedSender` here. Otherwise the receiving side (which is inside a tokio loop) will block, which is wrong.

src/dashboard/dashboard_events.rs
```diff
@@ -0,0 +1,76 @@
...
+        self.receiver.recv()
+    }
+
+    pub fn send_ui_state_update_event(&self, event: DashboardEvent) {
+        self.sender.send(event); //todo: handle send error!
+    }
+}
+
+///Periodic updates to the dashboard
+fn provide_clock(tx: Sender<DashboardEvent>) -> thread::JoinHandle<()> {
```
Kobzol, line 50: Creating a thread for this is unnecessary. Let's just spawn a new tokio task that will generate the tick events, i.e. inside start_ui_loop.

Kobzol (changes requested):

src/dashboard/ui/start.rs
```diff
@@ -10,15 +10,31 @@ use crate::dashboard::ui::utils::get_hw_overview;
...
-pub async fn start_ui_loop(
+pub async fn start_ui_loop (
     painter: WrappedRcRefCell<DashboardPainter>,
-    event_handler: &mut DashboardEventHandler,
+    event_handler: WrappedRcRefCell<DashboardEventHandler>,
     state: WrappedRcRefCell<DashboardState>,
-    global_settings: &GlobalSettings,
+    global_settings: & GlobalSettings,
 ) -> Result<(), anyhow::Error> {
+
```
vyomkeshj, line 24: @Kobzol How do I share the event_handler between the two parts

src/dashboard/ui/screens/draw_utils.rs
```diff
@@ -0,0 +1,110 @@
...
+    layout::{Constraint, Direction, Layout, Rect},
+    style::{Color, Modifier, Style},
+    text::{Span, Spans, Text},
+    widgets::{Block, Borders, Paragraph, Row},
+    Frame,
+};
+
+use crate::dashboard::models::StatefulTable;
+
+const DARK_FG_COLOR: Color = Color::White;
```
Kobzol, line 13: Styles (light vs dark) are unused for now, let's remove them and just keep a single style.

Kobzol (changes requested):

src/dashboard/ui/start.rs
Kobzol: Try something like this:
```rust
let (tx, mut rx) = tokio::sync::mpsc::unbounded_channel();
let ticker = async move {
    let mut interval = tokio::time::interval(Duration::from_micros(500));
    loop {
        tx.send(DashboardEvent::Tick);
        interval.tick().await;
    }
};
let event_loop = async move {
    loop {
        match rx.recv().await {
            Some(dashboard_event) => {
                match dashboard_event {
                    DashboardEvent::KeyPressEvent(input) => {
                        if input == Key::Char('q') {
                            // Quits the dashboard
                            break Ok(());
                        }
                    }

                    DashboardEvent::Tick => {
                        let overview = get_hw_overview(global_settings).await?;

                        //Draw the correct dashboard ui according to the current ui state
                        match state.get().ui_state {
                            CurrentActiveUi::WorkerHwMonitorScreen => {
                                painter.get_mut().draw_dashboard_home(overview);
                            }
                        }
                    }

                    DashboardEvent::ChangeUIStateEvent(_new_state) => {
                        //todo: change what is being drawn on the dashboard by changing the ui state!
                    }
                }
            }

            None => (),
        }
    }
};
tokio::select! {
    _ = ticker => { Ok(()) }
    result = event_loop => { result }
}
```

src/worker/start.rs
vyomkeshj: I've taken a note, I'll follow up with a PR after this one

src/bin/hq.rs
```diff
@@ -406,11 +406,12 @@ async fn command_dashboard_start(
     gsettings: GlobalSettings,
     _opts: DashboardOpts,
 ) -> anyhow::Result<()> {
+    let connection = get_client_connection(gsettings.server_directory()).await?;
```
vyomkeshj, line 409: @Kobzol  Shouldn't this be mutable? How is it allowed when start_ui_loop asks for mutable connection reference

src/dashboard/ui/start.rs
```diff
@@ -0,0 +1,82 @@
...
+use std::{io, thread};
+use termion::input::TermRead;
+use tokio::sync::mpsc::UnboundedSender;
+use tokio::time::Duration;
+
+///Starts the dashboard ui with the keyboard listener and tick provider
+pub async fn start_ui_loop(
+    painter: WrappedRcRefCell<DashboardPainter>,
+    state: DashboardState,
+    connection: &mut ClientConnection,
```
Kobzol, line 18: Let's create the `connection` inside `start_ui_loop`, it doesn't need to be passed to it at this moment.

Kobzol:
After the latest changes the dashboard does not show anything (just an empty terminal) on my PC.

spirali (changes requested):
It looks good. Last changes:

* Protect ``calculate_memory_usage_percent`` against division by zero (just return 0 when total is 0)
* Fix clippy warnings related to dashboard (use the newest Rust).

Please ping me when it is finished, then I will accept this PR.

PR #100, opened 2021-07-30
Streaming

spirali:


tests/test_stream.py
```diff
@@ -0,0 +1,230 @@
...
+    result = hq_env.command(["log", "mylog", "show"])
+    lines = set(result.split("\n"))
+    assert len(lines) == 41
+    for i in range(1, 21):
+        assert "{0:02}:0> Hello from {0}".format(i) in lines
+        assert "{0:02}: > stream closed".format(i) in lines
+
+    table = hq_env.command(["log", "mylog", "summary"], as_table=True)
+    table = hq_env.command(["log", "mylog", "summary"], as_table=True)
+    assert table[1] == ["Tasks", "20"]
```
Kobzol, line 39: `check_value_row`?

Cargo.toml
```diff
@@ -42,12 +42,13 @@ gethostname = "0.2.1"
 chrono = { version = "0.4.19", features = ["serde"] }
 orion = "0.16.0"
 hex = "0.4.3"
-# tako = { path = "../tako" }
```
Kobzol, line 45: Temporarily it would be better to select specific SHA in the remote repository, otherwise CI won't be able to build the PR.

src/stream/reader/logfile.rs
```diff
@@ -0,0 +1,373 @@
...
+    size: u32, // Currently chunk is actually limited to 128kB
+}
+
+pub struct InstanceInfo {
+    channels: [Vec<ChunkInfo>; 2],
+    finished: bool,
+}
+
+pub struct TaskInfo {
+    instances: Map<InstanceId, InstanceInfo>,
```
Kobzol, line 27: Since usually there will be only one instance and they will be mostly contiguous, maybe `(Small)Vec<(InstanceId, InstanceInfo>)` could take less memory?

src/stream/reader/logfile.rs
```diff
@@ -0,0 +1,373 @@
...
+            file,
+            index,
+            start_pos,
+        })
+    }
+
+    fn check_header(file: &mut BufReader<File>) -> anyhow::Result<()> {
+        let mut header = [0u8; 6];
+        file.read_exact(&mut header)?;
+        if &header != b"HQ:log" {
```
Kobzol, line 97: This should be put into a constant, since it's used in two different places.

src/stream/reader/logfile.rs
```diff
@@ -0,0 +1,373 @@
...
+    pub fn cat(&mut self, opts: &CatOpts) -> anyhow::Result<()> {
+        self.file.seek(SeekFrom::Start(self.start_pos))?;
+        let selected_channel_id = match opts.channel {
+            Channel::Stdout => 0,
+            Channel::Stderr => 1,
+        };
+        let mut buffer = Vec::new();
+        let mut last_pos: i64 = self.start_pos as i64;
+        let stdout = std::io::stdout();
+        let mut lock = stdout.lock();
```
Kobzol, line 161: `stdout` is by default line-buffered. Maybe we want a `BufWriter` here? If there are a lot of tasks, each one with small output, and they are interleaved, it *could* be slow to print that.

src/stream/reader/logfile.rs
```diff
@@ -0,0 +1,373 @@
...
+                last_pos = chunk.position as i64 + chunk.size as i64;
+            }
+        }
+        Ok(())
+    }
+
+    fn read_block(file: &mut BufReader<File>) -> anyhow::Result<Option<Block>> {
+        match file.read_u8() {
+            Ok(BLOCK_STREAM_START) => {
+                let task_id = file.read_u32::<byteorder::BigEndian>()?;
```
Kobzol, line 186: Why big endian? :D

src/stream/reader/logfile.rs
```diff
@@ -0,0 +1,373 @@
...
+            Color::Cyan,
+        ];
+
+        let selected_channel_id = opts.channel.as_ref().map(|c| match c {
+            Channel::Stdout => 0,
+            Channel::Stderr => 1,
+        });
+
+        let stdout = std::io::stdout();
+        let mut _lock = stdout.lock();
```
Kobzol, line 246: This `_lock` should be used instead of `println`.

src/stream/server/control.rs
```diff
@@ -0,0 +1,12 @@
+use crate::transfer::messages::StreamStats;
+use crate::JobId;
+use std::path::PathBuf;
+use tako::server::rpc::ConnectionDescriptor;
+use tokio::sync::oneshot;
+
+pub enum StreamServerControlMessage {
+    RegisterStream(JobId, PathBuf, oneshot::Sender<()>),
```
Kobzol, line 8: Could this be a named struct? It's unclear what do these fields mean (especially the last one).

src/worker/start.rs
```diff
@@ -114,35 +135,135 @@ fn replace_placeholders(program: &mut ProgramDefinition) {
...
-    let mut program = def.program;
-
-    if def.pin {
-        pin_program(&mut program, &allocation);
-        program.env.insert(HQ_PIN.into(), "1".into());
+async fn resend_stdio(
+    job_id: JobId,
+    job_task_id: JobTaskId,
+    channel: ChannelId,
+    stdio: Option<impl AsyncReadExt + Unpin>,
```
Kobzol, line 148: This should probably be `AsyncRead`?

src/worker/streamer.rs
```diff
@@ -0,0 +1,265 @@
...
+use tako::transfer::auth::{open_message, seal_message, serialize};
+use tako::worker::rpc::connect_to_server_and_authenticate;
+use tako::InstanceId;
+use tokio::sync::mpsc::channel;
+use tokio::sync::mpsc::Sender;
+use tokio::sync::oneshot;
+use tokio::task::spawn_local;
+use tokio_util::codec::{Framed, LengthDelimitedCodec};
+
+const STREAMER_BUFFER_SIZE: usize = 128;
```
Kobzol, line 26: It would be really nice to have some comments in this file x)

src/stream/server/rpc.rs
```diff
@@ -0,0 +1,263 @@
...
+        }
+    }
+}
+
+type StreamServerStateRef = WrappedRcRefCell<StreamServerState>;
+
+impl StreamServerStateRef {
+    fn new() -> Self {
+        WrappedRcRefCell::wrap(StreamServerState {
+            streams: Default::default(),
```
Kobzol, line 70: `#[derive(Default)]`?

src/server/client.rs
```diff
@@ -358,14 +374,26 @@ async fn handle_submit(
...
     };
+
+    if let Some(log) = message.log {
+        let (sender, receiver) = oneshot::channel();
+        tako_ref.send_stream_control(StreamServerControlMessage::RegisterStream(
+            job_id,
+            submit_dir.join(log),
+            sender,
+        ));
+        assert!(receiver.await.is_ok());
```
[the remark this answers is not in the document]
spirali: It is the same as unwrap, since there is exactly one error in oneshot (sender part was dropped)

src/stream/reader/logfile.rs
spirali: I think that the lock is reentrant and locking an already locked lock is fast, but yes, calling directly method on lock is probably slightly faster

src/stream/server/control.rs
spirali: :+1:

src/worker/start.rs
spirali: :+1:

src/worker/streamer.rs
spirali: I will add some

tests/test_stream.py
spirali: Yeah, I was lazy :)

Cargo.toml
spirali: Ah good point

src/stream/reader/logfile.rs
Kobzol: Yes, it is reentrant. But it looks really weird to lock it and then call `println!`, I originally expected that there will be a deadlock :)

src/stream/reader/logfile.rs
spirali: Yes, you are totally right. But I was trying to avoid some (premature) optimizations to make working version ASAP.

src/stream/reader/logfile.rs
spirali: :+1:

src/stream/reader/logfile.rs
spirali: :+1:

src/stream/reader/logfile.rs
spirali: Random choice:)

src/stream/reader/logfile.rs
spirali: It is also in perf book: https://nnethercote.github.io/perf-book/io.html :)

spirali: Ok, not. I sould read more carefully :)

Issue #102, opened 2021-08-01
Better waiting progress bar


spirali:
Write better progress for waiting:

* There should be two progressbars, one counting jobs, one tasks (as sum over all jobs)

```
[====>           ]  4/10 jobs
[======>         ]  1720/2500 tasks
```

* There should be part indicating which tasks is running/finished/failed/canceled (see progress bar in ``hq job <job_id>``)

Issue #101, opened 2021-08-01
Wait exist status


spirali:
Implement:
* If ``submit --wait`` is used and job and at least one task is canceled/failed, then return non-zero exit-code.
* In ``hq wait XX`` is used and at least one of its tasks is canceled/failed, then return non-zero exit-code.

PR #103, opened 2021-08-02
hq wait: bar + exit code

User3574:
Fixes: #101 
Fixes: #102


Kobzol:
I refactored the code slightly to be simpler.

Before, we had a very simple progress bar, but we used a library. Now there's a quite complicated progress bar and the library was removed :laughing:

Issue #62, opened 2021-06-17
Basic task statistics


spirali:
Show basic statistics about tasks computational time (and maybe an remaining time estimate?)

PR #121, opened 2021-08-09
Job & Task times

spirali:
Fixes: https://github.com/It4innovations/hyperqueue/issues/62


src/client/job.rs
```diff
@@ -202,12 +204,23 @@ pub fn print_job_detail(
             .cell(),
     ]);
 
+    rows.push(vec![
+        "Submission date".cell().bold(true),
+        job.submission_date.round_subsecs(0).cell(),
+    ]);
+
+    rows.push(vec![
+        "Makespan".cell().bold(true),
```
Kobzol, line 213: Maybe Duration instead of Makespan?

src/transfer/messages.rs
```diff
@@ -184,6 +184,12 @@ pub struct JobDetail {
     pub entries: Option<Vec<BString>>,
     pub max_fails: Option<JobTaskCount>,
     pub priority: tako::Priority,
+
+    // Date when job was submitted
+    pub submission_date: DateTime<Utc>,
+
+    // Time when job was completed or now if job is not completed
+    pub completion_date_or_now: DateTime<Utc>,
```
Kobzol, line 192: `completion_date_or_now` is sent over the wire instead of something like `Option<DateTime<Utc>>` to use server time for `now`?

Kobzol (approved):

src/client/job.rs
spirali: I want to reserve "duration" for time that is actively processed. Makespan is from submission to completion, counting even time when there were no workers at all or tasks of job was suppresed by a job with a higher prioirty.

src/transfer/messages.rs
spirali: To compute makespan of job/duration of tasks you basically always needs something like (completion_date_or_now - start) so I wanted to keeped as simple as possible and do not introduce two additional field.

Kobzol (approved):

PR #122, opened 2021-08-10
ENH: create all folder given from user defined stdout and stderr file path

fredykoci:


src/worker/start.rs
```diff
@@ -198,6 +199,15 @@ async fn launcher_main(streamer_ref: StreamerRef, task_ref: TaskRef) -> tako::Re
...
         replace_placeholders(&mut program);
+        // Creates all not existing folders given in stdout
+        let std_folders = match &program.stdout {
+            StdioDef::File(path) => match path.parent() {
+                Some(path) => path.to_str().unwrap(),
+                None => "",
+            },
+            _ => "",
+        };
+        std::fs::create_dir_all(std_folders)?;
```
Kobzol, line 210: You can pass `PathBuf` to `create_dir_all` directly. Try something like
```rust
if let StdioDef::File(path) = program.stdout {
    std::fs::create_dir_all(path)?;
}
```
Please also do the same for `stderr`.

Kobzol (changes requested):

src/worker/start.rs
```diff
@@ -163,6 +165,17 @@ async fn resend_stdio(
...
+    match file {
+        StdioDef::File(path) => match path.parent() {
+            Some(path) => std::fs::create_dir_all(path)?,
+            None => (),
+        },
+        _ => (),
+    };
+    io::Result::Ok(())
+}
+
```
Kobzol, line 178: I would suggest to rewrite it slightly to remove the useless else branches. And also rename it, the current name is not very descriptive.
```rust
fn create_directory_if_needed(file: &StdioDef) -> io::Result<()> {
    if let StdioDef::File(path) = file {
        if let Some(path) = path.parent() {
            std::fs::create_dir_all(path)?;
        }
    }
    Ok(())
}
```

Kobzol (changes requested):

Kobzol (approved):

spirali (commented):
Look good.
Next steps: 

* Remove all warnings
* Update user documentation
* Update changelog.md

Kobzol:
Rebased over master for merge. @fredykoci thanks!

Issue #117, opened 2021-08-05
Warning when more files is created


spirali:
Warn if suspiciously many files will be created and suggest to use "--log" option in the warning message.
(e.g. submit --array=1-10000 hostname will create 20k files).
There are some problems for discussion: What is good limit? Should we just warn or rather make it an error (that can be overridden by an option)

Issue #115, opened 2021-08-05
Generalize ``hq log PATH cat --task=X``


spirali:
* Allow to specify a list and ranges, e.g. --task=1,10-50
* Respect user defined order: i.e. 3,2,1 is has different output than 1,2,3

PR #124, opened 2021-08-12
Log tasks, warning for submitting large arrays

User3574:
Fixes: #115 
Fixes: #117

### DIFF
--- a/src/client/commands/log.rs
+++ b/src/client/commands/log.rs
@@ -1,7 +1,7 @@
 use crate::client::globalsettings::GlobalSettings;
+use crate::common::arraydef::IntArray;
 use crate::common::format::human_size;
 use crate::stream::reader::logfile::{LogFile, Summary};
-use crate::JobTaskId;
 use clap::Clap;
 use cli_table::{print_stdout, Cell, Style, Table};
 use std::path::{Path, PathBuf};
@@ -40,9 +40,9 @@ pub struct CatOpts {
     /// Channel name: "stdout" or "stderr"
     pub channel: Channel,
 
-    /// Print only specific task output
+    /// Print only the specified task(s) output. You can use the array syntax to specify multiple tasks.
     #[clap(long)]
-    pub task: Option<JobTaskId>,
+    pub task: Option<IntArray>,
 
     /// Allow unfinished channel
     #[clap(long)]
--- a/src/client/commands/submit.rs
+++ b/src/client/commands/submit.rs
@@ -23,6 +23,7 @@ use crate::transfer::messages::{
 };
 use crate::{rpc_call, JobId, JobTaskCount};
 
+const SUBMIT_ARRAY_LIMIT: JobTaskCount = 999;
 const DEFAULT_STDOUT_PATH: &str = "stdout.%{JOB_ID}.%{TASK_ID}";
 const DEFAULT_STDERR_PATH: &str = "stderr.%{JOB_ID}.%{TASK_ID}";
 
@@ -163,6 +164,7 @@ pub async fn submit_computation(
 ) -> anyhow::Result<()> {
     let resources = opts.resource_request();
     resources.validate()?;
+
     let (job_type, entries) = if let Some(filename) = opts.each_line {
         let lines = read_lines(&filename)?;
         let def = IntArray::from_range(0, lines.len() as JobTaskCount);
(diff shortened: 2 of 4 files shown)


src/client/commands/log.rs
```diff
@@ -42,7 +42,7 @@ pub struct CatOpts {
 
     /// Print only specific task output
```
Kobzol, line 43: `Print only the specified task(s) output. You can use the array syntax to specify multiple tasks.`

src/client/commands/submit.rs
```diff
@@ -169,6 +171,13 @@ pub async fn submit_computation(
         )
     };
 
+    let log = opts.log;
+    if let JobType::Array(array) = job_type.clone() {
```
Kobzol, line 175: `if let JobType::Array(ref array) = job_type {`

src/client/commands/submit.rs
```diff
@@ -22,6 +22,7 @@ use crate::transfer::messages::{
 };
 use crate::{rpc_call, JobId, JobTaskCount};
 
+const SUBMIT_ARRAY_LIMIT: JobTaskCount = 10000;
```
Kobzol, line 25: I would lower this, at least to `1000` or even `100`. @spirali?

src/stream/reader/logfile.rs
```diff
@@ -206,13 +206,15 @@ impl LogFile {
                 Ok(())
             };
 
-        if let Some(task_id) = opts.task {
-            if let Some(task_info) = self.index.get(&task_id) {
-                let instance = task_info.last_instance();
-                if !instance.finished && opts.allow_unfinished {
-                    anyhow::bail!("Stream for task {} is not finished", task_id);
+        if let Some(array) = opts.task.clone() {
```
Kobzol, line 209: `if let Some(ref array) = opts.task {`

src/stream/reader/logfile.rs
```diff
@@ -206,13 +206,15 @@ impl LogFile {
...
+            for task_id in array.iter() {
+                if let Some(task_info) = self.index.get(&task_id) {
+                    let instance = task_info.last_instance();
+                    if !instance.finished && opts.allow_unfinished {
+                        anyhow::bail!("Stream for task {} is not finished", task_id);
```
Kobzol, line 214: This should maybe now be a warning, and the task should be skipped instead?

Kobzol (changes requested):

src/stream/reader/logfile.rs
```diff
@@ -206,13 +206,16 @@ impl LogFile {
...
-        if let Some(task_id) = opts.task {
-            if let Some(task_info) = self.index.get(&task_id) {
-                let instance = task_info.last_instance();
-                if !instance.finished && opts.allow_unfinished {
-                    anyhow::bail!("Stream for task {} is not finished", task_id);
+        if let Some(ref array) = opts.task {
+            for task_id in array.iter() {
+                if let Some(task_info) = self.index.get(&task_id) {
+                    let instance = task_info.last_instance();
+                    if !instance.finished && opts.allow_unfinished {
```
Kobzol, line 213: Btw @spirali shouldn't this be `!opts.allow_unfinished`? It seems that it was wrong before.

src/client/commands/submit.rs
spirali: It is really hard question :) I would stick with 1000 (or 999 rather so it will triggered at 1000). But I can imagine that even 100 is a good limit for some users

Kobzol: Ok let's go with `999`.

src/client/commands/submit.rs
```diff
@@ -169,6 +171,13 @@ pub async fn submit_computation(
         )
     };
 
+    let log = opts.log;
+    if let JobType::Array(ref array) = job_type {
+        if array.id_count() > SUBMIT_ARRAY_LIMIT && log.is_none() {
+            log::warn!("It's recommended to use --log when submitting large arrays");
```
spirali, line 177: * Check also that both stdout, stderr is not disabled
* Make a message more descriptive (and detailed with the actual number of files) "The job will create XXX number of files for stdout/stderr. You may consider using --log option to stream all outputs into one file". where XXX is 2*number of tasks if both stdout/stderr are not disabled)

src/stream/reader/logfile.rs
```diff
@@ -206,13 +206,16 @@ impl LogFile {
...
+                        log::warn!("Stream for task {} is not finished", task_id);
```
spirali, line 214: Look at the code below (for all tasks). First check that all selected tasks are completed, then print them. If at least one of them is not completed (and opts.allow_unfinished is false) then RETURN an error and do not print any of the tasks

src/stream/reader/logfile.rs
spirali: You are right. May be there should be test on this:)

src/stream/reader/logfile.rs
```diff
@@ -206,13 +206,22 @@ impl LogFile {
                 Ok(())
             };
 
-        if let Some(task_id) = opts.task {
-            if let Some(task_info) = self.index.get(&task_id) {
-                let instance = task_info.last_instance();
-                if !instance.finished && opts.allow_unfinished {
-                    anyhow::bail!("Stream for task {} is not finished", task_id);
+        if let Some(ref array) = opts.task {
```
Kobzol, line 209: Now that I look at it, both branches are pretty much the same. Could you try to unify the code? Either get all of the indices into an Vec or first load a Vec of tasks and then use it.

spirali (changes requested):

PR #125, opened 2021-08-12
Normalization of stream end when job is canceled

spirali:


docs/streaming.md
```diff
@@ -68,6 +68,14 @@ Disabling stderr and streaming only stdout into log file.
 ``hq submit --log=my-log --stderr=none ...``
 
 
+# Guarantees
+
+When a task is *finished* or *failed* with a non-streaming error then it is guaranteed that its stream is fully flushed into the log file.
+
+When a task is *canceled* or *failed* with a streaming error, then the stream is not necessarily fully written into the log file in the moment when the state occurs
+ and some part may be written later, but the stream is eventually closed. In this case, HQ is also allow to drop any suffix of buffered part of the stream.
```
Kobzol, line 76:
```suggestion
 and some part may be written later, but the stream will be eventually closed. In this case, HQ is also allowed to drop any suffix of the buffered part of the stream.
```

src/stream/server/rpc.rs
```diff
@@ -22,37 +22,55 @@ use crate::stream::reader::logfile::{
...
 }
 
-impl StreamServerState {
-    fn get_stream(&mut self, job_id: JobId) -> anyhow::Result<Sender<StreamMessage>> {
-        if let Some(s) = self.streams.get(&job_id) {
+type StreamServerStateRef = WrappedRcRefCell<StreamServerState>;
+
+impl StreamServerStateRef {
+    fn new() -> Self {
+        WrappedRcRefCell::wrap(StreamServerState {
```
Kobzol, line 50: `#[derive(Default)]`?

src/stream/server/rpc.rs
```diff
@@ -61,23 +79,13 @@ impl StreamServerState {
...
-    let msg = ToStreamerMessage::Error(message);
-    let data = tako::transfer::auth::serialize(&msg).unwrap();
-    if sender.send(data.into()).is_err() {
-        log::debug!("Sendiding stream error failed");
+fn send_error(sender: Option<UnboundedSender<Bytes>>, message: String) {
+    if let Some(sender) = sender {
+        let msg = ToStreamerMessage::Error(message);
+        let data = tako::transfer::auth::serialize(&msg).unwrap();
+        if sender.send(data.into()).is_err() {
+            log::debug!("Sendiding stream error failed");
```
Kobzol, line 87:
```suggestion
            log::debug!("Sending stream error failed");
```

Kobzol (approved):
Some comments would be nice :)

PR #126, opened 2021-08-13
Time limit

spirali:


docs/arrays.md
```diff
@@ -59,6 +59,10 @@ By default, when a task fails the computation of job continues.
 You can change it by ``--max-fails=X`` where ``X`` is non-negative integer.
 If more tasks then ``X`` fails, then the rest of non-finished tasks are canceled.
 
+## Time limit
+
+Time limit setted by ``--time-limit`` is a time limit for each task separatatelly.
```
Kobzol, line 64: You can set a time limit for each task separately with the ``--time-limit`` option.

docs/jobs.md
```diff
@@ -47,6 +47,8 @@ from which was the job submitted.
 
 ### Output of the job
 
+!!! If you want to avoid creating many files, see the section about streaming
```
Kobzol, line 50:
```suggestion
!!! Warning

    If you want to avoid creating too many files, see the section about [streaming](streaming.md).
```

docs/streaming.md
```diff
@@ -70,10 +70,13 @@ Disabling stderr and streaming only stdout into log file.
 
 # Guarantees
 
-When a task is *finished* or *failed* with a non-streaming error then it is guaranteed that its stream is fully flushed into the log file.
+When a task is *finished*, *failed* with a non-streaming error then it is guaranteed that its stream is fully flushed into the log file.
```
Kobzol, line 73: The comma seems weird here.

docs/streaming.md
```diff
@@ -70,10 +70,13 @@ Disabling stderr and streaming only stdout into log file.
...
 
-When a task is *canceled* or *failed* with a streaming error, then the stream is not necessarily fully written into the log file in the moment when the state occurs
- and some part may be written later, but the stream will be eventually closed. In this case, HQ is also allowed to drop any suffix of the buffered part of the stream.
+When a task is *canceled* or *failed* with a streaming error, then the stream is not necessarily fully written into the log file in the moment when the state occurs and some part may be written later, but the stream will be eventually closed.
+
+When a task is *canceled*, *failed* because of its time limit, or streaming error
+then all task activities are terminated immediately including streaming and
+a suffix of the stream buffered in the worker may be lost.
```
Kobzol, line 79: These sentences don't make sense. Both contain `canceled` and `streaming error` situations, but with different outocmes.

src/client/commands/submit.rs
```diff
@@ -137,6 +138,9 @@ pub struct SubmitOpts {
     #[clap(long, default_value = "0")]
     priority: tako::Priority,
 
+    #[clap(long)]
+    time_limit: Option<ArgDuration>,
```
Kobzol, line 142: Add documentation please.

src/worker/start.rs
```diff
@@ -245,47 +271,58 @@ async fn run_task(
...
+
+        let guard_fut = async move {
+            let result = tokio::select! {
+                biased;
+                    r = end_receiver => {
+                        Ok(r.unwrap().into())
+                    }
+                    r = main_fut => r
+            };
             stream.close().await.map_err(streamer_error)?;
```
Kobzol, line 296: Maybe the errors should be combined here somehow? If `result` has a failure and stream closing also fails, the first failure will be lost.

Kobzol (approved):

Cargo.toml
```diff
@@ -43,7 +43,7 @@ chrono = { version = "0.4.19", features = ["serde"] }
 orion = "0.16.0"
 hex = "0.4.3"
 # tako = { path = "../tako" }
-tako = { git = "https://github.com/spirali/tako", branch = "main" }
+tako = { git = "https://github.com/spirali/tako", branch = "timelimit" }
```
Kobzol, line 46: Also don't forget to update this before merging.

Kobzol (changes requested):

docs/arrays.md
spirali: My formulation is bad but your wording suggest that you can set different time limit for each task

docs/streaming.md
spirali: They speak about two different things, the first is about the state of log file
The second one says that some suffix may be dropped

src/worker/start.rs
spirali: Streaming error is an infrastructure error, so it has a higher precedence (and it should almost never happen), so i think that it is ok here

Cargo.toml
spirali: Thank you for reminding :)

src/worker/start.rs
Kobzol: Ok, I agree that it should be rare (hopefullly :) ).

PR #134, opened 2021-09-08
ENH: Apply placeholders to log paths

fredykoci:


Kobzol:
Merged the implementation with https://github.com/It4innovations/hyperqueue/pull/133 to avoid multiple `placeholders.rs` files and added a simple test.

Issue #104, opened 2021-08-03
Create overhead benchmark


Kobzol:
Benchmark HyperQueue and other related tools with a simple benchmark that measures the overhead of the system.
The benchmark should run 10k, 100k, 1 million tasks that "do nothing" (`sleep 0`, `echo`, `hostname`) etc. on an HPC cluster. The benchmark should not output any stdout/stderr files if possible.

Tools to benchmark:
- [ ] HyperQueue
- [ ] #105
- [ ] #106

The experiment can be performed on Barbora or Karolina.

Kobzol:
Measure overhead of HQ so far:
- Barbora/Karolina - 1ms
- Notebook - 0.1ms

TODO: profile what causes higher overhead on Barbora/Karolina. It seems to be caused by the worker.

Kobzol:
We have a lot of overhead benchmarks in the repository now.

PR #136, opened 2021-09-10
Reimplement `hq wait` to avoid using polling

Kobzol:
Previously, waiting simply polled the server once every second, which meant that it couldn't be used for benchmarking short jobs because of the added polling pauses.

Now waiting connects to the server, waits until the jobs are finished, and then ends immediately.

A disadvantage is that the interactive progress bar is no longer shown. But it can be emulated with `watch hq job <job-id>`. If needed, the progress bar could be moved to another command (`hq wait --interactive` or something like that).

This should unblock https://github.com/It4innovations/hyperqueue/issues/104

### DIFF
--- a/src/server/client.rs
+++ b/src/server/client.rs
@@ -17,7 +17,7 @@ use crate::common::manager::info::ManagerType;
 use crate::common::serverdir::ServerDir;
 
 use crate::server::autoalloc::{PbsHandler, QueueDescriptor, QueueInfo};
-use crate::server::job::{Job, JobState};
+use crate::server::job::{Job, JobState, JobTaskCounters};
 use crate::server::rpc::Backend;
 use crate::server::state::{State, StateRef};
 use crate::stream::server::control::StreamServerControlMessage;
@@ -140,7 +140,19 @@ pub async fn client_rpc_loop<
     }
 }
 
+/// Waits until all jobs matched by the `selector` are finished (either by completing successfully,
+/// failing or being canceled).
 async fn handle_wait_for_jobs_message(state_ref: &StateRef, selector: Selector) -> ToClientMessage {
+    let update_counters = |response: &mut WaitForJobsResponse, counters: &JobTaskCounters| {
+        if counters.n_canceled_tasks > 0 {
+            response.canceled += 1;
+        } else if counters.n_failed_tasks > 0 {
+            response.failed += 1;
+        } else {
+            response.finished += 1;
+        }
+    };
+
     let (receivers, mut response) = {
         let mut state = state_ref.get_mut();
         let job_ids: Vec<JobId> = get_job_ids(&state, selector);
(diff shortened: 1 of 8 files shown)


User3574 (approved):

src/server/client.rs
```diff
@@ -136,6 +140,52 @@ pub async fn client_rpc_loop<
...
+
+        let mut response = WaitForJobsResponse::default();
+        let mut receivers = vec![];
+
+        for job_id in job_ids {
+            match state.get_job_mut(job_id) {
+                Some(job) => {
+                    if job.is_terminated() {
+                        if job.counters.has_unsuccessful_tasks() {
+                            response.failed += 1;
```
spirali, line 156: This does not seem right, has_unsucessful_tasks counts also "canceled"; however, "response.failed" looks like it counts only fail states (so some error messages later may be wrong?)

src/server/job.rs
```diff
@@ -107,6 +112,12 @@ pub struct Job {
 
     pub submission_date: DateTime<Utc>,
     pub completion_date: Option<DateTime<Utc>>,
+
+    /// Holds channels that will receive information about the job task counters after the job
+    /// finishes in any way.
+    ///
+    /// You can subscribe to the finish updates with [`Self::subscribe_to_completion`].
+    completion_handlers: Vec<oneshot::Sender<JobTaskCounters>>,
```
spirali, line 120: Why handlers? Isnt it raher "callbacks" or "senders"?

spirali, line 120: Btw: Isnt it more flexible just sent job id rather than just counters? I understand that i would need an extra lookup but it seems to me more flexible and sending counters looks as a kind of hack

src/server/client.rs
Kobzol: `failed` counts both failed and canceled tasks. It's purpose is to cause `wait` to exit with a non-success error code if `failed > 0`.

src/server/job.rs
Kobzol: Good point, `callback` is a better name here.

Kobzol: Currently, there is only one usage that needs job counters, so it could be generalized later. But it's a small thing, I'll change it to job id. The current implementation would also work fine if it returned just `()`.

src/server/client.rs
spirali: Then it is incorrectly named, "failed" has a quite precise meaning in HQ.

src/server/job.rs
spirali: JobId :+1:

src/server/client.rs
Kobzol: So should `wait` also track canceled jobs? Or just rename it to e.g. `unsuccessful`?

spirali (approved):

Issue #95, opened 2021-07-23
Quiet flag


spirali:
Implement ``hq --quiet XXX`` flag that will make minimalistic (and parsable) output for standard commands. In particular 

``hq --quiet submit XXXX`` should just print job id on stdout + newline in case of success (exit code 0).

Issue #28, opened 2021-05-07
Option to print tables in a simplified way (for easier automatic processing)


spirali:
* Export to JSON or CSV?
* Simplified view without borders?

PR #139, opened 2021-09-15
Printing generalization for HyperQueue

User3574:
Current state: Moved printer into GlobalSettings. 

This PR is a first step towards solving #95 and #28.


src/client/worker.rs
```diff
@@ -1,72 +1,64 @@
...
-        Some(WorkerExitInfo {
-            reason: LostWorkerReasonInfo::IdleTimeout,
-            ..
-        }) => "IDLE TIMEOUT".cell().foreground_color(Some(Color::Cyan)),
-        Some(WorkerExitInfo {
-            reason: LostWorkerReasonInfo::Stopped,
-            ..
-        }) => "STOPPED".cell().foreground_color(Some(Color::Magenta)),
-    }
+#[derive(Serialize, Deserialize)]
```
Kobzol, line 15: What is the purpose of this struct?

Cargo.toml
```diff
@@ -52,6 +52,7 @@ bstr = { version = "0.2", features = ["serde1"] }
 colored = "2"
 byteorder = "1.4"
 smallvec = "1.6.1"
+csv = "1.1.6"
```
Kobzol, line 55: Leftover change.

src/client/worker.rs
User3574: This is a leftover that is currently not needed, since it was used for Json outputting by serde_json.

src/client/output/cli.rs
```diff
@@ -0,0 +1,577 @@
...
+                record.worker_port().cell(),
+            ],
+            vec![
+                "Start date".cell().bold(true),
+                record.start_date().format("%F %T %Z").cell(),
+            ],
+            vec!["Version".cell().bold(true), record.version().cell()],
+        ];
+        let table = rows.table().color_choice(self.color_policy);
+        assert!(print_stdout(table).is_ok());
```
Kobzol, line 152: These two lines are repeated in almost every method, let's extract them into a function.

spirali:
Just an idea. Why not to put printer in GlobalSettings? So it does not have be passed as an extra argument all over the place.

Kobzol:
After this PR, most of the functions won't need `GSettings`, just the printer. But it could be unified if you want.

spirali:
Then rather pass GlobalSettings rather printer as if we add a global flag (e.g. a kind "debug" flag), we do not have to pass everything again.

Kobzol:
Ok. @User3574 After you get back to this PR, please put `Box<dyn Printer>` into the global settings and pass the settings to the functions instead of the printer.

src/client/commands/autoalloc.rs
```diff
@@ -221,69 +185,29 @@ async fn print_event_log(
...
+        self.started_at
+    }
+    pub fn get_finished_at(&self) -> Option<SystemTime> {
+        self.finished_at
+    }
+}
+
+pub fn get_allocation_times(allocation: &Allocation) -> AllocationTimes {
     let mut started = None;
     let mut finished = None;
```
Kobzol, line 212: Let's move this into `cli.rs`. It's used only for printing.

src/client/commands/wait.rs
```diff
@@ -1,6 +1,6 @@
 use std::time::SystemTime;
 
-use crate::client::utils::{TASK_COLOR_CANCELED, TASK_COLOR_FAILED, TASK_COLOR_FINISHED};
+use crate::client::output::cli::{TASK_COLOR_CANCELED, TASK_COLOR_FAILED, TASK_COLOR_FINISHED};
 use crate::rpc_call;
 use crate::transfer::connection::ClientConnection;
 use crate::transfer::messages::{FromClientMessage, Selector, ToClientMessage, WaitForJobsRequest};
```
Kobzol, line 6: Printify the output in `wait_for_jobs` too, please.

src/client/job.rs
```diff
@@ -34,212 +24,7 @@ pub async fn get_worker_map(connection: &mut ClientConnection) -> anyhow::Result
...
-            &job.info.counters,
-            &worker_map,
-        );
-    }
-}
-
-const MAX_DISPLAYED_WORKERS: usize = 2;
-
-fn format_job_workers(job: &JobDetail, worker_map: &WorkerMap) -> String {
+pub fn format_job_workers(job: &JobDetail, worker_map: &WorkerMap) -> String {
```
Kobzol, line 27: Let's also move these `format_*` functions to `cli.rs`. If you're using IntelliJ, you can move them using `F6`.

src/client/worker.rs
```diff
@@ -1,72 +1,61 @@
...
-use cli_table::{print_stdout, Cell, CellStruct, Color, Style, Table};
-
-use crate::client::globalsettings::GlobalSettings;
 use crate::common::manager::info::GetManagerInfo;
 use crate::transfer::messages::{LostWorkerReasonInfo, WorkerExitInfo, WorkerInfo};
+use crate::WorkerId;
+
+use std::string::String;
 
 pub enum WorkerState {
```
Kobzol, line 7: This seems to be unused.

src/client/worker.rs
```diff
@@ -1,72 +1,61 @@
...
-        Some(WorkerExitInfo {
-            reason: LostWorkerReasonInfo::IdleTimeout,
-            ..
-        }) => "IDLE TIMEOUT".cell().foreground_color(Some(Color::Cyan)),
-        Some(WorkerExitInfo {
-            reason: LostWorkerReasonInfo::Stopped,
-            ..
-        }) => "STOPPED".cell().foreground_color(Some(Color::Magenta)),
-    }
+pub struct WorkerDetails {
```
Kobzol, line 13: Let's move this into `cli.rs`.

src/client/globalsettings.rs
```diff
@@ -1,24 +1,19 @@
+use crate::client::output::outputs::Output;
 use std::path::{Path, PathBuf};
 
-use cli_table::ColorChoice;
-
 pub struct GlobalSettings {
-    color_policy: ColorChoice,
     server_dir: PathBuf,
+    pub printer: Box<dyn Output>,
```
Kobzol, line 6: Please keep it private and create a getter (like `pub fn printer(&self) -> &dyn Output`). You can create a getter using `Alt + Insert` on the struct in IntelliJ.

Kobzol (changes requested):

src/client/output/cli.rs
```diff
@@ -0,0 +1,877 @@
+use cli_table::format::Justify;
```
Kobzol, line 1: Let's remove all `pub` from this file, except for `CliOutput` and its field.

Kobzol (changes requested):

Kobzol (approved):
Great :) LGTM.

spirali (approved):

PR #141, opened 2021-09-16
WIP: Overview Streaming

vyomkeshj:
The PR is to add a feature that allows the hq server to poll for hardware overview and stream it to a file for storage. 
It uses Streaming and modifies it to accept the overview message from the hq server and then persist it to file.

Later the stored data can be used to provide post mortem analysis on the dashboard.

### DIFF
--- a/src/stream/reader/logfile.rs
+++ b/src/stream/reader/logfile.rs
@@ -8,6 +8,7 @@ use std::collections::BTreeMap;
 use std::fs::File;
 use std::io::{BufReader, BufWriter, ErrorKind, Read, Seek, SeekFrom, Write};
 use std::path::Path;
+use tako::messages::gateway::CollectedOverview;
 use tako::InstanceId;
 
 pub const HQ_LOG_HEADER: &[u8] = b"HQ:log";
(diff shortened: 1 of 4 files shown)


vyomkeshj:
It's not ready yet, I'd like to confirm if this is the right way :)

Kobzol:
Please write a PR description that describes what is this PR trying to achieve and what specific questions do you have. I don't know what exactly is going on here and there's not much code to review.

src/stream/reader/logfile.rs
```diff
@@ -8,13 +8,15 @@ use std::collections::BTreeMap;
...
 use std::path::Path;
+use tako::messages::gateway::CollectedOverview;
 use tako::InstanceId;
 
 pub const HQ_LOG_HEADER: &[u8] = b"HQ:log";
 pub const HQ_LOG_VERSION: u32 = 0;
 
 pub const BLOCK_STREAM_START: u8 = 0;
 pub const BLOCK_STREAM_CHUNK: u8 = 1;
+pub const BLOCK_STREAM_OVERVIEW: u8 = 3;
```
spirali, line 19: It should be named as just BLOCK_OVERVIEW. In log there are saved two things: Streams & Overview.

spirali: Other then that, the code skeleton seems ok so far.

Kobzol:
This will be probably handled by the new event monitoring.

Issue #143, opened 2021-09-23
Auto allocation discussion


spirali:
Some comments after more detailed experimentation:

1. hq auto-alloc is quite long, what about "hq alloc" only?

2. hq alloc-alloc add: What about making name optional, e.g. using queue name as default

3. hq auto-alloc allocations => If it ends with "s" I would expect that it returns list without additional argument. What about "hq alloc detail X"

4. hq auto-alloc stop?

5. Refresh interval in info - it seems quite confusing to me, I would expect something like "watch" behavior if "refresh interval" is shown

6. hq auto-alloc allocations produces large tables. It cannot be snown in my  console. Some shortening? More lines per entry?

spirali:
Second thoughs:

7. What about same naming scheme as "hq worker"?
hq auto-alloc info => hq auto-alloc list
hq auto-alloc allocations X => hq auto-alloc info X


8. What about removing names and automatically assigning IDs and then just use IDs as in case of job/workers?

PR #144, opened 2021-09-24
Autoalloc improvements

Kobzol:
Best reviewed commit by commit.

TODO (possibly for future PRs):
- [ ] `hq alloc stop`
- [ ] Size of `hq alloc info <id>` table

Related issue: https://github.com/It4innovations/hyperqueue/issues/143


spirali (approved):
LGTM, but please rather do not create short arguments. I would like to create them at once with some a central logic, rather than ad-hoc shortenings. (But for this code, just leave them as it is)

PR #163, opened 2021-10-11
Better error message for submit

fredykoci:


src/server/bootstrap.rs
```diff
@@ -59,15 +59,15 @@ pub async fn get_client_connection(server_directory: &Path) -> anyhow::Result<Cl
     let sd = ServerDir::open(server_directory).context("Invalid server directory")?;
     let access_record = sd.read_access_record().with_context(|| {
         format!(
-            "No running instance of HQ found (cannot read access record from {:?})",
+            "No running instance of HQ found (cannot read access record from {:?}), start server: <hq server start> or specify access record",
```
Kobzol, line 62: It would be nicer to show the exact command to start the server, like:
```rust
"No running instance of HQ found (cannot read access record from {:?}). Try to start the server using `hq server start`".
```

If the server-directory does not match the default one, the example command should also contain `--server-dir <server-directory>`.

Kobzol (changes requested):

src/server/bootstrap.rs
```diff
@@ -56,22 +56,29 @@ pub async fn init_hq_server(
 }
 
 pub async fn get_client_connection(server_directory: &Path) -> anyhow::Result<ClientConnection> {
+    let mut default_home = dirs::home_dir().unwrap_or_else(std::env::temp_dir);
```
Kobzol, line 59: Use `default_server_directory_path` to avoid unexpected changes in the default directory. Move this function to `client/mod.rs` to make it available here.

src/server/bootstrap.rs
```diff
@@ -56,22 +56,29 @@ pub async fn init_hq_server(
...
+    default_home.push(".hq-server");
+
     let sd = ServerDir::open(server_directory).context("Invalid server directory")?;
     let access_record = sd.read_access_record().with_context(|| {
         format!(
-            "No running instance of HQ found (cannot read access record from {:?})",
+            "No running instance of HQ found (cannot read access record from {:?}), start server: <hq server start> or specify access record",
```
Kobzol, line 65: I think that we should show the `--server-dir` flag also in this meesage.

src/server/bootstrap.rs
```diff
@@ -56,22 +56,29 @@ pub async fn init_hq_server(
...
     })?;
+    let server_dir_msg = if default_home != server_directory { format!("--server-dir {}", server_directory.to_str().unwrap()) } else { String::new()};
+
     let connection = HqConnection::connect_to_server(&access_record)
         .await
-        .with_context(|| {
+        .with_context(||
             format!(
-                "Cannot connect to HQ server at port {}:{}",
+                "Access token found but HQ server {}:{} is unreachable. Try to start the server using `hq server start {}`",
```
Kobzol, line 75: Use `hq server start{}` and add the space conditionally to avoid unnecessary space.

src/server/bootstrap.rs
```diff
@@ -88,7 +95,7 @@ async fn get_server_status(server_directory: &Path) -> crate::Result<ServerStatu
 async fn initialize_server(
     gsettings: &GlobalSettings,
     server_cfg: ServerConfig,
-) -> anyhow::Result<(impl Future<Output = anyhow::Result<()>>, Rc<Notify>)> {
+) -> anyhow::Result<(impl Future<Output=anyhow::Result<()>>, Rc<Notify>)> {
```
Kobzol, line 98: It seems that your local formatter is overriding the formatter that we use. What Rust version do you use, 1.55?
If this formatting is produced by IntelliJ, please enable `Use rustfmt instead of built-in formatter` in `Settings -> Languages & Frameworks -> Rust -> Rustfmt`.

PR #187, opened 2021-10-29
Dashboard: Worker Connection Chart

vyomkeshj:
Added worker connection chart to dashboard.
Screenshot 2021-10-29 at 19 29 54

### DIFF
--- a/crates/hyperqueue/src/dashboard/ui/screens/home/chart_section.rs
+++ b/crates/hyperqueue/src/dashboard/ui/screens/home/chart_section.rs
@@ -1,0 +1,104 @@
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
+use tako::messages::gateway::CollectedOverview;
+
+use crate::dashboard::ui::terminal::DashboardFrame;
+use tui::layout::Rect;
+use tui::style::{Color, Modifier, Style};
+use tui::symbols;
+use tui::text::Span;
+use tui::widgets::{Axis, Block, Borders, Chart, Dataset};
+
+/// Stores the number of connected workers at a time.
+struct WorkerCountRecord {
+    time: SystemTime,
+    worker_count: u32,
+}
+
+pub struct ClusterOverviewChart {
+    /// Worker count at different times.
+    worker_records: Vec<WorkerCountRecord>,
+    /// The time for which the data is plotted for.
+    view_size: Duration,
+}
+
+impl ClusterOverviewChart {
+    pub fn update(&mut self, overview: &CollectedOverview) {
+        self.worker_records.push(WorkerCountRecord {
+            time: SystemTime::now(),
+            worker_count: overview.worker_overviews.len() as u32,
+        });
+    }
+
+    pub fn draw(&mut self, rect: Rect, frame: &mut DashboardFrame) {
+        let workers_in_view = self.get_workers_in_view();
+        let max_workers_in_view = workers_in_view
+            .iter()
+            .map(|&(_, count)| count)
+            .max()
+            .unwrap_or(0) as f64;
+
+        let worker_counts: Vec<(f64, f64)> = workers_in_view
+            .iter()
+            .map(|x| (x.0 as f64, x.1 as f64))
+            .collect();
+        let datasets = vec![Dataset::default()
+            .name("workers_connected")
+            .marker(symbols::Marker::Braille)
+            .style(Style::default().fg(Color::White))
+            .data(&worker_counts)];
+
+        let chart = Chart::new(datasets)
+            .block(
+                Block::default()
+                    .title(Span::styled(
+                        "Worker Connection History",
+                        Style::default()
+                            .fg(Color::White)
+                            .add_modifier(Modifier::BOLD),
+                    ))
+                    .borders(Borders::ALL),
+            )
+            .x_axis(
+                Axis::default()
+                    .title("time ->")
+                    .style(Style::default().fg(Color::Gray))
+                    .bounds([
+                        get_time_as_secs(SystemTime::now() - self.view_size) as f64,
+                        get_time_as_secs(SystemTime::now()) as f64,
+                    ]),
+            )
+            .y_axis(
+                Axis::default()
+                    .style(Style::default().fg(Color::Gray))
+                    .bounds([0.0, max_workers_in_view])
+                    .labels(vec![
+                        Span::from(0.to_string()),
+                        Span::from(max_workers_in_view.to_string()),
+                    ]),
+            );
+        frame.render_widget(chart, rect);
+    }
+
+    /// Returns the worker connection chart data for the view size.
+    fn get_workers_in_view(&self) -> Vec<(u64, u64)> {
+        let current_time = SystemTime::now();
+        self.worker_records
+            .iter()
+            .filter(|count| count.time > current_time - self.view_size)
+            .map(|record| (get_time_as_secs(record.time), record.worker_count as u64))
+            .collect()
+    }
+}
+
+impl Default for ClusterOverviewChart {
+    fn default() -> Self {
+        Self {
+            worker_records: vec![],
+            view_size: Duration::from_secs(300),
+        }
+    }
+}
+
+fn get_time_as_secs(time: SystemTime) -> u64 {
+    time.duration_since(UNIX_EPOCH).unwrap().as_secs()
+}
--- a/crates/hyperqueue/src/dashboard/ui/screens/home/screen.rs
+++ b/crates/hyperqueue/src/dashboard/ui/screens/home/screen.rs
@@ -1,34 +1,39 @@
 use termion::event::Key;
-use tui::layout::{Constraint, Direction, Rect};
 
 use crate::dashboard::ui::screen::{ClusterState, Screen};
+use crate::dashboard::ui::screens::home::chart_section::ClusterOverviewChart;
 use crate::dashboard::ui::screens::home::info_section::WorkerInfoTable;
 use crate::dashboard::ui::screens::home::worker_utilization_table::WorkerUtilTable;
 use crate::dashboard::ui::styles::style_header_text;
 use crate::dashboard::ui::terminal::DashboardFrame;
 use crate::dashboard::ui::widgets::text::draw_text;
 
+use tui::layout::{Constraint, Direction, Layout, Rect};
+
 #[derive(Default)]
 pub struct HomeScreen {
     worker_info_table: WorkerInfoTable,
     worker_util_table: WorkerUtilTable,
+    worker_info_chart: ClusterOverviewChart,
 }
 
 impl Screen for HomeScreen {
     fn draw(&mut self, frame: &mut DashboardFrame) {
-        let layout = Layout::new(frame);
+        let layout = HomeLayout::new(frame);
         draw_text("HQ top", layout.header_chunk, frame, style_header_text());
         // Update the current selected worker for info table
         let selected_worker = self.worker_util_table.get_selected_item();
         self.worker_info_table.change_info(selected_worker);
 
+        self.worker_info_chart.draw(layout.chart_chunk, frame);
         self.worker_info_table.draw(layout.info_chunk, frame);
         self.worker_util_table.draw(layout.body_chunk, frame);
     }
 
     fn update(&mut self, state: ClusterState) {
-        self.worker_util_table.update(state.overview);
+        self.worker_util_table.update(&state.overview);
         self.worker_info_table.update(state.worker_info);
+        self.worker_info_chart.update(&state.overview)
     }
 
     fn handle_key(&mut self, key: Key) {
@@ -41,30 +46,40 @@ impl Screen for HomeScreen {
 }
 
 /**
-*  _____________________
-   |     HEADER        |
-   |-------------------|
-   |      BODY         |
-   ---------------------
+*  __________________________
+   |     Chart |    Info   |
+   |--------Header---------|
+   |-----------------------|
+   |          BODY         |
+   -------------------------
  **/
-struct Layout {
+struct HomeLayout {
+    chart_chunk: Rect,
     info_chunk: Rect,
     header_chunk: Rect,
     body_chunk: Rect,
 }
 
-impl Layout {
+impl HomeLayout {
     fn new(frame: &DashboardFrame) -> Self {
-        let base_chunks = vertical_chunks(
-            vec![
+        let base_chunks = tui::layout::Layout::default()
+            .constraints(vec![
                 Constraint::Percentage(30),
                 Constraint::Percentage(10),
-                Constraint::Percentage(60),
-            ],
-            frame.size(),
-        );
+                Constraint::Percentage(30),
+            ])
+            .direction(Direction::Vertical)
+            .split(frame.size());
+
+        let info_chunks = Layout::default()
+            .constraints(vec![Constraint::Percentage(30), Constraint::Percentage(70)])
+            .direction(Direction::Horizontal)
+            .margin(0)
+            .split(base_chunks[0]);
+
         Self {
-            info_chunk: base_chunks[0],
+            chart_chunk: info_chunks[0],
+            info_chunk: info_chunks[1],
             header_chunk: base_chunks[1],
             body_chunk: base_chunks[2],
         }
@@ -75,5 +90,17 @@ pub fn vertical_chunks(constraints: Vec<Constraint>, size: Rect) -> Vec<Rect> {
     tui::layout::Layout::default()
         .constraints(constraints.as_ref())
         .direction(Direction::Vertical)
+        .split(size)
+}
+
+pub fn horizontal_chunks_with_margin(
+    constraints: Vec<Constraint>,
+    size: Rect,
+    margin: u16,
+) -> Vec<Rect> {
+    Layout::default()
+        .constraints(constraints.as_ref())
+        .direction(Direction::Horizontal)
+        .margin(margin)
         .split(size)
 }
--- a/crates/hyperqueue/src/dashboard/ui/screens/home/mod.rs
+++ b/crates/hyperqueue/src/dashboard/ui/screens/home/mod.rs
@@ -1,3 +1,4 @@
+pub mod chart_section;
 pub mod info_section;
 pub mod screen;
 pub mod worker_utilization_table;
--- a/crates/hyperqueue/src/dashboard/ui/screens/home/worker_utilization_table.rs
+++ b/crates/hyperqueue/src/dashboard/ui/screens/home/worker_utilization_table.rs
@@ -14,7 +14,7 @@ pub struct WorkerUtilTable {
 }
 
 impl WorkerUtilTable {
-    pub fn update(&mut self, overview: CollectedOverview) {
+    pub fn update(&mut self, overview: &CollectedOverview) {
         let rows = create_rows(overview);
         self.table.set_items(rows);
     }
@@ -91,7 +91,7 @@ struct WorkerUtilRow {
     collection_timestamp: Option<u64>,
 }
 
-fn create_rows(overview: CollectedOverview) -> Vec<WorkerUtilRow> {
+fn create_rows(overview: &CollectedOverview) -> Vec<WorkerUtilRow> {
     overview
         .worker_overviews
         .iter()


crates/hyperqueue/src/dashboard/ui/screens/home/screen.rs
```diff
@@ -77,3 +90,15 @@ pub fn vertical_chunks(constraints: Vec<Constraint>, size: Rect) -> Vec<Rect> {
         .direction(Direction::Vertical)
         .split(size)
 }
+
+pub fn horizontal_chunks_with_margin(
```
Kobzol, line 94: It seems like it would be shorter to just inline `vertical_chunks` and `horizontal_chunks_with_margin`.

crates/hyperqueue/src/dashboard/ui/screens/home/chart_section.rs
```diff
@@ -0,0 +1,147 @@
...
+                        Span::from(0.to_string()),
+                        Span::from((self.max_observed_workers / 2).to_string()),
+                        Span::from((self.max_observed_workers as f64).to_string()),
+                    ]),
+            );
+        frame.render_widget(chart, rect);
+    }
+}
+
+fn get_current_time() -> u128 {
```
Kobzol, line 114: Let's represent a timestamp in some unified and semantic way. For example `SystemTime` or `Instant`. Not by a "raw" number.

crates/hyperqueue/src/dashboard/ui/screens/home/chart_section.rs
```diff
@@ -0,0 +1,147 @@
...
+    /// Current number of workers.
+    WorkerCountUpdate(u32),
+    /// The average cpu utilization of the worker.
+    AvgCpuUsageUpdate(f32),
+}
+
+#[derive(Clone)]
+pub struct ClusterOverviewChart {
+    /// The state of the worker at different times.
+    timed_info_vec: Vec<WorkerStateAtTime>,
```
Kobzol, line 30: The data for the chart is updated in discrete steps, and in each step we get new information for each observed value on the Y axis. So I think that it would make more sense to represent the dataset as a collection of items (x_value, y_values). There's no need to repeat the X-value inside the Vec here.

crates/hyperqueue/src/dashboard/ui/screens/home/chart_section.rs
```diff
@@ -0,0 +1,147 @@
...
+}
+
+impl ClusterOverviewChart {
+    pub fn update(&mut self, overview: CollectedOverview) {
+        // Update the number of workers connected
+        let time = get_current_time();
+        let num_workers = overview.worker_overviews.len();
+        if num_workers > self.max_observed_workers as usize {
+            self.max_observed_workers = num_workers as u64;
+        }
```
Kobzol, line 46: `self.max_observed_workers = self.max_observed_workers.max(num_workers as u64);`
but I don't think that we have to cache it here.

crates/hyperqueue/src/dashboard/ui/screens/home/chart_section.rs
```diff
@@ -0,0 +1,147 @@
...
+    timed_info_vec: Vec<WorkerStateAtTime>,
+    /// Used to set the min bound of the chart's x-axis
+    time_reference: u128,
+    /// Used to set the max bound of the chart's y-axis
+    max_observed_workers: u64,
+    /// Current number of workers connected
+    current_conn_workers: u64,
+}
+
+impl ClusterOverviewChart {
```
Kobzol, line 39: The chart currently behaves in a quite weird way, because it start at the "beginning of time" and then it will just shrink forever to show the actual state. I think that instead we should do the following:
- Store the complete history of X and Y axis values received from the server (if this grows too large, we can instead use some rotating buffer, but let's ignore that for now).
- When drawing the chart, take some fixed amount of data from the event log (e.g. last 5 minutes) and only draw this "view" of the data. Later we can add moving in history, so the amount of history and its "location" will be able to be changed using e.g. keybinds.
- Compute the maximum of the Y axis based only on the subset of the data that is currently being drawn.
- Do not store `time_reference`. Use "current time" - "view size" (e.g. 5 minutes) as the reference value.

crates/hyperqueue/src/dashboard/ui/screens/home/chart_section.rs
```diff
@@ -0,0 +1,147 @@
...
+        if num_workers > self.max_observed_workers as usize {
+            self.max_observed_workers = num_workers as u64;
+        }
+        self.timed_info_vec.push(WorkerStateAtTime {
+            at_time: time,
+            event: TimeEventType::WorkerCountUpdate(num_workers as u32),
+        });
+        // Add more points at the same time to smoothen the transition in plot
+        let interpolation =
+            interpolate_y_axis(self.current_conn_workers as f64, num_workers as f64, 30_f64);
```
Kobzol, line 53: Maybe just remove this for the moment. It might look a bit weird, but the number of workers is a discrete value, so it makes sense that it will change in discrete steps.

crates/hyperqueue/src/dashboard/ui/screens/home/screen.rs
```diff
@@ -1,34 +1,39 @@
...
         self.worker_info_table.change_info(selected_worker);
 
+        self.worker_info_chart.draw(layout.chart_chunk, frame);
         self.worker_info_table.draw(layout.info_chunk, frame);
         self.worker_util_table.draw(layout.body_chunk, frame);
     }
 
     fn update(&mut self, state: ClusterState) {
-        self.worker_util_table.update(state.overview);
+        self.worker_util_table.update(state.overview.clone());
```
Kobzol, line 34: I suppose that now is the time to change the parameter of the update method to `&CollectedOverview`.

Kobzol (changes requested):

crates/hyperqueue/src/dashboard/ui/screens/home/chart_section.rs
vyomkeshj: what about humantime::Timestamp? can I use it?

Kobzol: That's just a wrapper around `SystemTime`, let's just use `SystemTime` directly.

crates/hyperqueue/src/dashboard/ui/screens/home/chart_section.rs
```diff
@@ -0,0 +1,132 @@
...
+impl Default for ClusterOverviewChart {
+    fn default() -> Self {
+        Self {
+            worker_count_vec: vec![],
+            view_size: Duration::new(30, 0),
+        }
+    }
+}
+
+fn get_time_as_secs() -> Option<u64> {
```
Kobzol, line 127: Just `unwrap` this. It should be super rare and its inside the client, so it's not critical if it crahes.

crates/hyperqueue/src/dashboard/ui/screens/home/chart_section.rs
```diff
@@ -0,0 +1,114 @@
...
+use tui::layout::{Alignment, Rect};
+use tui::style::{Color, Modifier, Style};
+use tui::symbols;
+use tui::text::{Span, Spans};
+use tui::widgets::{Axis, Block, Borders, Chart, Dataset, Paragraph, Wrap};
+
+#[derive(Clone)]
+pub struct ClusterOverviewChart {
+    /// (time, #workers)
+    worker_count_vec: Vec<(f64, f64)>,
```
Kobzol, line 15: I would keep some struct here. If you can encode something using the type system, it's better than putting it into comments. There's no reason not to create a struct here.

crates/hyperqueue/src/dashboard/ui/screens/home/chart_section.rs
```diff
@@ -0,0 +1,132 @@
...
+
+use crate::dashboard::ui::terminal::DashboardFrame;
+use tui::layout::{Alignment, Rect};
+use tui::style::{Color, Modifier, Style};
+use tui::symbols;
+use tui::text::{Span, Spans};
+use tui::widgets::{Axis, Block, Borders, Chart, Dataset, Paragraph, Wrap};
+
+/// A new WorkerTimeState is created everytime there is a #worker change event
+#[derive(Clone)]
```
Kobzol, line 12: Remove `Clone`.

crates/hyperqueue/src/dashboard/ui/screens/home/chart_section.rs
```diff
@@ -0,0 +1,132 @@
...
+struct WorkerCountAtTime {
+    pub at_time: SystemTime,
+    pub worker_count: u32,
+}
+
+#[derive(Clone)]
```
Kobzol, line 18: Remove `Clone`.

crates/hyperqueue/src/dashboard/ui/screens/home/chart_section.rs
```diff
@@ -0,0 +1,132 @@
...
+use crate::dashboard::ui::terminal::DashboardFrame;
+use tui::layout::{Alignment, Rect};
+use tui::style::{Color, Modifier, Style};
+use tui::symbols;
+use tui::text::{Span, Spans};
+use tui::widgets::{Axis, Block, Borders, Chart, Dataset, Paragraph, Wrap};
+
+/// A new WorkerTimeState is created everytime there is a #worker change event
+#[derive(Clone)]
+struct WorkerCountAtTime {
```
Kobzol, line 13: AtTime is a bit unwieldy. Let's use some unified name for items that can be drawn on a chart. Also remove `pub`.
For example:

```rust
struct WorkerCountRecord {
    time: SystemTime,
    worker_count: u32
}
```

crates/hyperqueue/src/dashboard/ui/screens/home/chart_section.rs
```diff
@@ -0,0 +1,132 @@
...
+    pub at_time: SystemTime,
+    pub worker_count: u32,
+}
+
+#[derive(Clone)]
+pub struct ClusterOverviewChart {
+    /// The state of the worker at different times.
+    worker_count_vec: Vec<WorkerCountAtTime>,
+    /// The time for which the data is plotted for
```
Kobzol, line 22: `worker_records`

crates/hyperqueue/src/dashboard/ui/screens/home/chart_section.rs
```diff
@@ -0,0 +1,132 @@
...
+            .ok()
+            .unwrap_or(None)
+    }
+}
+
+impl Default for ClusterOverviewChart {
+    fn default() -> Self {
+        Self {
+            worker_count_vec: vec![],
+            view_size: Duration::new(30, 0),
```
Kobzol, line 122: `view_size: Duration::from_secs(30)` to make it more readable.

crates/hyperqueue/src/dashboard/ui/screens/home/chart_section.rs
```diff
@@ -0,0 +1,132 @@
...
+        }
+    }
+}
+
+fn get_time_as_secs() -> u64 {
```
Kobzol, line 127:
```rust
fn get_time_as_secs(time: SystemTime) -> u64 {
    time
        .duration_since(UNIX_EPOCH)
        .unwrap()
        .as_secs()
}
```

crates/hyperqueue/src/dashboard/ui/screens/home/chart_section.rs
```diff
@@ -0,0 +1,132 @@
...
+                .block(text_block)
+                .alignment(Alignment::Left)
+                .wrap(Wrap { trim: true });
+            frame.render_widget(paragraph, rect);
+        }
+    }
+
+    /// Returns the maximum number of workers connected in the past n seconds specified by view_size
+    fn get_max_workers_in_view(&self) -> Option<u32> {
+        SystemTime::now()
```
Kobzol, line 101: Let's use `get_time_as_secs` here, also for the `count.at_time` calculation.

crates/hyperqueue/src/dashboard/ui/screens/home/chart_section.rs
```diff
@@ -0,0 +1,132 @@
...
+            .duration_since(UNIX_EPOCH)
+            .map(|current_time| {
+                self.worker_count_vec
+                    .iter()
+                    .filter(|count| {
+                        count.at_time.duration_since(UNIX_EPOCH).unwrap().as_secs()
```
Kobzol, line 107: You can compare `SystemTime`s directly, you don't have to convert them to seconds.

crates/hyperqueue/src/dashboard/ui/screens/home/chart_section.rs
```diff
@@ -0,0 +1,132 @@
...
+    pub fn update(&mut self, overview: &CollectedOverview) {
+        self.worker_count_vec.push(WorkerCountAtTime {
+            at_time: SystemTime::now(),
+            worker_count: overview.worker_overviews.len() as u32,
+        });
+    }
+
+    pub fn draw(&mut self, rect: Rect, frame: &mut DashboardFrame) {
+        let max_workers_in_history = self.get_max_workers_in_view().unwrap_or(0);
+        if max_workers_in_history > 0 {
```
Kobzol, line 36: Fill the worker counts with zeros if there are no records in the time window.

crates/hyperqueue/src/dashboard/ui/screens/home/chart_section.rs
```diff
@@ -0,0 +1,132 @@
...
+impl ClusterOverviewChart {
+    pub fn update(&mut self, overview: &CollectedOverview) {
+        self.worker_count_vec.push(WorkerCountAtTime {
+            at_time: SystemTime::now(),
+            worker_count: overview.worker_overviews.len() as u32,
+        });
+    }
+
+    pub fn draw(&mut self, rect: Rect, frame: &mut DashboardFrame) {
+        let max_workers_in_history = self.get_max_workers_in_view().unwrap_or(0);
```
Kobzol, line 35: Don't call `get_max_workers_in_view` multiple times here, it's redundant. Also I think that it might be better if this method directly returned only the workers in the view, and you would then only put this limited number of workers into the dataset.

Kobzol (changes requested):

crates/hyperqueue/src/dashboard/ui/screens/home/chart_section.rs
vyomkeshj: Slightly confused about this one, it is 0 because of

 ```    pub fn update(&mut self, overview: &CollectedOverview) {
        self.worker_records.push(WorkerCountRecord {
            at_time: SystemTime::now(),
            worker_count: overview.worker_overviews.len() as u32,
        });
    }
```
perhaps I am missing something

File: crates/hyperqueue/src/dashboard/ui/screens/home/chart_section.rs (hunk 19 above)
Comment (vyomkeshj) [reply crates/hyperqueue/src/dashboard/ui/screens/home/chart_section.rs:35]: Let's get back to the draw only the part in the current view in the PR with cpu util graph. We anyway will have to make changes to add the feature

File: crates/hyperqueue/src/dashboard/ui/screens/home/chart_section.rs (hunk 18 above)
Comment (Kobzol) [reply crates/hyperqueue/src/dashboard/ui/screens/home/chart_section.rs:36]: The chart could draw 0 on the Y axis for time before the server was started. But it's probably useless.
In any case, this condition should be always true, so remove it and the `else` case. If there are no workers yet, just don't use an empty dataset.

File: crates/hyperqueue/src/dashboard/ui/screens/home/chart_section.rs [hunk 20]
```diff
@@ -30,84 +30,70 @@ impl ClusterOverviewChart {
...
+                        Span::from(max_workers_in_view.to_string()),
+                    ]),
+            );
+        frame.render_widget(chart, rect);
     }
 
-    /// Returns the maximum number of workers connected in the past n seconds specified by view_size
-    fn get_max_workers_in_view(&self) -> Option<u32> {
+    /// Returns the worker connection chart data for the view size
+    fn get_workers_in_view(&self) -> Vec<(u64, u64)> {
```
vyomkeshj, line 83: or should I rename it to get_data_in_view or something?

crates/hyperqueue/src/dashboard/ui/screens/home/chart_section.rs
```diff
@@ -0,0 +1,103 @@
...
+
+use crate::dashboard::ui::terminal::DashboardFrame;
+use tui::layout::Rect;
+use tui::style::{Color, Modifier, Style};
+use tui::symbols;
+use tui::text::Span;
+use tui::widgets::{Axis, Block, Borders, Chart, Dataset};
+
+/// A new WorkerTimeState is created everytime there is a #worker change event
+struct WorkerCountRecord {
```
Kobzol, line 12: Rename first field to `time` + remove `pub` + fix comment.

crates/hyperqueue/src/dashboard/ui/screens/home/chart_section.rs
```diff
@@ -0,0 +1,103 @@
...
+    pub at_time: SystemTime,
+    pub worker_count: u32,
+}
+
+pub struct ClusterOverviewChart {
+    /// The state of the worker at different times.
```
Kobzol, line 18: `/// Count of workers at different times.`

crates/hyperqueue/src/dashboard/ui/screens/home/chart_section.rs
Kobzol: It's ok like this.

crates/hyperqueue/src/dashboard/ui/screens/home/chart_section.rs
```diff
@@ -0,0 +1,103 @@
...
+                    ]),
+            );
+        frame.render_widget(chart, rect);
+    }
+
+    /// Returns the worker connection chart data for the view size
+    fn get_workers_in_view(&self) -> Vec<(u64, u64)> {
+        self.worker_records
+            .iter()
+            .filter(|count| count.at_time > SystemTime::now() - self.view_size)
```
Kobzol, line 86: Extract `SystemTime::now()` to a local variable so that it is the same for all workers.

crates/hyperqueue/src/dashboard/ui/screens/home/chart_section.rs
```diff
@@ -0,0 +1,103 @@
...
+    pub fn update(&mut self, overview: &CollectedOverview) {
+        self.worker_records.push(WorkerCountRecord {
+            at_time: SystemTime::now(),
+            worker_count: overview.worker_overviews.len() as u32,
+        });
+    }
+
+    pub fn draw(&mut self, rect: Rect, frame: &mut DashboardFrame) {
+        let workers_in_view = self.get_workers_in_view();
+        let max_workers_in_view = workers_in_view
```
Kobzol, line 34:
```rust
        let max_workers_in_view = workers_in_view
            .iter()
            .map(|&(_, count)| count)
            .max()
            .unwrap_or(0) as f64;
```

crates/hyperqueue/src/dashboard/ui/screens/home/chart_section.rs
```diff
@@ -0,0 +1,103 @@
...
+                            .add_modifier(Modifier::BOLD),
+                    ))
+                    .borders(Borders::ALL),
+            )
+            .x_axis(
+                Axis::default()
+                    .title("time ->")
+                    .style(Style::default().fg(Color::Gray))
+                    .bounds([
+                        (get_time_as_secs(SystemTime::now()) - self.view_size.as_secs()) as f64,
```
Kobzol, line 66:
```rust
get_time_as_secs(SystemTime::now() - self.view_size) as f64,
                        get_time_as_secs(SystemTime::now()) as f64,
```

crates/hyperqueue/src/dashboard/ui/screens/home/chart_section.rs
```diff
@@ -0,0 +1,103 @@
...
+                Block::default()
+                    .title(Span::styled(
+                        "Worker Connection History",
+                        Style::default()
+                            .fg(Color::White)
+                            .add_modifier(Modifier::BOLD),
+                    ))
+                    .borders(Borders::ALL),
+            )
+            .x_axis(
```
Kobzol, line 61: It would be nice to show some labels on the X axis. Let's show a tick (label) for each minute.

crates/hyperqueue/src/dashboard/ui/screens/home/chart_section.rs
```diff
@@ -0,0 +1,103 @@
...
+                    .bounds([
+                        (get_time_as_secs(SystemTime::now()) - self.view_size.as_secs()) as f64,
+                        get_time_as_secs(SystemTime::now()) as f64,
+                    ]),
+            )
+            .y_axis(
+                Axis::default()
+                    .style(Style::default().fg(Color::Gray))
+                    .bounds([0.0, max_workers_in_view])
+                    .labels(vec![
```
Kobzol, line 74: Let's create more labels than just 0 and maximum. Let's create equally spaced 10 labels between 0 and max.

Kobzol (changes requested):

PR #191, opened 2021-11-02
Change worker list filtering options from `--running` and `--offline` to `--all`

Kobzol:
Only online workers will now be shown by default. With `--all`, also offline workers will be shown.


spirali (approved):
LGTM :+1: 
(Just thinking... maybe displaying only running workers should be a default?)

Kobzol:
Then there wouldn't be a way of showing all using the current interface :/ But indeed it makes more sense.

spirali:
I know, but we can introduce instead of filter something like option "--all"

PR #194, opened 2021-11-02
Rewrite documentation

Kobzol:
- Documentation was completely rewritten to be more approachable and "sequential"
- Adds documentation versioning
  - Every `v-*` tag automatically gets a new documentation version
  - Latest commit in `main` automatically gets the `latest` version
  - Users can switch the version in the docs
- Uses `mkdocs-material` theme
- Every page has a git revision timestamp (when it was created, when it was last modified)
- HTML, CSS and JS are minified

You can see the documentation deployed [here](https://kobzol.github.io/hyperqueue/latest/).

This form of documentation can serve as a template for other future projects.

Before merging:
Clean `gh-pages` branch and run `mike set-default --push stable`


spirali (approved):
:heart:

PR #213, opened 2021-11-13
Refactor and shuffle tests around

Kobzol:
In preparation for `criterion` benchmarks, that could use some of the existing test utility functions and buidlers.


crates/tako/src/tests/utils/resources.rs
```diff
@@ -0,0 +1,57 @@
...
+        self.generic
+            .get_or_insert_with(|| vec![])
+            .push(GenericResourceRequest {
+                resource: id.into(),
+                amount,
+            });
+        self
+    }
+
+    pub fn finish(self) -> ResourceRequest {
```
spirali, line 36: Why is it called `finish` and not `build`?

spirali (approved):

crates/tako/src/tests/utils/resources.rs
Kobzol: There is already a `build` method on the builder, but it returns a `Result`, which is a bit annoying in tests. Actually I see now that the `build` method could be [renamed](https://docs.rs/derive_builder/0.10.2/derive_builder/#build-method-customization), so we could rename the default one and then we can rename this to `build`.

Usually you would have a domain struct `X` and a corresponding builder `XBuilder`. But since we use `derive_builder` and it is only used for tests, we have a domain struct `X`, then a config `XConfig` that mirrors `X` and describes the builder, and then the actual builder `XConfigBuilder`. This method serves to make the additional layer of indirection smaller.

spirali: Thanks, I understand now.

PR #215, opened 2021-11-15
Use monitoring events in dashboard

vyomkeshj:
This pull request aims to add worker message logging.

1. Worker messages for the past n minutes are stored on the server. (to be implemented)
2. The client can query the server to get either all events that the server currently has or events after a particular event_id
This will allow for the dashboard to have more information available to work with and ability to update its state through small event messages.


crates/tako/src/messages/gateway.rs
```diff
@@ -57,10 +58,16 @@ pub enum FromGatewayMessage {
...
     GetTaskInfo(TaskInfoRequest),
     ServerInfo,
+    GetMonitoringEvents(EventRequest),
     GetOverview(OverviewRequest),
     StopWorker(StopWorkerRequest),
 }
 
+#[derive(Deserialize, Serialize, Debug)]
+pub struct EventRequest {
+    pub after_id: Option<u32>,
```
Kobzol, line 68: Let's put a comment here that says how to use this attribute. And rename the struct to e.g. `MonitoringEventRequest` or just inline it, this name seems to be too generic.

crates/tako/src/server/core.rs
```diff
@@ -21,6 +22,7 @@ pub type CustomConnectionHandler = Box<dyn Fn(ConnectionDescriptor)>;
 pub struct Core {
     tasks: Map<TaskId, TaskRef>,
     workers: Map<WorkerId, Worker>,
+    event_queue: Vec<MonitoringEvent>,
```
Kobzol, line 25: Let's create some struct in `monitoring.rs` that will handle storing and fetching monitoring events. It shouldn't be the responsibility of `Core` nor the client code. `Core` will just hold this struct inside of it.

crates/tako/src/server/monitoring.rs
```diff
@@ -0,0 +1,73 @@
...
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct MonitoringEvent {
+    pub message: MonitoringEventsPayload,
+    pub event_id: u32,
+    pub event_time: SystemTime,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub enum MonitoringEventsPayload {
```
Kobzol, line 17:
```suggestion
pub enum MonitoringEventPayload {
```

crates/tako/src/server/monitoring.rs
```diff
@@ -0,0 +1,73 @@
...
+use crate::WorkerId;
+
+use serde::{Deserialize, Serialize};
+use std::time::SystemTime;
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct MonitoringEvent {
+    pub message: MonitoringEventsPayload,
+    pub event_id: u32,
+    pub event_time: SystemTime,
```
Kobzol, line 13:
```suggestion
    id: u32,
    time: SystemTime,
```

`id` and `time` should not be publicly writable. There should be some function on an event manager that will just receive the payload, and the manager will automatically assign the time and ID.

crates/tako/src/server/monitoring.rs
```diff
@@ -0,0 +1,73 @@
...
+    pub fn on_overview_response(&mut self, id: WorkerId, overview: WorkerOverview) {
+        let mut core = self.core_ref.get_mut();
+        core.insert_monitoring_event(MonitoringEvent {
+            message: MonitoringEventsPayload::OverviewReceived(id, overview),
+            event_id: self.get_last_event_id() + 1,
+            event_time: SystemTime::now(),
+        })
+    }
+
+    fn get_last_event_id(&self) -> u32 {
```
Kobzol, line 61: This should be stored inside an attribute, no need to recompute it everytime.

crates/tako/src/server/monitoring.rs
```diff
@@ -0,0 +1,73 @@
...
+    WorkerConnected(WorkerId),
+    WorkerLost(WorkerId, LostWorkerReason),
+    OverviewReceived(WorkerId, WorkerOverview),
+}
+
+pub struct ServerStateMonitor {
+    core_ref: CoreRef,
+}
+
+impl ServerStateMonitor {
```
Kobzol, line 27: There's a circular dependency between `Core` and the events, let's break it.
This struct should serve to:
1) Hold the events in memory.
2) Provide helper methods for various events that can be monitored.
It shouldn't know anything about `Core`.

Kobzol (changes requested):

crates/tako/src/worker/rpc.rs
```diff
@@ -403,6 +403,7 @@ async fn worker_message_loop(
                     .insert(msg.worker_id, msg.address)
                     .is_none())
             }
+            //todo: look
```
vyomkeshj, line 406: This commit is a bit extra undercooked, but I am putting it out for discussion before I continue.

crates/hyperqueue/src/dashboard/utils.rs
```diff
@@ -22,6 +23,21 @@ pub async fn get_hw_overview(
     response
 }
 
+pub async fn get_events_after(
+    connection: &mut ClientConnection,
+    after_id: Option<u32>,
+) -> Result<Vec<MonitoringEvent>, HqError> {
```
Kobzol: `crate::Result<Vec<MonitoringEvent>>`

crates/hyperqueue/src/dashboard/data_source.rs
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    pub async fn fetch_latest(&mut self, conn: &mut ClientConnection) -> anyhow::Result<()> {
+        let mut new_events = get_events_after(conn, Option::from(self.fetched_until)).await?;
+        self.events.append(&mut new_events);
+        Ok(())
+    }
+}
+
+impl Default for DashboardData {
```
Kobzol, line 45: You can just `#[derive(Default)]`.

crates/hyperqueue/src/dashboard/data_source.rs
```diff
@@ -0,0 +1,52 @@
...
+impl DashboardData {
+    pub fn get_cluster_state(&self) -> Option<ClusterState> {
+        //todo: generate cluster state from the events
+        //fixme: not optional
+        None
+    }
+
+    pub async fn fetch_latest(&mut self, conn: &mut ClientConnection) -> anyhow::Result<()> {
+        let mut new_events = get_events_after(conn, Option::from(self.fetched_until)).await?;
+        self.events.append(&mut new_events);
```
Kobzol, line 40: Don't forget to update `self.fetched_until` after downloading the events.

crates/hyperqueue/src/dashboard/data_source.rs
```diff
@@ -0,0 +1,52 @@
...
+
+impl DashboardData {
+    pub fn get_cluster_state(&self) -> Option<ClusterState> {
+        //todo: generate cluster state from the events
+        //fixme: not optional
+        None
+    }
+
+    pub async fn fetch_latest(&mut self, conn: &mut ClientConnection) -> anyhow::Result<()> {
+        let mut new_events = get_events_after(conn, Option::from(self.fetched_until)).await?;
```
Kobzol, line 39: `Some(self.fetched_until)`

crates/tako/src/server/rpc.rs
```diff
@@ -168,6 +170,19 @@ async fn worker_rpc_loop(
     let snd_loop =
         forward_queue_to_sealed_sink(queue_receiver, connection.sender, connection.sealer);
 
+    let overview_event_loop = async move {
```
Kobzol, line 173: This RPC loop runs for every worker, so like this you would periodically download overviews for all workers, inside each worker's RPC loop.

crates/tako/src/server/monitoring.rs
```diff
@@ -0,0 +1,66 @@
...
+
+pub struct EventHistory {
+    event_queue: Vec<MonitoringEvent>,
+    last_event_id: u32,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct MonitoringEvent {
+    pub message: MonitoringEventPayload,
+    pub id: u32,
```
Kobzol, line 17: Remove `pub` from fields.

crates/tako/src/server/monitoring.rs
```diff
@@ -0,0 +1,66 @@
...
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub enum MonitoringEventPayload {
+    WorkerConnected(WorkerId, WorkerConfiguration),
+    WorkerLost(WorkerId, LostWorkerReason),
+    OverviewUpdate(CollectedOverview),
+}
+
+impl EventHistory {
+    pub fn get_events(&self) -> &Vec<MonitoringEvent> {
```
Kobzol, line 29: It's slightly cleaner to not expose implementation details and just return `&[MonitoringEvent]`.

crates/tako/src/server/monitoring.rs
```diff
@@ -0,0 +1,66 @@
...
+        &self.event_queue
+    }
+
+    pub fn on_worker_added(&mut self, id: WorkerId, configuration: WorkerConfiguration) {
+        self.event_queue.push(MonitoringEvent {
```
Kobzol, line 34: `self.event_queue.push`, assigning the `id` and `time` is duplicated in all `on_...` methods. Please create a method for this, which will only receive the payload.

Kobzol (changes requested):

crates/hyperqueue/src/dashboard/data_source.rs
```diff
@@ -29,24 +30,33 @@ pub struct DashboardData {
 }
 
 impl DashboardData {
```
vyomkeshj, line 31: To be ignored for now

crates/tako/src/server/monitoring.rs
vyomkeshj: And access them with methods? isn't it extra code since the methods are just getters?

crates/tako/src/worker/rpc.rs
```diff
@@ -203,6 +207,11 @@ pub async fn run_worker(
         }
     };
 
+    let overview_update_interval = match overview_update_interval {
+        None => Either::Left(futures::future::pending()),
```
vyomkeshj, line 211: How so I make it explicit, why does it break even though the ```hw_polling_process``` part looks exactly the same?

crates/tako/src/worker/rpc.rs
```diff
@@ -457,6 +443,47 @@ pub async fn connection_initiator(
...
+    state.send_message_to_server(message);
+}
+
+async fn update_overview(state: WorkerStateRef, overview_update_interval: Duration) {
+    let mut update_interval = tokio::time::interval(overview_update_interval);
+    loop {
+        update_interval.tick().await;
+        send_overview_to_server(
+            state.get_mut(),
+            OverviewRequest {
```
vyomkeshj, line 480: I'll fix these in a later commit, this commit is just to ask the question about the futures::pending confusion

crates/tako/src/server/monitoring.rs
Kobzol: It is extra code, but we will deny modifying the data from the outside, the event should be read-only. Also if all attributes are `pub`, you can create an instance of the struct from outside this module, which we also don't want to allow.

crates/hyperqueue/src/dashboard/data_source.rs
```diff
@@ -0,0 +1,58 @@
...
+        //     .last()
+        //     .unwrap()
+        //     .clone();
+
+        None
+    }
+
+    pub async fn fetch_latest(&mut self, conn: &mut ClientConnection) -> anyhow::Result<()> {
+        let mut new_events = get_events_after(conn, Some(self.fetched_until)).await?;
+        self.events.append(&mut new_events);
```
Kobzol, line 50: The `append` call will reset `new_events`, so after it the vector will be empty. Also, if there are no new events, you shouldn't reset `fetched_until` to zero.

```rust
self.fetched_until = new_events
    .iter()
    .map(|event| event.id)
    .max()
    .unwrap_or(self.fetched_until);
self.events.append(&mut new_events);
```

Kobzol (changes requested):

crates/hyperqueue/src/dashboard/data_source.rs
```diff
@@ -0,0 +1,110 @@
...
+#[derive(Default)]
+pub struct DashboardData {
+    /// The event_id until which the data has already been fetched for
+    fetched_until: u32,
+    /// All events received from the client
+    events: Vec<MonitoringEvent>,
+}
+
+impl DashboardData {
+    pub fn get_cluster_state(&self) -> Option<ClusterState> {
```
vyomkeshj, line 35: I'm going to clean the closures

Kobzol:
There are two modes of event data usage in dashboard:
1) Real-time: data is fetched incrementally from the server every e.g. 1 sec.
When the dashboard starts, the server may have already forgotten some data, so the dashboard has to be bootstrapped from some recent snapshot of the whole cluster. Otherwise the dashboard wouldn't know e.g. how many workers are currently connected (as the events only contain incremental diff data).
    
    See https://github.com/It4innovations/hyperqueue/pull/215/files#diff-961b2cb56a83ff35fa86e3b0677c639411c95f706d36cc66da5d2ea3deb5bb57R26

2) Post-mortem: all events are loaded from a file. (not implemented yet).
Since there will be no live bootstrap available, the file should either store all events or contain some "snapshot" events that will contain the whole state of the cluster.

CC @spirali for comments.

PR #219, opened 2021-11-22
Remove `TaskRef` from worker

Kobzol:
`TaskRef` is replaced by `TaskId`. The only place where `TaskRef` was left is dependency/download, since that will be rewritten soon.


crates/tako/src/worker/taskmap.rs
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    #[inline]
+    pub fn find_mut(&mut self, task_id: TaskId) -> Option<&mut Task> {
+        self.map.get_mut(&task_id)
+    }
+
+    #[inline]
+    pub fn insert(&mut self, task_id: TaskId, task: Task) {
+        self.map.insert(task_id, task);
```
spirali, line 34: Assert here that task not already here (assert that result is_none())

spirali (approved):

PR #221, opened 2021-11-23
Remove self reference from worker state

Kobzol:
Removes the annoying self reference from worker state.


crates/tako/src/worker/pool.rs
```diff
@@ -44,7 +44,7 @@ impl GenericResourcePool {
         }
     }
 
-    pub fn release_allocation(&mut self, allocation: GenericResourceAllocationValue) {
+    pub fn release_allocation(&mut self, allocation: &GenericResourceAllocationValue) {
```
spirali, line 47: Why referenec was added?

It was on purpose moved so it cannot be accidentally removed twice.

PR #227, opened 2021-12-01
Allow specifying autoalloc walltime using HMS format

Kobzol:


CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+# DEV
+
+## Changes
```
spirali, line 3: It is not "change" as it backward compatible so it is a new feature.

spirali (changes requested):

CHANGELOG.md
Kobzol: Fixed.

spirali (approved):

PR #232, opened 2021-12-08
Initial Snakemake benchmark integration

User3574:
Added snakemake into benchmarks with SleepSnake example using Descriptors.

### DIFF
--- a/benchmarks/src/environment/snake.py
+++ b/benchmarks/src/environment/snake.py
@@ -5,7 +5,7 @@ import logging
 import os
 
 from pathlib import Path
-from snakemake import snakemake
+# from snakemake import snakemake
 from . import Environment, EnvironmentDescriptor
 from .utils import EnvStateManager
 
@@ -32,7 +32,7 @@ class SnakeEnvironmentDescriptor(EnvironmentDescriptor):
 
 class SnakeEnvironment(Environment, EnvStateManager):
     def __init__(self, info: SnakeClusterInfo):
-        super(EnvStateManager, self).__init__()
+        EnvStateManager.__init__(self)
         self.info = info
         self.snakefile = os.path.join(info.workdir, "Snakefile")
 
--- a/benchmarks/src/environment/snake.py
+++ b/benchmarks/src/environment/snake.py
@@ -2,10 +2,8 @@ from typing import Dict, Any
 
 import dataclasses
 import logging
-import os
 
 from pathlib import Path
-# from snakemake import snakemake
 from . import Environment, EnvironmentDescriptor
 from .utils import EnvStateManager
 
@@ -34,7 +32,7 @@ class SnakeEnvironment(Environment, EnvStateManager):
     def __init__(self, info: SnakeClusterInfo):
         EnvStateManager.__init__(self)
         self.info = info
-        self.snakefile = os.path.join(info.workdir, "Snakefile")
+        self.snakefile = info.workdir / "Snakefile"
 
     def __enter__(self):
         self.start()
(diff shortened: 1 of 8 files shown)


benchmarks/src/environment/snake.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def metadata(self) -> Dict[str, Any]:
+        return {}
+
+
+class SnakeEnvironment(Environment):
+    def __init__(self, info: SnakeClusterInfo):
+        self.info = info
+        self.snakefile = "Snakefile"
+        self.state = "initial"
```
Kobzol, line 36: The state checking will be needed in most environments, I added a mixin class for it (`EnvStateManager`) in `main`. Please use it here.

benchmarks/src/environment/snake.py
```diff
@@ -0,0 +1,68 @@
...
+        return {}
+
+    def metadata(self) -> Dict[str, Any]:
+        return {}
+
+
+class SnakeEnvironment(Environment):
+    def __init__(self, info: SnakeClusterInfo):
+        self.info = info
+        self.snakefile = "Snakefile"
```
Kobzol, line 35: The Snakefile should be located in `info.workdir`.

benchmarks/src/environment/snake.py
```diff
@@ -0,0 +1,68 @@
...
+        self.stop()
+
+    @property
+    def workdir(self) -> Path:
+        return self.info.workdir
+
+    def start(self):
+        logging.info("Creating Snakefile")
+        file = open(self.snakefile, "w")
+        file.close()
```
Kobzol, line 52:
```suggestion
        with open(self.snakefile, "w") as f:
            pass
```
But if it's empty, it's useless to create it here.

benchmarks/src/environment/snake.py
```diff
@@ -0,0 +1,68 @@
...
+        self.state = "started"
+
+    def stop(self):
+        logging.info("Deleting Snakefile")
+        os.remove(self.snakefile)
```
Kobzol, line 57: Let's not delete it so that the user can check what was submitted into SnakeMake.

benchmarks/src/environment/snake.py
```diff
@@ -0,0 +1,68 @@
...
+        self.state = "stopped"
+
+    def submit(self, args: List[str]):
```
Kobzol, line 60: What is in `args`? This should probably be multiple parameters, one into Snakefile and the other into the CLI command.

BTW does SnakeMake have a Python API that could be used here?

benchmarks/src/environment/snake.py
```diff
@@ -0,0 +1,68 @@
...
+        logging.info(f"Snake Submitting {args[0], args[1]}")
+        file = open(self.snakefile, "a")
+        file.writelines(args[0])
+        file.close()
+
+        result = subprocess.call([f'snakemake --quiet --cores {args[1]}'], shell=True, stdout=subprocess.PIPE)
```
Kobzol, line 66: We definitely do not want `shell=True` here. Please check the `execute_process` function, which handles process errors and stores `stdout`/`stderr` to disk.

Kobzol (changes requested):

benchmarks/src/environment/snake.py
```diff
@@ -0,0 +1,64 @@
+from typing import Dict, Any
+
+import dataclasses
+import logging
+import os
+
+from pathlib import Path
+# from snakemake import snakemake
```
Kobzol, line 8: The user that creates the benchmark should not be forced to have `snakemake` installed. Therefore snakemake should only be imported when the benchmark is actually being executed. I suggest to import it locally in the `submit` method.

benchmarks/src/environment/snake.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def metadata(self) -> Dict[str, Any]:
+        return {}
+
+
+class SnakeEnvironment(Environment, EnvStateManager):
+    def __init__(self, info: SnakeClusterInfo):
+        EnvStateManager.__init__(self)
+        self.info = info
+        self.snakefile = os.path.join(info.workdir, "Snakefile")
```
Kobzol, line 37:
```suggestion
        self.snakefile = info.workdir / "Snakefile"
```

benchmarks/src/environment/snake.py
```diff
@@ -0,0 +1,64 @@
...
+
+    @property
+    def workdir(self) -> Path:
+        return self.info.workdir
+
+    def start(self):
+        self.state_start()
+        logging.info("Creating Snakefile")
+        with open(self.snakefile, "w") as f:
+            pass
```
Kobzol, line 54: The `start` and `stop` don't really have to do anything right now, is that correct? In that case, just call `state_start` and `state_stop` inside of them. Don't log anything nor create an empty snakefile.

Kobzol (changes requested):

Kobzol:
I made some slights edits, otherwise LGTM. The next step would be to add support for multiple worker nodes.

PR #237, opened 2021-12-09
Deps & Sharing configuration

spirali:


crates/tako/src/messages/worker.rs
```diff
@@ -35,10 +36,13 @@ pub struct ComputeTaskMsg {
...
 
-    pub configuration: TaskConfiguration,
-
     pub user_priority: Priority,
     pub scheduler_priority: Priority,
+
+    pub resources: crate::common::resources::ResourceRequest,
+    pub time_limit: Option<Duration>,
+    pub n_outputs: u32,
+    pub body: Vec<u8>,
```
Kobzol, line 45: `#[serde(with = "serde_bytes")]`?

crates/tako/src/server/task.rs
```diff
@@ -58,18 +59,53 @@ bitflags::bitflags! {
     }
 }
 
+pub struct TaskInput {
+    task: TaskRef,
+    output_id: u32, // MAX = pure dependency on task, not real output id
```
Kobzol, line 64: Let's use an enum for this, instead of some magic MAX value. We can use `NonZero` for the output ID to avoid increasing the size of the enum.

crates/tako/src/tests/test_reactor.rs
```diff
@@ -1042,3 +1041,23 @@ fn cancel_tasks<T: Into<TaskId> + Copy>(core: &mut Core, task_ids: &[T]) {
...
+
+#[test]
+fn test_task_deps() {
+    let mut core = Core::default();
+    //create_test_workers(&mut core, &[1, 1, 1]);
+    submit_example_3(&mut core);
+    assert_eq!(core.get_read_to_assign().len(), 2);
+    create_test_workers(&mut core, &[1]);
+    start_and_finish_on_worker(&mut core, 2, 100, 0);
+    assert!(core.get_task_by_id_or_panic(3.into()).get().is_waiting());
```
Kobzol, line 1053: Let's introduce helper methods like `assert_waiting([3, 4, 6])` and `assert_ready([5])`. It's PITA to change tests that have some method call repeated many times when the method changes.

crates/tako/src/server/task.rs
spirali: The problem is that 0 is valid output_id so I need something like NonMax if I do not want some "+1/-1" magic. I know it seems 1quite a premature optimization but I expect a lot of TaskInputs.

Kobzol: What about something like this? https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=2f37ca86e5a006308703f8010a55ff85

spirali: Oookay. I somehow still like my max solution, but you are right :)

Kobzol:
Formatter is not happy. You can run `./scripts/check.sh` before pushing to make (sure that) everything is OK.

Kobzol (approved):

PR #258, opened 2021-12-13
Bound event list

vyomkeshj:
This PR puts an upper bound on the number of events stored by the tako server in its core.

### DIFF
--- a/crates/hyperqueue/src/bin/hq.rs
+++ b/crates/hyperqueue/src/bin/hq.rs
@@ -126,6 +126,10 @@ struct ServerStartOpts {
     /// Port for worker connections
     #[clap(long)]
     worker_port: Option<u16>,
+
+    /// The maximum number of events tako server will store in memory
+    #[clap(long, default_value = "1000000")]
+    event_store_size: usize,
 }
 
 #[derive(Parser)]
@@ -276,6 +280,7 @@ async fn command_server_start(
         autoalloc_interval: opts.autoalloc_interval.map(|x| x.unpack()),
         client_port: opts.client_port,
         worker_port: opts.worker_port,
+        event_store_size: opts.event_store_size,
     };
 
     init_hq_server(&gsettings, server_cfg).await
--- a/crates/hyperqueue/src/server/rpc.rs
+++ b/crates/hyperqueue/src/server/rpc.rs
@@ -56,6 +56,7 @@ impl Backend {
         key: Arc<SecretKey>,
         idle_timeout: Option<Duration>,
         worker_port: Option<u16>,
+        event_store_size: usize,
     ) -> crate::Result<(Backend, impl Future<Output = crate::Result<()>>)> {
         let msd = Duration::from_millis(20);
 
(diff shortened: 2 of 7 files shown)


crates/tako/src/server/monitoring.rs
```diff
@@ -30,6 +31,7 @@ pub enum MonitoringEventPayload {
 
 // Keep the size of the event structure in check
 const _: () = assert!(std::mem::size_of::<MonitoringEventPayload>() == 32);
+const MAX_EVENTS: usize = 10_1000;
```
Kobzol, line 34: I suggest to set this to `10_000_000`, i.e. 10 million events. Currently, the event record takes `32` bytes, so this should amount to ~300 MiB. @spirali thoughts?

spirali: This probably should not be a build in but rather a configurable option. I would vote for 100k-1M as default. 10M seems to me as much for a default value.

Kobzol: Ok, sounds reasonable. @vyomkeshj please add a CLI parameter to `hq server start` that will configure this, with `1M` as a default value. We can discuss it on MM if you have any questions.

vyomkeshj: on it

crates/hyperqueue/src/server/rpc.rs
```diff
@@ -56,6 +56,7 @@ impl Backend {
         key: Arc<SecretKey>,
         idle_timeout: Option<Duration>,
         worker_port: Option<u16>,
+        event_store_size: Option<u32>,
```
Kobzol, line 59: Let's make this `usize` and move the responsibility of choosing the default to the CLI.

crates/tako/src/server/monitoring.rs
```diff
@@ -4,11 +4,13 @@ use crate::WorkerId;
 
 use crate::messages::common::WorkerConfiguration;
 use serde::{Deserialize, Serialize};
+use std::collections::VecDeque;
 use std::time::SystemTime;
 
 #[derive(Default)]
```
Kobzol, line 10: Remove the `Default` derive, storage with event size `0` doesn't make sense.

crates/tako/src/server/monitoring.rs
```diff
@@ -32,6 +34,13 @@ pub enum MonitoringEventPayload {
 const _: () = assert!(std::mem::size_of::<MonitoringEventPayload>() == 32);
 
 impl EventStorage {
+    pub fn new(with_size: u32) -> Self {
```
Kobzol, line 37:
```suggestion
    pub fn new(event_store_size: usize) -> Self {
```

crates/hyperqueue/src/bin/hq.rs
```diff
@@ -126,6 +126,10 @@ struct ServerStartOpts {
     /// Port for worker connections
     #[clap(long)]
     worker_port: Option<u16>,
+
+    /// The maximum number of events tako server will store
+    #[clap(long)]
+    event_store_size: Option<u32>,
```
Kobzol, line 132:
```suggestion

    /// The maximum number of monitoring events the server will store in memory
    #[clap(long, default_value = "1000000")]
    event_buffer_size: usize,
```

Kobzol (changes requested):

Kobzol:
Thanks!

PR #260, opened 2021-12-13
Disabling HyperThreading: Worker start option --cpus="no-ht"

### LINKED ISSUE #244: Is hyperthreading automatically detected?
On one of our machines, we have 128 physical cores on each node, but with hyperthreading. `hq` detects 256 CPUs (8x32). However, some of our applications are CPU-intensive and are actually slowed down when using 256 MPI processes, and need to be run on 128 maximum. 
Is there a way for `hq` to detect hyperthreading, and a way to send jobs and say "I don't want to use hyperthreading"? In particular, I want to be able to submit 128-core jobs to `hq` and guarantee that it only puts 1 per node, and not 2 per node at the same time.

spirali:
Closes #244

### DIFF
--- a/crates/hyperqueue/src/bin/hq.rs
+++ b/crates/hyperqueue/src/bin/hq.rs
@@ -32,7 +32,7 @@ use hyperqueue::server::bootstrap::{
 use hyperqueue::transfer::messages::{
     FromClientMessage, JobInfoRequest, Selector, ToClientMessage,
 };
-use hyperqueue::worker::hwdetect::{detect_cpus, detect_generic_resource};
+use hyperqueue::worker::hwdetect::{detect_cpus, detect_cpus_no_ht, detect_generic_resource};
 use hyperqueue::worker::start::{start_hq_worker, WorkerStartOpts};
 use hyperqueue::WorkerId;
 use tako::common::resources::ResourceDescriptor;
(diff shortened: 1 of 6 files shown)


docs/jobs/cresources.md
```diff
@@ -107,13 +107,17 @@ Worker automatically detect number of CPUs and on Linux system it also detects p
...
 ``--cpus`` parameter; for example as follows:
 
 - 8 CPUs for worker
   ``$ hq worker start --cpus=8``
 
 - 2 sockets with 12 cores per socket
-  ``$ hq worker --cpus=2x12``
+  ``$ hq worker start --cpus=2x12``
+
+- Automatic detection of CPUs but ignores HyperThreading (it will detect only the first virtual CPU of each physical CPU)
```
Kobzol, line 122: `(hyper)-thread of each physical core`

crates/hyperqueue/src/bin/hq.rs
```diff
@@ -227,7 +227,7 @@ struct WorkerInfoOpts {
 #[derive(Parser)]
 struct HwDetectOpts {
     #[clap(long)]
-    no_ht: bool,
+    no_hyperthreading: bool,
```
Kobzol, line 230: Please also add a doc comment here to avoid an undocumented CLI parameter.

Kobzol (changes requested):

Kobzol (approved):

PR #264, opened 2021-12-16
Added test for event sourcing

vyomkeshj:


crates/tako/src/tests/integration/mod.rs
```diff
@@ -1,4 +1,5 @@
 mod test_basic;
+mod test_event_sourcing;
```
Kobzol, line 2: Move this to `test_monitoring.rs` please.

crates/tako/src/tests/integration/test_event_sourcing.rs
```diff
@@ -0,0 +1,43 @@
+use super::utils::worker::WorkerConfigBuilder;
+use crate::tests::integration::utils::api::get_events_after;
+use crate::tests::integration::utils::server::run_test;
+use crate::tests::integration::utils::worker::WorkerHandle;
+use std::time::Duration;
+
+#[tokio::test]
+async fn test_event_sourcing() {
```
Kobzol, line 8: Let's create several smaller tests that check specific scenarios. Having one big test is bad for debugging, if something fails it's not clear immediately what is wrong.

You can test scenarios like:
1) Check that a worker added event was created when a worker was connected
2) Check that a worker removed event was created when a worker was disconnected
3) Check that ID filtering works, i.e. that you really only get events after the specified ID.

Also please check specific event values, don't just check the number of events.

Kobzol (changes requested):

PR #266, opened 2021-12-16
Get rid of Task refs

Kobzol:
Micro-benchmarks suffered quite a lot, I will look at them in a follow-up PR if I can find any improvements. Macro-benchmarks (sleep) did not show any notable regressions.

Best reviewed commit by commit :)


crates/tako/src/server/core.rs
```diff
@@ -237,47 +250,64 @@ impl Core {
...
-        assert!(!task.has_consumers());
-        assert!(self.tasks.remove(&task.id).is_some());
-        std::mem::replace(&mut task.state, TaskRuntimeState::Released)
+    pub fn remove_task(&mut self, task_id: TaskId) -> TaskRuntimeState {
+        trace_task_remove(task_id);
+
+        let state = {
+            let task = self.tasks.get_task_mut(task_id);
+            assert!(!task.has_consumers());
+            std::mem::replace(&mut task.state, TaskRuntimeState::Released)
```
spirali, line 268: Maybe TaskRuntimeState::Released can be completely removed now, as it was mainly a marker that "you have a reference on a task that is removed"

Old state can be safely taken from removed without replacing

crates/tako/src/server/core.rs
```diff
@@ -74,6 +74,16 @@ impl CoreRef {
 }
 
 impl Core {
+    #[inline]
+    pub fn split_tasks_workers(&self) -> (&TaskMap, &WorkerMap) {
```
spirali, line 78: Is it a good name? At least in Tokio "split" methods takes "self".

crates/tako/src/scheduler/state.rs
```diff
@@ -152,55 +151,59 @@ impl SchedulerState {
...
         }
 
-        for (worker_id, mut task_refs) in task_computes {
-            task_refs.sort_unstable_by_key(|tr| {
-                let t = tr.get();
-                Reverse((t.configuration.user_priority, t.scheduler_priority))
+        for (worker_id, mut task_ids) in task_computes {
+            task_ids.sort_unstable_by_key(|&task_id| {
+                let task = core.get_task(task_id);
+                Reverse((task.configuration.user_priority, task.scheduler_priority))
```
spirali, line 157: Maybe sort_by_cached_key to avoid repeated lookup?

spirali (changes requested):

crates/tako/benches/benchmarks/scheduler.rs
```diff
@@ -23,17 +21,11 @@ fn bench_b_level(c: &mut BenchmarkGroup<WallTime>) {
...
-                            })
-                            .collect();
-                        (core, tasks)
+                        add_tasks(&mut core, task_count);
+                        core
                     },
-                    |(_core, tasks)| {
-                        compute_b_level_metric(tasks);
+                    |core| {
+                        compute_b_level_metric(core.task_map_mut());
```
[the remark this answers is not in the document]
Kobzol: Hmm, that's true, but since the Core is originally empty, both codes compute the B-level for the same amount of tasks ,right?

crates/tako/src/server/core.rs
Kobzol: Good idea.

crates/tako/src/server/core.rs
Kobzol: Do you have an idea for a better name? The `split` represents a "splitting borrow" (https://doc.rust-lang.org/nomicon/borrow-splitting.html).

crates/tako/src/scheduler/state.rs
Kobzol: Good idea. In the `"scheduler/schedule/tasks=100000, workers=32"` micro-benchmark it seems to be ~5% faster.

crates/tako/benches/benchmarks/scheduler.rs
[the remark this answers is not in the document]
spirali: What if you submit subsequently several independent submissions?

crates/tako/src/server/core.rs
spirali: I do not have:)

spirali:
LGTM :+1: 
[Maybe some squashes?]

spirali (approved):

Kobzol:
:tada: :tada: :tada:

PR #284, opened 2022-01-07
Removed "Removed" task state + fix spawn/remove race

spirali:


crates/tako/src/worker/state.rs
```diff
@@ -250,12 +252,9 @@ impl WorkerState {
                 self.schedule_task_start();
                 self.ready_task_queue.release_allocation(allocation);
             }
-            TaskState::Removed => {
-                unreachable!();
-            }
         }
 
-        assert!(self.tasks.remove(&task_id).is_some());
+        //assert!(self.tasks.remove(&task_id).is_some());
```
Kobzol, line 257: Just remove it please.

Kobzol (commented):
Good! I suppose that we cannot reliably test the race situation?

spirali:
> Good! I suppose that we cannot reliably test the race situation?

AFAIK no

Kobzol (approved):

PR #287, opened 2022-01-07
Remove most trace events

Kobzol:
The tracing infrastructure is a left-over from RSDS. I removed most of the methods, since they will be stored inside the event log instead. I still kept some of the infrastructure because it might be useful for logging profiling information in the future.

@vyomkeshj The removed trace events should be eventually moved into monitoring events.


spirali (approved):
LGTM (There will be just conflict with the recent race/spawn race fix)

spirali (approved):

PR #289, opened 2022-01-10
Introduce task description

Kobzol:
This PR refactors job submission. Task-specific information is now stored and passed in a way that should simplify creating arbitrary task graphs per job in the future.

`TaskConf` should also be unified with `TaskDescription` somehow (does not have to be in this PR).


spirali (approved):
LGTM, but squash it please. It helps the review but for browsing it quite a fine grained

PR #295, opened 2022-01-12
Rename Python binding package from pyhq to hyperqueue

Kobzol:
The Rust package of the binding name still remains named `pyhq` to avoid clash with the (library and binary) Rust package `hyperqueue`.


Kobzol:
I found a way to keep `pyhq` Rust package while using `hyperqueue` wheel and Python module. @spirali

Issue #286, opened 2022-01-07
Add command to read stdout/stderr of tasks of a job


Kobzol:
Add a new command (called e.g. `hq job cat`) that would read the output (stdout/stderr) of task(s) of the given job and print it to stdout. Only tasks that are running or have finished should be outputted, waiting tasks do not have any output yet.

- Allow selecting all tasks of a job or a specific task list (using [selectors](https://it4innovations.github.io/hyperqueue/stable/tips/cli-shortcuts/#id-selector))
- Allow parametrizing whether to output `stdout` or `stderr`
- Handle error situations (output might be disabled or streamed into a log, path might not exist)
- Allow filtering tasks by status (e.g. only display the output of successfully completed tasks)
- There could be multiple tasks in the job. Add an optional flag `--task-header` that will print a header line before starting the output of a task. For example:

  - With `--task-header`:
    ```
    # Task 1
    task1 line 0
    task1 line 1
    # Task 3
    task3 line 0
    task3 line 1
    ```
  - Without `--task-header`:
    ```
    task1 line 0
    task1 line 1
    task3 line 0
    task3 line 1
    ```

Potential interface:
```bash
$ hq job cat [stdout/stderr] <job-id> --tasks <task-selector> --task-status <task status1, task status2, ...> --task-header
```

Kobzol:
The basic functionality was implemented in https://github.com/It4innovations/hyperqueue/pull/308. What's left is to implement task headers and task status selector.

PR #308, opened 2022-01-14
Add `hq job cat` command for printing job stdout/stderr

User3574:
Initial version of HQ Job cat
Issue: #286

### DIFF
--- a/crates/hyperqueue/src/client/commands/job.rs
+++ b/crates/hyperqueue/src/client/commands/job.rs
@@ -1,5 +1,6 @@
 use crate::client::globalsettings::GlobalSettings;
 use crate::client::job::get_worker_map;
+use crate::client::output::outputs::OutputStream;
 use crate::client::status::{job_status, Status};
 use crate::common::arraydef::IntArray;
 use crate::common::cli::SelectorArg;
@@ -34,6 +35,20 @@ pub struct JobTasksOpts {
 pub struct JobCancelOpts {
     /// Select job(s) to cancel
     pub selector_arg: SelectorArg,
+}
+
+#[derive(Parser)]
+pub struct JobCatOpts {
+    /// Select job
+    pub job_id: u32,
+
+    /// Select task(s) outputs to view
+    #[clap(long)]
+    pub tasks: Option<SelectorArg>,
+
+    /// Type of output stream to display
+    #[clap(possible_values = &["stdout", "stderr"])]
+    pub stream: OutputStream,
 }
 
 pub async fn get_last_job_id(connection: &mut ClientConnection) -> crate::Result<Option<JobId>> {
@@ -121,6 +136,30 @@ pub async fn output_job_tasks(
             .print_job_tasks(job, get_worker_map(connection).await?);
     } else {
         log::error!("Job {} not found", job_id);
+    }
+    Ok(())
+}
+
+pub async fn output_job_cat(
+    gsettings: &GlobalSettings,
+    connection: &mut ClientConnection,
+    job_id: u32,
+    task_selector: Option<Selector>,
+    output_stream: OutputStream,
+) -> anyhow::Result<()> {
+    let message = FromClientMessage::JobDetail(JobDetailRequest {
+        selector: Selector::Specific(IntArray::from_id(job_id)),
+        include_tasks: true,
+    });
+    let mut responses =
+        rpc_call!(connection, message, ToClientMessage::JobDetailResponse(r) => r).await?;
+
+    if let Some(job) = responses.pop().and_then(|v| v.1) {
+        return gsettings
+            .printer()
+            .print_job_output(job, task_selector, output_stream);
+    } else {
+        log::error!("Job {job_id} not found");
     }
     Ok(())
 }
--- a/crates/hyperqueue/src/client/output/cli.rs
+++ b/crates/hyperqueue/src/client/output/cli.rs
@@ -4,7 +4,7 @@ use cli_table::{print_stdout, Cell, CellStruct, Color, ColorChoice, Style, Table
 use std::fmt::{Display, Write};
 
 use crate::client::job::WorkerMap;
-use crate::client::output::outputs::{Output, MAX_DISPLAYED_WORKERS};
+use crate::client::output::outputs::{Output, OutputStream, MAX_DISPLAYED_WORKERS};
 use crate::client::status::{job_status, task_status, Status};
 use crate::common::env::is_hq_env;
 use crate::common::format::{human_duration, human_size};
@@ -16,10 +16,10 @@ use crate::server::autoalloc::{
 use crate::server::job::{JobTaskCounters, JobTaskInfo, JobTaskState, StartedTaskData};
 use crate::stream::reader::logfile::Summary;
 use crate::transfer::messages::{
-    AutoAllocListResponse, JobDescription, JobDetail, JobInfo, StatsResponse, TaskDescription,
-    WaitForJobsResponse, WorkerExitInfo, WorkerInfo,
+    AutoAllocListResponse, JobDescription, JobDetail, JobInfo, Selector, StatsResponse,
+    TaskDescription, WaitForJobsResponse, WorkerExitInfo, WorkerInfo,
 };
-use crate::{JobTaskCount, WorkerId};
+use crate::{JobTaskCount, JobTaskId, WorkerId};
 
 use chrono::{DateTime, Local, SubsecRound, Utc};
 use core::time::Duration;
@@ -37,7 +37,8 @@ use crate::worker::start::WORKER_EXTRA_PROCESS_PID;
 use anyhow::Error;
 use colored::Color as Colorization;
 use colored::Colorize;
-use std::collections::BTreeSet;
+use std::collections::{BTreeSet, HashMap};
+use std::fs::File;
 use tako::messages::gateway::{LostWorkerReason, ResourceRequest};
 
 pub const TASK_COLOR_CANCELED: Colorization = Colorization::Magenta;
@@ -485,7 +486,7 @@ impl Output for CliOutput {
             .iter()
             .map(|task| {
                 let (start, end) = get_task_time(&task.state);
-                let (cwd, stdout, stderr) = get_task_paths(&task_to_paths, task);
+                let (cwd, stdout, stderr) = format_task_paths(&task_to_paths, task);
 
                 vec![
                     task.task_id.cell().justify(Justify::Right),
@@ -558,6 +559,15 @@ impl Output for CliOutput {
             humantime::format_duration(duration),
             msgs.join(", ")
         );
+    }
+
+    fn print_job_output(
+        &self,
+        job: JobDetail,
+        task_selector: Option<Selector>,
+        output_stream: OutputStream,
+    ) -> anyhow::Result<()> {
+        print_job_output(job, task_selector, output_stream)
     }
 
     fn print_summary(&self, filename: &Path, summary: Summary) {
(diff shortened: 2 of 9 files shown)


crates/hyperqueue/src/client/commands/job.rs
```diff
@@ -32,6 +32,22 @@ pub struct JobCancelOpts {
...
+#[derive(Parser)]
+pub struct JobCatOpts {
+    /// Select job(s) to display
+    pub job_selector_arg: SelectorArg,
+
+    /// Select task(s) of a job to display
+    pub task_selector_arg: SelectorArg,
+
+    /// Display stdout, stderr
+    #[clap(long)]
```
Kobzol, line 44: Let's not combine `stdout`/`stderr` to make it simpler (otherwise we would have to distinguish between them). As stated in the issue, let the user select either `stdout` or `stderr` using a subcommand.

crates/hyperqueue/src/client/commands/job.rs
```diff
@@ -32,6 +32,22 @@ pub struct JobCancelOpts {
     pub selector_arg: SelectorArg,
 }
 
+#[derive(Parser)]
+pub struct JobCatOpts {
+    /// Select job(s) to display
+    pub job_selector_arg: SelectorArg,
```
Kobzol, line 38: To keep it simple, let's support only a single job for now. Selecting task IDs across jobs would be quite weird anyway :)

crates/hyperqueue/src/client/commands/job.rs
```diff
@@ -102,6 +118,35 @@ pub async fn output_job_detail(
...
+
+    for response in responses {
+        if let Some(job) = response.1 {
+            gsettings.printer().print_job_cat(
+                job,
+                stdout,
+                stderr,
+            );
+        } else {
+            log::error!("Job {} not found", response.0);
```
Kobzol, line 144: This should return an error instead of just logging. Once only a single job is supported, it should be easy.

crates/hyperqueue/src/client/output/cli.rs
```diff
@@ -534,6 +535,66 @@ impl Output for CliOutput {
         );
     }
 
+    fn print_job_cat(&self, job: JobDetail, stdout: bool, stderr: bool) {
```
Kobzol, line 538: Let's rename this method to `print_job_output` and create an enum, e.g. `OutputStream` that will contain `Stdout` and `Stderr` variants.

Also please make this method return a `Result` and return an error from the JSON printer. It should not support this output, because the data might be binary and it's not clear how to put it into JSON. And if someone uses the JSON output mode, they can process the files themselves.

The quiet mode should have the same implementation as CLI. Either reuse some function that will be called from both Quiet and CLI or just put the implementation as a default impl into the trait.

crates/hyperqueue/src/client/output/cli.rs
```diff
@@ -534,6 +535,66 @@ impl Output for CliOutput {
...
+        let JobDescription::Array { ids, .. } = &job_desc;
+        let (resources, program_def, time_limit, priority, pin) = match &job_desc {
+            JobDescription::Array { task_desc, .. } => (
+                &task_desc.resources,
+                &task_desc.program,
+                task_desc.time_limit,
+                task_desc.priority,
+                task_desc.pin,
+            ),
+        };
```
Kobzol, line 557: Just access stdout/stderr of the selected tasks, no need for this large amount of code :)
Don't use the stdout/stderr of the job, it might contain unresolved paths with placeholders.

crates/hyperqueue/src/client/output/cli.rs
```diff
@@ -534,6 +535,66 @@ impl Output for CliOutput {
...
+            return;
+        }
+
+        for id in ids.iter() {
+            println!("# Task {}", id);
+
+            if stdout {
+                println!("# Stdout");
+                let str_path = stdout_path.unwrap().to_str().unwrap().replace("%{TASK_ID}", &id.to_string());
+                let stream = fs::read_to_string(str_path)
```
Kobzol, line 583: The data might be binary, so we cannot operate with `str`/`String`. Also we cannot afford to load the data into memory, it might be larger than RAM. We need to stream the file to `stdout` effectively. You can use [`io::copy`](https://static.rust-lang.org/doc/master/std/io/fn.copy.html) for this.

crates/hyperqueue/src/client/output/cli.rs
```diff
@@ -534,6 +535,66 @@ impl Output for CliOutput {
...
+            StdioDef::File(file) => Some(file),
+            _ => None,
+        };
+        if stderr_path.is_none() && stderr {
+            log::error!("Job doesn't have stderr task files, consider submitting job with --stderr");
+            return;
+        }
+
+        for id in ids.iter() {
+            println!("# Task {}", id);
```
Kobzol, line 578: Until something like `--task-header` is implemented, just don't output any metadata.

Kobzol (changes requested):

crates/hyperqueue/src/client/output/cli.rs
```diff
@@ -534,6 +536,53 @@ impl Output for CliOutput {
...
+
+            let file = File::open(stdin).expect("Unable to open file");
+            let mut reader = BufReader::new(file);
+            for line in reader.lines() {
+                let line = line.expect("Unable to read line");
+                println!("{}", line);
+            }
+        }
+
+        let JobDetail {
```
Kobzol, line 555: Only `tasks` are needed, so there's no need to destructure `JobDetail` here. Just access `job.tasks`. Destructuring is used in places where we want to be notified when a new attribute is added to some struct.

crates/hyperqueue/src/client/output/cli.rs
```diff
@@ -534,6 +536,53 @@ impl Output for CliOutput {
...
+        fn read_stream(task_state: &JobTaskState, output_stream: &OutputStream) {
+            let (_, stdout, stderr) = get_task_paths(task_state);
+            let stdin = match output_stream {
+                OutputStream::Stdout => {stdout}
+                OutputStream::Stderr => {stderr}
+            };
+
+            let file = File::open(stdin).expect("Unable to open file");
+            let mut reader = BufReader::new(file);
+            for line in reader.lines() {
```
Kobzol, line 549: Please use `io::copy` here to make this more efficient and support binary data.

Kobzol (changes requested):

crates/hyperqueue/src/client/output/cli.rs
```diff
@@ -536,49 +541,52 @@ impl Output for CliOutput {
         );
     }
 
-    fn print_job_output(&self, job: JobDetail, task_selector: &Selector, output_stream: &OutputStream) -> anyhow::Result<()> {
+    fn print_job_output(
+        &self,
+        job: JobDetail,
+        task_selector: &Selector,
+        output_stream: &OutputStream,
```
Kobzol, line 548: Let's just pass `OutputStream` and `Selector` by value, no need for references here.

crates/hyperqueue/src/client/output/cli.rs
```diff
@@ -536,49 +541,52 @@ impl Output for CliOutput {
...
                     read_stream(&task.state, output_stream);
                 });
             }
             Selector::Specific(arr) => {
-                arr.iter().for_each(|id| {
-                    read_stream(&tasks[id.to_usize()].state, output_stream);
-                });
+                let tasks: Vec<(u32, Option<&JobTaskInfo>)> = arr
+                    .iter()
+                    .map(|id| (id, job.tasks.get(id.to_usize())))
```
Kobzol, line 574: I wasn't clear enough I think last time, sorry :D You can't use the user's task ID as an index into tasks, there is no correspondence between the index of the vector and task IDs. You have to build a map from `task ID` to `JobTaskInfo` and then use the user's task IDs as keys into that map.

crates/hyperqueue/src/client/output/cli.rs
```diff
@@ -536,49 +541,52 @@ impl Output for CliOutput {
...
+                    .collect();
+                for (id, task) in tasks {
+                    if task.is_none() {
+                        log::error!("Task {} not found", id);
+                        return Ok(());
```
Kobzol, line 579: If a task is not found, let's just continue the loop (and print a warning, not an error).

crates/hyperqueue/src/client/output/quiet.rs
```diff
@@ -63,7 +66,12 @@ impl Output for Quiet {
...
     fn print_job_detail(&self, _job: JobDetail, _show_tasks: bool, _worker_map: WorkerMap) {}
     fn print_job_wait(&self, _duration: Duration, _response: &WaitForJobsResponse) {}
-    fn print_job_output(&self, _job: JobDetail, _task_selector: &Selector, _output_stream: &OutputStream) -> anyhow::Result<()> {
+    fn print_job_output(
+        &self,
+        _job: JobDetail,
+        _task_selector: &Selector,
+        _output_stream: &OutputStream,
+    ) -> anyhow::Result<()> {
         Ok(())
```
Kobzol, line 75: This should output the same thing as the CLI output.

crates/hyperqueue/src/client/output/cli.rs
```diff
@@ -536,49 +541,52 @@ impl Output for CliOutput {
...
-            let mut reader = BufReader::new(file);
-            for line in reader.lines() {
-                let line = line.expect("Unable to read line");
-                println!("{}", line);
-            }
+            match File::open(stdin) {
+                Ok(mut file) => {
+                    std::io::copy(&mut file, &mut std::io::stdout());
+                }
+                Err(_) => log::warn!("File: {} cannot be opened", stdin),
```
Kobzol, line 561:
```suggestion
                Err(error) => log::warn!("File {} cannot be opened: {:?}", stdin, error),
```
Let's give more context to the user.

crates/hyperqueue/src/client/output/cli.rs
```diff
@@ -536,49 +541,52 @@ impl Output for CliOutput {
...
-    fn print_job_output(&self, job: JobDetail, task_selector: &Selector, output_stream: &OutputStream) -> anyhow::Result<()> {
+    fn print_job_output(
+        &self,
+        job: JobDetail,
+        task_selector: &Selector,
+        output_stream: &OutputStream,
+    ) -> anyhow::Result<()> {
         fn read_stream(task_state: &JobTaskState, output_stream: &OutputStream) {
             let (_, stdout, stderr) = get_task_paths(task_state);
             let stdin = match output_stream {
```
Kobzol, line 552: I suggest to rename it to e.g. `path`, `stdin` is quite misleading here.

crates/hyperqueue/src/client/commands/job.rs
```diff
@@ -36,11 +36,14 @@ pub struct JobCancelOpts {
...
 #[derive(Parser)]
 pub struct JobCatOpts {
+    /// Select job
     pub job_id: u32,
 
+    /// Select task(s) outputs to view
     pub task_selector_arg: SelectorArg,
 
     #[clap(long)]
+    /// Type of output(stdout, stderr) to display
```
Kobzol, line 46:
```suggestion
    /// Type of output to display
    #[clap(possible_values = &["stdout", "stderr"])]
```
But as mentioned earlier, I would prefer this to be a subcommand, i.e. `$ hq job cat [stdout/stderr] <cat-options>`.

Kobzol (changes requested):

crates/hyperqueue/src/client/commands/job.rs
```diff
@@ -1,13 +1,52 @@
...
+}
+
+#[derive(Parser)]
+pub struct JobCatOpts {
+    /// Select job
+    pub job_id: u32,
+
+    /// Select task(s) outputs to view
+    #[clap(long)]
+    pub tasks: SelectorArg,
```
Kobzol, line 44: This should be optional (if not specified, all tasks should be outputted).
`pub tasks: Option<SelectorArg>`

crates/hyperqueue/src/client/output/outputs.rs
```diff
@@ -33,6 +35,14 @@ impl FromStr for Outputs {
     }
 }
 
+#[derive(Parser, Clone, Copy)]
```
Kobzol, line 38: It's not super important, but this does not need to be `Clone` nor `Copy`, let's save some compile time by removing these derives.

crates/hyperqueue/src/client/output/cli.rs
```diff
@@ -847,6 +857,67 @@ pub(crate) fn job_progress_bar(
...
+        let path = match output_stream {
+            OutputStream::Stdout => stdout,
+            OutputStream::Stderr => stderr,
+        };
+
+        match File::open(path) {
+            Ok(mut file) => {
+                let copy = std::io::copy(&mut file, &mut std::io::stdout().lock());
+                if let Err(error) = copy {
+                    log::warn!("File: {} cannot be copied: {:?}", path, error);
```
Kobzol, line 876: Since recently, you can use identifiers in formatting placeholders directly :)
`File: {path} cannot be copied: {error:?}`

Kobzol (changes requested):

Kobzol:
Rebased over `main` to fix conflicts. Please add tests and we can merge it :)

tests/test_job.py
```diff
@@ -833,3 +833,22 @@ def test_job_shell_script_read_interpreter(hq_env: HqEnv):
     wait_for_job_state(hq_env, 1, "FINISHED")
 
     check_file_contents("out.txt", "Hello\n")
+
+
+def test_job_output(hq_env: HqEnv):
```
Kobzol, line 838: Please also test `stderr` output.

Kobzol (changes requested):

tests/test_job.py
```diff
@@ -833,3 +833,48 @@ def test_job_shell_script_read_interpreter(hq_env: HqEnv):
...
+    output = hq_env.command(["job", "cat", "--tasks", "2", "1", "stdout"])
+    assert output.strip() == "hello"
+
+    output = hq_env.command(["job", "cat", "--tasks", "0", "1", "stdout"])
+    assert "Task 0 not found" in output.strip()
+
+    output = hq_env.command(["job", "cat", "1", "stderr"])
+    assert output == ""
+
+    hq_env.command(["submit", "--array", "1-2", "sleep"])
```
Kobzol, line 856: Please write some string explicitly to stderr, either using Python (you can look at other tests how it's done) or Bash (https://stackoverflow.com/questions/2990414/echo-that-outputs-to-stderr).

On another system, sleep could output a different message and this test would fail.

Kobzol (changes requested):

Kobzol:
Thanks @User3574! Great PR.

PR #313, opened 2022-01-15
Enforce that all CLI commands and flags are documented (have help string)

Kobzol:


spirali (approved):

spirali:
Please update changelog

Issue #318, opened 2022-01-15
horizontal separator lines in hq submit output


goxeq:
The horizontal separator lines in `hq submit` output make the upper part of the table to scroll away from the screen.  My terminal is usually 24 lines high (default Putty behaviour). If the horizontal lines would be removed, the text would be easier to read, because it would fit in one screen.

```
+-----------------+---------------------------------+
| Id              | 114                             |
+-----------------+---------------------------------+
| Name            | a                               |
+-----------------+---------------------------------+
| State           | SUBMITTED                       |
+-----------------+---------------------------------+
| Tasks           | 1                               |
+-----------------+---------------------------------+
| Workers         |                                 |
+-----------------+---------------------------------+
| Resources       | cpus: 1 compact                 |
+-----------------+---------------------------------+
| Priority        | 0                               |
+-----------------+---------------------------------+
| Command         | a                               |
+-----------------+---------------------------------+
| Stdout          | job-%{JOB_ID}/%{TASK_ID}.stdout |
+-----------------+---------------------------------+
| Stderr          | job-%{JOB_ID}/%{TASK_ID}.stderr |
+-----------------+---------------------------------+
| Environment     |                                 |
+-----------------+---------------------------------+
| Working Dir     | %{SUBMIT_DIR}                   |
+-----------------+---------------------------------+
| Submission date | 2022-01-15 20:13:11 UTC         |
+-----------------+---------------------------------+
| Task time limit | None                            |
+-----------------+---------------------------------+
| Makespan        | 0s                              |
+-----------------+---------------------------------+
```
versus 
```
+-----------------+---------------------------------+
| Id              | 114                             |
| Name            | a                               |
| State           | SUBMITTED                       |
| Tasks           | 1                               |
| Workers         |                                 |
| Resources       | cpus: 1 compact                 |
| Priority        | 0                               |
| Command         | a                               |
| Stdout          | job-%{JOB_ID}/%{TASK_ID}.stdout |
| Stderr          | job-%{JOB_ID}/%{TASK_ID}.stderr |
| Environment     |                                 |
| Working Dir     | %{SUBMIT_DIR}                   |
| Submission date | 2022-01-15 20:13:11 UTC         |
| Task time limit | None                            |
| Makespan        | 0s                              |
+-----------------+---------------------------------+
```

Kobzol:
Some of the rows can actually be multiline (for example command when you input multiple arguments). That by itself is not ideal, but we probably don't have a better way for now.

[...]

PR #319, opened 2022-01-15
Make CLI tables more dense

Kobzol:
Removes horizontal separators between table cells to make tables more dense.

Before (vertical table):
image

After (vertical table):
image

Before (horizontal table):
image

After (horizontal table):
image

Vertical and horizontal tables are now also handled separately, both in CLI output code and in tests. Autoalloc tables that were previously incorrectly vertical are now correctly horizontal.

Fixes: https://github.com/It4innovations/hyperqueue/issues/318


spirali (approved):
Changelog please

Issue #311, opened 2022-01-15
Produce a better error message when running a script without ./ or absolute path


goxeq:
When I create a script and I try to submit it like `hq submit script.sh`, the job fails without giving any reasonable error, just 

```
Wait finished in 8ms 286us 671ns: 1 job failed
Some jobs have failed or have been canceled"`
```
After some experimentation I found out that it has to be submitted as `hq submit ./script.sh` 

Could this requirement to include `./ ` be dropped? Or at least give some better error message.

Kobzol:
Hi! This is actually a bit more subtle. I'll first explain why we can't easily prepend `./` to the command and then talk about potential solutions.

HyperQueue will try to execute the command that you pass to it exactly as you would have executed it in your shell. If you just write `script.sh` in your shell, it will most likely also not work.

If we e.g. prepended `./` to the command, or turned the first submitted command to an absolute path, it could change the meaning of the executed command.

For example, let's assume that we submit `hq submit program` from directory `/foo`, set the working directory to `/bar` and we have `program` located at `/baz` in `$PATH`.

- By default, when a worker executes this task, `program` will be looked up in `$PATH`, so `/baz/program` gets executed.
- If we prepended `./`, the worker would instead try to execute `/bar/program`.
- If we made the path absolute (relative to the submit directory), the worker would try to execute `/foo/program`.

We can't foresee the user's intentions, so the safest way is not to preprocess the command in any way.

However, there are two things that are not ideal in the current state:
- The produced error is quite misleading and frankly just bad, we will definitely improve it.
- Similarly to PBS, the user might actually want to submit a script containing directives, which should be immediately parsed by HyperQueue, instead of being lazily executed. In this situation `script.sh` should be fine, because it will be just some path resolved by HyperQueue at the moment of submission. We also plan to implement this (https://github.com/It4innovations/hyperqueue/issues/6).

Kobzol:
We have altered the error message slightly when a command fails to execute:
```
Before: IO error: No such file or directory (os error 2)
After: Error: Cannot execute "test.sh": No such file or directory (os error 2)
```
Is this error message be clearer for you? Or is there more information that you would like to see here?

goxeq:
The most helpful error message would be:

`Error: Command not found: "test.sh"` optionally, if the file is in working directory and is executable, followed by `Did you mean "./test.sh?"`

Also, where do these error messages go? I don't see them in the command output. I would expect them to go to stderr of the job, but that is empty.

[...]

PR #320, opened 2022-01-16
Improve error displayed when the executed command of a task is not found

Kobzol:
Execution error:
```
Error: Cannot execute "test.sh": <some error>
```

Execution error when the program was not found:
```
Error: Cannot execute "test.sh": No such file or directory (os error 2)
The program that you have tried to execute (test.sh) was not found.
```

Execution error when the program was not found, but it exists in the working directory of the task:
```
Error: Cannot execute "test.sh": No such file or directory (os error 2)
The program that you have tried to execute (test.sh) was not found.
The file "/projects/it4i/hyperqueue/test.sh" exists, maybe you have meant `./test.sh` instead?
```

CC @goxeq

Fixes: https://github.com/It4innovations/hyperqueue/issues/311

### DIFF
--- a/crates/hyperqueue/src/worker/start.rs
+++ b/crates/hyperqueue/src/worker/start.rs
@@ -1,5 +1,9 @@
+use std::ffi::OsStr;
+use std::fmt::Write;
 use std::io;
-use std::path::PathBuf;
+use std::io::ErrorKind;
+use std::os::unix::prelude::OsStrExt;
+use std::path::{Path, PathBuf};
 use std::process::ExitStatus;
 use std::rc::Rc;
 use std::time::Duration;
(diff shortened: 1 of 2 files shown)


crates/hyperqueue/src/worker/start.rs
```diff
@@ -225,6 +229,47 @@ async fn run_task(
...
+            let file = &program.args[0];
+            let mut msg = format!(
+                "\nThe program that you have tried to execute (`{}`) was not found.",
+                file
+            );
+
+            let path = Path::new(OsStr::from_bytes(file.as_bytes()));
+            if !path.is_absolute() && !path.starts_with(".") {
+                let possible_path = program.cwd.as_ref().unwrap().join(path);
+                if possible_path.exists() {
```
spirali, line 245: Maybe ``is_file`` is better than ``exists``?

Kobzol: Good point! We could also check that it's executable, but I'm already a bit afraid of FS latency here, since it's synchronous, so let's keep it at `is_file` for now.

spirali (approved):

PR #325, opened 2022-01-17
HQ directives

### LINKED ISSUE #6: Implement #HQ directives
Implement equivalent to #PBS directives in shell scripts

spirali:
Closes #6

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,12 @@
 
 ## New features
 
+
+### Job configurations
+   * ``#HQ`` directives - Job arguments can be defined inside a submitted script, e.g.
+     ``#HQ --cpus=4``.
+
+
 ### Worker configuration
   * You can now select what should happen when a worker loses its connection to the server using the
     new `--on-worker-lost` flag available for `worker start` and `hq alloc add` commands. You can find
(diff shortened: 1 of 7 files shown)


crates/hyperqueue/src/client/commands/submit.rs
```diff
@@ -309,7 +386,7 @@ pub async fn submit_computation(
         program: program_def,
         resources,
         pin,
-        priority,
+        priority: priority.unwrap_or(0),
```
Kobzol, line 389: Maybe we could introduce a separate function that will take the opts and fill in all defaults, so that it is on one place? Or at least we could fill the defaults on a single spot at the beginning of this function.

crates/hyperqueue/src/client/commands/directives.rs
```diff
@@ -0,0 +1,146 @@
+use crate::client::commands::submit::SubmitJobConfOpts;
```
Kobzol, line 1: I would put this file into the parent directory (`client`), or create a new directory `commands/submit` and put it there. The files in this directory should more or less correspond to individual command (or command groups).

Kobzol (approved):

PR #331, opened 2022-01-19
Dashboard: Added worker info table

vyomkeshj:
This PR aims to reintroduce worker_info_table to the dashboard.

### DIFF
--- a/crates/hyperqueue/src/dashboard/ui/screens/home/screen.rs
+++ b/crates/hyperqueue/src/dashboard/ui/screens/home/screen.rs
@@ -8,11 +8,13 @@ use crate::dashboard::ui::terminal::DashboardFrame;
 use crate::dashboard::ui::widgets::text::draw_text;
 
 use crate::dashboard::data::DashboardData;
+use crate::dashboard::ui::screens::home::worker_info_table::WorkerInfoTable;
 use tui::layout::{Constraint, Direction, Layout, Rect};
 
 #[derive(Default)]
 pub struct HomeScreen {
     worker_util_table: WorkerUtilTable,
+    worker_info_table: WorkerInfoTable,
     cluster_overview: ClusterOverviewChart,
 }
 
@@ -23,11 +25,14 @@ impl Screen for HomeScreen {
 
         self.cluster_overview.draw(layout.chart_chunk, frame);
         self.worker_util_table.draw(layout.body_chunk, frame);
+        self.worker_info_table.draw(layout.info_chunk, frame);
     }
 
     fn update(&mut self, data: &DashboardData) {
         self.worker_util_table.update(data);
         self.cluster_overview.update(data);
+        self.worker_info_table
+            .update(data, self.worker_util_table.get_selected_item());
     }
 
     /// Handles key presses for the components of the screen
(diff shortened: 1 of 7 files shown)


crates/hyperqueue/src/dashboard/ui/screens/home/screen.rs
```diff
@@ -23,18 +25,26 @@ impl Screen for HomeScreen {
...
     }
 
     /// Handles key presses for the components of the screen
     fn handle_key(&mut self, key: Key) {
         match key {
-            Key::Down => self.worker_util_table.select_next_worker(),
-            Key::Up => self.worker_util_table.select_previous_worker(),
+            Key::Down => {
+                self.worker_util_table.select_next_worker();
+                self.worker_info_table.change_info(self.worker_util_table.get_selected_item());
```
Kobzol, line 42: I would rather put the `change_info` call into `HomeScreen::update`. Or add `worker_id: Option<WorkerId>` as a parameter to `WorkerInfoTable::update`.

crates/hyperqueue/src/dashboard/ui/screens/home/worker_info_table.rs
```diff
@@ -0,0 +1,110 @@
...
+
+#[derive(Default)]
+pub struct WorkerInfoTable {
+    table: StatefulTable<WorkerInfoDataRow>,
+    current_worker_id: Option<WorkerId>,
+}
+
+#[derive(Default, Debug)]
+struct WorkerInfoDataRow {
+    pub label: String,
```
Kobzol, line 17:
```suggestion
    pub label: &'static str,
```
No need for this to be a heap-allocated `String`, since the labels are constant strings.

crates/hyperqueue/src/dashboard/ui/screens/home/worker_info_table.rs
```diff
@@ -0,0 +1,110 @@
...
+            TableColumnHeaders {
+                title: "Worker Info".to_string(),
+                inline_help: "".to_string(),
+                table_headers: None,
+                column_widths: vec![Constraint::Percentage(30), Constraint::Percentage(70)],
+            },
+            |data| {
+                Row::new(vec![
+                    Cell::from(data.label.to_string()),
+                    Cell::from(data.data.to_string()),
```
Kobzol, line 50:
```suggestion
                    Cell::from(data.data.clone()),
```
Again, we want to convey that we clone the string, not that we convert a string into a string :)

Kobzol (changes requested):

crates/hyperqueue/src/dashboard/ui/screens/home/worker_info_table.rs
```diff
@@ -0,0 +1,102 @@
...
+        self.table.draw(
+            rect,
+            frame,
+            TableColumnHeaders {
+                title: "Worker Info".to_string(),
+                inline_help: "".to_string(),
+                table_headers: None,
+                column_widths: vec![Constraint::Percentage(30), Constraint::Percentage(70)],
+            },
+            |data| Row::new(vec![Cell::from(data.label), Cell::from(data.data.clone())]),
```
Kobzol, line 44:
```suggestion
            |data| Row::new(vec![Cell::from(data.label), Cell::from(data.data.as_str())]),
```
Now we can avoid the clone :)

Issue #327, opened 2022-01-18
Make task list less width


spirali:
The task list table is so width that is no usable. 

![image](https://user-images.githubusercontent.com/263349/149905300-bcd007be-295b-4d54-bbc1-401eb1d6a762.png)

Lets make it again usable on ~100 column with terminal (if paths has a reasonable size).

Proposed modifications:
* Do not show start/end times by default (only when ``--show-times`` are used). And then show it in the same column "Time" like:
   ```
      999ms
      Start: xxxxxxxxxxx
      End: yyyyyyyyyyy
   ```
* Show paths in one column as:
  ```
  cwd: /path/xxxxxxxxxxxxxx
  out: /path/yyyyyyyyyyyyyyy
  err: /path/zzzzzzzzzzzzzzzz
  ```

PR #335, opened 2022-01-20
Make task table more narrow (dense) by putting times and paths into a single column

Kobzol:
This PR reduces the width of `hq job tasks` table. It also changes the format of printed times to be shorter.

Before:
image

After:
image

We should also clip the errors somehow probably, because these can be arbitrarily long.

Fixes: https://github.com/It4innovations/hyperqueue/issues/327

### DIFF
--- a/crates/hyperqueue/src/client/output/cli.rs
+++ b/crates/hyperqueue/src/client/output/cli.rs
@@ -124,11 +124,11 @@ impl CliOutput {
         ]);
         rows.push(vec![
             "Stdout".cell().bold(true),
-            stdio_to_cell(&program.stdout),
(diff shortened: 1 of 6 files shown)


crates/hyperqueue/src/client/output/cli.rs
```diff
@@ -519,12 +527,8 @@ impl Output for CliOutput {
             "Task ID".cell().bold(true),
```
spirali, line 527: Maybe just "ID" to save some columns in smaller jobs?

Kobzol: Good idea, I shortened it to just `ID`.

spirali (approved):

PR #341, opened 2022-01-21
Attaching stdin to tasks

### LINKED ISSUE #300: Allow reading `submit` command from `stdin`
It would be nice to allow something like `hq submit --input < commands.txt` to read the command to be executed from e.g. a file.

spirali:
Closes #300


tests/test_directives.py
```diff
@@ -77,7 +77,7 @@ def test_hq_directives_mode_always(hq_env: HqEnv):
     Path("program").write_text(content)
 
     hq_env.command(
-        ["submit", "--directives", "always", "program"],
+        ["submit", "--directives", "file", "program"],
```
Kobzol, line 80: Rename test (always -> file) and maybe add a test for `--directives input`?

docs/jobs/jobs.md
```diff
@@ -279,6 +279,19 @@ There are three ways of waiting until a job completes:
...
         ```
 
+## Attaching stdio
+
+When ``--stdio`` flag is used, then HQ capture stdio and attach it to each task of a job.
+This can be used to submitting scripts  without creating file.
+The following command will capture stdin and executes it in Bash 
+
+    ```bash
+    $ hq submit --stdio bash
```
Kobzol, line 289: This should be `--stdin` and not `--stdio`, right? Same in changelog.

Kobzol (changes requested):

Kobzol (approved):

PR #342, opened 2022-01-24
Event JSON export

Kobzol:
Adds a command that can export evens from a log file into line-delimited JSON (one event per line).


crates/hyperqueue/src/bin/hq.rs
```diff
@@ -111,6 +112,7 @@ enum SubCommand {
     /// Automatic allocation management
     #[clap(name = "alloc")]
     AutoAlloc(AutoAllocOpts),
+    Event(EventOpts),
```
spirali, line 115: Maybe call it ``Events`` or ``EventLog`` ? (as you are working with a more than one event

spirali (approved):

Issue #245, opened 2021-12-11
Change `hq job list` to only show active jobs by default


Kobzol:
Details can be found in https://github.com/It4innovations/hyperqueue/issues/241.

PR #347, opened 2022-01-27
Change default behaviour of `hq job list`

Kobzol:
Now only queued and running jobs are shown by default in `hq job list`.

Fixes: https://github.com/It4innovations/hyperqueue/issues/245

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -97,6 +97,11 @@ individual tasks of a job, please use the new `hq job tasks <job-id>` command.
     # Now: submits a single task that executes `foo --array 1-4`
     ```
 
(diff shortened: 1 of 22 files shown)


spirali (commented):
Btw: In readme we speak about "$ hq job list" as the only command for inspection. Just for the sake of "getting started" if user submit invalid task/too fast command he will se an empty table. 

One option can be something like show statistics under the table (something like "10 tasks in other states"?).  Or maybe just add a sentence about "hq job list --all" in the readme.

Kobzol:
Good idea. I changed the command in the Readme and also included a hint in `hq job list` when some jobs are filtered.

spirali (approved):

PR #349, opened 2022-01-27
Added Worker Overview Screen and Screen Switching

vyomkeshj:
This PR aims to add WorkerOverviewScreen to the dashboard along with a screen switcher.

1. Press `right arrow` when a worker is selected to go to its worker overview screen.
2. If the worker is lost, the screen changes back to ClusterOverview.


crates/hyperqueue/src/dashboard/state.rs
```diff
@@ -1,28 +1,49 @@
...
+use crate::dashboard::ui::screen::{DashboardScreenState, Screen, ScreenController};
+use crate::dashboard::ui::screens::home::screen::ClusterOverviewScreen;
+use crate::dashboard::ui::screens::worker::screen::WorkerOverviewScreen;
+use std::ops::DerefMut;
 use tako::common::WrappedRcRefCell;
 
 pub struct DashboardState {
-    screens: Vec<Box<dyn Screen>>,
-    current_screen: usize, // TODO: usize newtype wrapper or enum
+    cluster_overview_screen: Box<ClusterOverviewScreen>,
```
Kobzol, line 9: There's probably currently no need to box the screens if we're not keeping them inside a `Vec`.

crates/hyperqueue/src/dashboard/state.rs
```diff
@@ -1,28 +1,49 @@
...
+                    let screen = self.worker_info_screen.deref_mut();
+                    screen.set_worker_id(*worker_id);
+                    return (screen, &mut self.controller);
+                }
+                (home_screen, &mut self.controller)
+            }
+        }
+    }
+
+    pub fn switch_screen(&mut self, to_screen: DashboardScreenState) {
```
Kobzol, line 46: This method shouldn't be here, it's currently unused anyway.

crates/hyperqueue/src/dashboard/state.rs
```diff
@@ -1,28 +1,49 @@
...
 
-    pub fn get_current_screen_mut(&mut self) -> &mut dyn Screen {
-        self.screens[self.current_screen].as_mut()
+    pub fn get_current_screen_and_switcher(&mut self) -> (&mut dyn Screen, &mut ScreenController) {
+        let home_screen = self.cluster_overview_screen.deref_mut();
+
+        match self.controller.get_current_screen() {
+            DashboardScreenState::ClusterOverviewScreen => (home_screen, &mut self.controller),
+            DashboardScreenState::WorkerOverviewScreen(worker_id) => {
+                if let Some(worker_id) = worker_id {
```
Kobzol, line 36: This should be done only when a screen is being switched, not when getting the screen and switcher, that should be a read-only operation.

crates/hyperqueue/src/dashboard/ui/screen.rs
```diff
@@ -1,10 +1,47 @@
...
+    fn handle_key(&mut self, key: Key, controller: &mut ScreenController);
+}
+
+#[derive(Clone, Copy)]
+pub enum DashboardScreenState {
+    ClusterOverviewScreen,
+    WorkerOverviewScreen(Option<WorkerId>),
+}
+
+pub struct ScreenController {
```
Kobzol, line 19: Please turn `screen.rs` into a module and put the controller and the state into a separate file `controller.rs`.

crates/hyperqueue/src/dashboard/state.rs
```diff
@@ -1,28 +1,49 @@
...
     }
 
     pub fn get_data_source(&self) -> &WrappedRcRefCell<DashboardData> {
         &self.data_source
     }
 
-    pub fn get_current_screen_mut(&mut self) -> &mut dyn Screen {
-        self.screens[self.current_screen].as_mut()
+    pub fn get_current_screen_and_switcher(&mut self) -> (&mut dyn Screen, &mut ScreenController) {
+        let home_screen = self.cluster_overview_screen.deref_mut();
```
Kobzol, line 31: When we have screens as separate attributes of the state, and we want to preserve their state, in this function you should match on the enum contained within the controller, and return the corresponding screen.

crates/hyperqueue/src/dashboard/ui/screen.rs
```diff
@@ -1,10 +1,47 @@
...
-    fn update(&mut self, data: &DashboardData);
-    fn handle_key(&mut self, key: Key);
+    fn update(&mut self, data: &DashboardData, controller: &mut ScreenController);
+    fn handle_key(&mut self, key: Key, controller: &mut ScreenController);
+}
+
+#[derive(Clone, Copy)]
+pub enum DashboardScreenState {
+    ClusterOverviewScreen,
+    WorkerOverviewScreen(Option<WorkerId>),
```
Kobzol, line 16: The individual screens already hold their inner state, so I don't think that it's correct to duplicate that state here. It would be useful if we recreated the screens from scratch everytime, but we decided that we don't want to do that. So this should be just a simple enum.

crates/hyperqueue/src/dashboard/ui/screens/home/screen.rs
```diff
@@ -1,45 +1,47 @@
...
     }
 
     /// Handles key presses for the components of the screen
-    fn handle_key(&mut self, key: Key) {
+    fn handle_key(&mut self, key: Key, controller: &mut ScreenController) {
         match key {
             Key::Down => self.worker_util_table.select_next_worker(),
             Key::Up => self.worker_util_table.select_previous_worker(),
+            Key::Right => {
+                change_to_worker_overview_screen(
```
Kobzol, line 40: This would be more natural as a method on `ScreenController` I think. Also the method should receive worker ID, not option (you shouldn't be able to switch to the screen without providing the ID). Therefore it should only be enabled if some ID (worker) is selected.

Kobzol (changes requested):

crates/hyperqueue/src/dashboard/ui/screen/controller.rs
```diff
@@ -0,0 +1,37 @@
+use crate::dashboard::events::{DashboardEvent, ScreenChangeEvt};
+use tako::WorkerId;
+use tokio::sync::mpsc::UnboundedSender;
+
+#[derive(Clone, Copy)]
+pub enum DashboardScreenState {
```
Kobzol, line 6: Let's rename this to `ChangeScreenCommand`, this should represent an action that changes the screen.

crates/hyperqueue/src/dashboard/ui/screen/controller.rs
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    pub fn show_cluster_overview(&mut self) {
+        self.send_change_event(ScreenChangeEvt::create(
+            DashboardScreenState::ClusterOverviewScreen,
+        ))
+    }
+
+    fn send_change_event(&mut self, event: ScreenChangeEvt) {
+        if let Err(err) = self.event_channel.send(DashboardEvent::ScreenChange(event)) {
```
Kobzol, line 33: Let's use `expect` and crash the app if the send fails. It's a non-recoverable situation.

crates/hyperqueue/src/dashboard/events.rs
```diff
@@ -6,4 +7,18 @@ pub enum DashboardEvent {
     KeyPressEvent(Key),
     /// Updates the dashboard ui with the latest data
     UiTick,
+    /// Changes the current screen
+    ScreenChange(ScreenChangeEvt),
+}
+
+#[derive(Clone, Copy)]
+pub struct ScreenChangeEvt(DashboardScreenState);
```
Kobzol, line 15: Use `ChangeScreenCommand` enum instead of this.

crates/hyperqueue/src/dashboard/ui_loop.rs
```diff
@@ -80,9 +84,9 @@ fn draw(state: &mut DashboardState, terminal: &mut DashboardTerminal) {
     terminal
         .draw(|frame| {
             let data = state.get_data_source().clone();
-            let screen = state.get_current_screen_mut();
+            let (screen, switcher) = state.get_current_screen_and_switcher();
```
Kobzol, line 87: `switcher` -> `controller`

crates/hyperqueue/src/dashboard/state.rs
```diff
@@ -1,28 +1,51 @@
...
 
 pub struct DashboardState {
-    screens: Vec<Box<dyn Screen>>,
-    current_screen: usize, // TODO: usize newtype wrapper or enum
+    cluster_overview_screen: ClusterOverviewScreen,
+    worker_overview_screen: WorkerOverviewScreen,
+
     data_source: WrappedRcRefCell<DashboardData>,
+
+    current_screen: DashboardScreenState,
```
Kobzol, line 14: Let's create the `DashboardScreenState` enum here. It should be a simple enum with fieldless variants, it simply says what screen is active, without holding any other state. It's something different than a command to change a screen (even though it looks similar).

We could also represent the screen state by a vector of screens and an index that would remember which state is active. But for now this design is OK.

crates/hyperqueue/src/dashboard/state.rs
```diff
@@ -1,28 +1,51 @@
...
 
     pub fn get_data_source(&self) -> &WrappedRcRefCell<DashboardData> {
         &self.data_source
     }
 
-    pub fn get_current_screen_mut(&mut self) -> &mut dyn Screen {
-        self.screens[self.current_screen].as_mut()
+    /// Updates data for the next screen if required and changes the dashboard state to show it.
+    pub fn change_current_screen(&mut self, next_screen: DashboardScreenState) {
+        if let DashboardScreenState::WorkerOverviewScreen(worker_id) = next_screen {
```
Kobzol, line 35: Please use a `match` here (without the `_` pattern), so that if new commands to change screen are added, we will have to add support for them here explicitly.

crates/hyperqueue/src/dashboard/ui/screens/worker/screen.rs
```diff
@@ -0,0 +1,129 @@
...
+use crate::dashboard::data::DashboardData;
+use crate::dashboard::ui::screen::controller::ScreenController;
+use crate::dashboard::ui::screens::worker::worker_info_table::WorkerInfoTable;
+use tako::WorkerId;
+use tui::layout::{Constraint, Direction, Layout, Rect};
+
+#[derive(Default)]
+pub struct WorkerOverviewScreen {
+    /// The worker info screen shows data for this worker
+    worker_id: WorkerId,
```
Kobzol, line 18: Let's store this as `Option<WorkerId>`. With the current implementation, 0 will be used as a default, and if the screen is rendered in such state, it will draw some garbage or crash in some random place. If an `Option` is here, if the screen is drawn with the option being `None`, we can react to it explicitly (either by crashing or by switching the state to the overview screen).

crates/hyperqueue/src/dashboard/ui/screens/worker/worker_info_table.rs
```diff
@@ -19,20 +18,11 @@ struct WorkerInfoDataRow {
...
-                .any(|id| id == worker_id)
-            {
-                let info = data.query_worker_info_for(&worker_id);
-                if let Some(info) = info {
-                    let rows = create_rows(info);
-                    self.table.set_items(rows);
-                }
-            }
-        }
+    pub fn update(&mut self, data: &DashboardData, worker_id: WorkerId) {
```
Kobzol, line 21: What about passing the `WorkerConfiguration` to this function directly? Then we could avoid the `unwrap` and handle it in the worker info screen.

Kobzol (changes requested):

crates/hyperqueue/src/dashboard/ui/screens/worker/screen.rs
```diff
@@ -0,0 +1,129 @@
...
+            frame,
+            style_overview_footer(),
+        );
+
+        self.worker_info_table
+            .draw(layout.worker_info_table_chunk, frame);
+    }
+
+    fn update(&mut self, data: &DashboardData, controller: &mut ScreenController) {
+        match data
```
Kobzol, line 49: This could probably be rewritten with `if/else` :)

Kobzol (changes requested):

crates/hyperqueue/src/dashboard/ui/screens/worker/screen.rs
```diff
@@ -46,11 +50,16 @@ impl Screen for WorkerOverviewScreen {
     }
 
     fn update(&mut self, data: &DashboardData, controller: &mut ScreenController) {
-        match data
-            .query_connected_worker_ids(SystemTime::now())
-            .any(|id| id == self.worker_id)
+        match self.worker_id.is_some()
```
Kobzol, line 53: Let's `match` on `self.worker_id` to avoid the `unwrap`s.

PR #350, opened 2022-01-28
HQ job cat task status

User3574:
Allows filtering of task(s) for job cat

Closes: https://github.com/It4innovations/hyperqueue/issues/286

### DIFF
--- a/crates/hyperqueue/src/client/commands/job.rs
+++ b/crates/hyperqueue/src/client/commands/job.rs
@@ -1,17 +1,20 @@
 use crate::client::globalsettings::GlobalSettings;
 use crate::client::job::get_worker_map;
 use crate::client::output::outputs::OutputStream;
-use crate::client::status::{job_status, Status};
+use crate::client::output::resolve_task_paths;
+use crate::client::status::{get_task_status, job_status, Status};
 use crate::common::arraydef::IntArray;
 use crate::common::cli::SelectorArg;
-use crate::rpc_call;
+use crate::server::job::JobTaskInfo;
 use crate::transfer::connection::ClientConnection;
 use crate::transfer::messages::{
     CancelJobResponse, CancelRequest, FromClientMessage, JobDetailRequest, JobInfoRequest,
     Selector, ToClientMessage,
 };
-use crate::JobId;
+use crate::{rpc_call, JobTaskId};
+use crate::{JobId, Set};
 use clap::Parser;
+use std::collections::HashMap;
 
 #[derive(Parser)]
 pub struct JobListOpts {
@@ -56,6 +59,16 @@ pub struct JobCatOpts {
     /// Select task(s) outputs to view
     #[clap(long)]
     pub tasks: Option<SelectorArg>,
+
+    /// Display only task(s) with the given states.
+    /// You can use multiple states separated by a comma.
+    #[clap(
+        long,
+        multiple_occurrences(false),
+        use_delimiter(true),
+        possible_values = &["waiting", "running", "finished", "failed", "canceled"]
+    )]
+    pub task_status: Vec<Status>,
 
     /// Prepend the output of each task with a header line that identifies the task
     /// which produced that output.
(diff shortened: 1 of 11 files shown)


crates/hyperqueue/src/client/output/quiet.rs
```diff
@@ -75,8 +75,9 @@ impl Output for Quiet {
         task_selector: Option<Selector>,
         output_stream: OutputStream,
         task_header: bool,
+        task_status: Vec<Status>,
```
Kobzol, line 78: The function signature has gotten quite complex. I think that it would be nicer if the function now only received a list of tasks and the filtering by ID and status was done in `commands::job.rs`.

crates/hyperqueue/src/client/output/cli.rs
```diff
@@ -958,6 +960,10 @@ pub fn print_job_output(
 
     match task_selector {
         None | Some(Selector::All) => {
+            if !task_status.is_empty() {
```
Kobzol, line 963: Probably we can just filter the tasks at the beginning of the function once to avoid doing it three times? To avoid the "task not found" warning for tasks that were found, but also were filtered, we can build a set of filtered tasks and only print the warning if the task is not inside that set.

Kobzol (changes requested):

crates/hyperqueue/src/client/output/cli.rs
```diff
@@ -917,18 +917,21 @@ pub(crate) fn job_progress_bar(
...
 pub fn print_job_output(
-    mut job: JobDetail,
-    task_selector: Option<Selector>,
+    tasks: Vec<JobTaskInfo>,
     output_stream: OutputStream,
     task_header: bool,
+    task_paths: TaskToPathsMap,
 ) -> anyhow::Result<()> {
-    let task_to_paths = resolve_task_paths(&job);
+    if tasks.is_empty() {
```
Kobzol, line 925: Please move this to `job.rs`, this code should print the output even if the task list is empty.

crates/hyperqueue/src/client/output/mod.rs
```diff
@@ -1,5 +1,5 @@
 pub mod cli;
-mod common;
+pub(crate) mod common;
```
Kobzol, line 2: You can also just export a single function like this:
```rust
pub use common::resolve_task_paths;
```
and keep `common` private.

crates/hyperqueue/src/client/commands/job.rs
```diff
@@ -174,13 +188,59 @@
...
 
-    if let Some(job) = responses.pop().and_then(|v| v.1) {
-        return gsettings.printer().print_job_output(
-            job,
-            task_selector,
-            output_stream,
-            task_header,
-        );
+    if let Some(mut job) = responses.pop().and_then(|v| v.1) {
+        let all_ids: Vec<JobTaskId> = job.tasks.iter().map(|task| task.task_id).collect();
```
Kobzol, line 192: Please change this to `Set`, otherwise it would be O(n^2) when checking the task ID, which could be nasty if there are a lot of tasks.

crates/hyperqueue/src/client/commands/job.rs
```diff
@@ -174,13 +188,59 @@
...
+                            log::warn!("Task {task_id} not found");
+                        }
+                        task_opt
+                    })
+                    .collect();
+                selected_tasks
+                    .into_iter()
+                    .flatten()
+                    .map(|task| task.clone())
+                    .collect()
```
Kobzol, line 229:
```rust
                arr.iter()
                    .filter_map(|task_id| {
                        let task_id: JobTaskId = task_id.into();
                        let task_opt = tasks_map.get(&task_id);

                        if task_opt.is_none() && !all_ids.contains(&task_id) {
                            log::warn!("Task {task_id} not found");
                        }
                        task_opt.cloned()
                    })
                    .collect()
```

Kobzol (changes requested):

Issue #338, opened 2022-01-20
Filter tasks in `JobDetail`


Kobzol:
Currently, the `JobDetail` client message only allows returning no tasks or all tasks of a job. It would be nice to instead make it take a task selector.

PR #351, opened 2022-02-01
Filter tasks by ID and status on the server

User3574:
This PR moves filtering task logic into server.
Filtering by status is included.

Closes: https://github.com/It4innovations/hyperqueue/issues/338

### DIFF
--- a/crates/hyperqueue/src/bin/hq.rs
+++ b/crates/hyperqueue/src/bin/hq.rs
@@ -29,7 +29,7 @@ use hyperqueue::client::output::json::JsonOutput;
 use hyperqueue::client::output::outputs::{Output, Outputs};
 use hyperqueue::client::output::quiet::Quiet;
 use hyperqueue::client::status::Status;
-use hyperqueue::common::cli::SelectorArg;
+use hyperqueue::common::cli::{get_task_selector, IdSelectorArg};
 use hyperqueue::common::fsutils::absolute_path;
 use hyperqueue::common::setup::setup_logging;
 use hyperqueue::dashboard::ui_loop::start_ui_loop;
@@ -146,7 +146,7 @@ enum WorkerCommand {
 #[derive(Parser)]
 struct WorkerStopOpts {
     /// Select worker(s) to stop
-    selector_arg: SelectorArg,
+    selector_arg: IdSelectorArg,
 }
 
 #[derive(Parser)]
@@ -214,13 +214,13 @@ enum JobCommand {
 #[derive(Parser)]
 pub struct JobWaitOpts {
     /// Select job(s) to wait for
-    selector_arg: SelectorArg,
+    selector_arg: IdSelectorArg,
 }
 
 #[derive(Parser)]
 pub struct JobProgressOpts {
     /// Select job(s) to observe
-    selector_arg: SelectorArg,
+    selector_arg: IdSelectorArg,
 }
 
 #[derive(Parser)]
@@ -254,7 +254,7 @@ async fn command_job_list(gsettings: &GlobalSettings, opts: JobListOpts) -> anyh
 }
 
 async fn command_job_detail(gsettings: &GlobalSettings, opts: JobInfoOpts) -> anyhow::Result<()> {
-    if matches!(opts.selector_arg, SelectorArg::All) {
+    if matches!(opts.selector_arg, IdSelectorArg::All) {
         log::warn!(
             "Job detail doesn't support the `all` selector, did you mean to use `hq job list --all`?"
         );
@@ -267,7 +267,13 @@ async fn command_job_detail(gsettings: &GlobalSettings, opts: JobInfoOpts) -> an
 
 async fn command_job_tasks(gsettings: &GlobalSettings, opts: JobTasksOpts) -> anyhow::Result<()> {
     let mut connection = get_client_connection(gsettings.server_directory()).await?;
-    output_job_tasks(gsettings, &mut connection, opts.job_id.into()).await
+    output_job_tasks(
+        gsettings,
+        &mut connection,
+        opts.job_id.into(),
+        get_task_selector(Some(opts.task_selector)),
+    )
+    .await
 }
 
 async fn command_job_cat(gsettings: &GlobalSettings, opts: JobCatOpts) -> anyhow::Result<()> {
@@ -276,10 +282,9 @@ async fn command_job_cat(gsettings: &GlobalSettings, opts: JobCatOpts) -> anyhow
         gsettings,
         &mut connection,
         opts.job_id,
-        opts.tasks.map(|selector| selector.into()),
+        get_task_selector(Some(opts.task_selector)),
         opts.stream,
         opts.print_task_header,
-        opts.task_status,
     )
     .await
 }
--- a/crates/hyperqueue/src/client/commands/job.rs
+++ b/crates/hyperqueue/src/client/commands/job.rs
@@ -2,19 +2,17 @@ use crate::client::globalsettings::GlobalSettings;
 use crate::client::job::get_worker_map;
 use crate::client::output::outputs::OutputStream;
 use crate::client::output::resolve_task_paths;
-use crate::client::status::{get_task_status, job_status, Status};
+use crate::client::status::{job_status, Status};
 use crate::common::arraydef::IntArray;
-use crate::common::cli::SelectorArg;
-use crate::server::job::JobTaskInfo;
+use crate::common::cli::{IdSelectorArg, TaskSelectorArg};
+use crate::rpc_call;
 use crate::transfer::connection::ClientConnection;
 use crate::transfer::messages::{
-    CancelJobResponse, CancelRequest, FromClientMessage, JobDetailRequest, JobInfoRequest,
-    Selector, ToClientMessage,
+    CancelJobResponse, CancelRequest, FromClientMessage, IdSelector, JobDetailRequest,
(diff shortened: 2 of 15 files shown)


crates/hyperqueue/src/client/commands/job.rs
```diff
@@ -36,19 +34,29 @@ pub struct JobListOpts {
...
+
+    /// Display only task(s) with the given states.
+    /// You can use multiple states separated by a comma.
+    #[clap(
+        long,
+        multiple_occurrences(false),
+        use_delimiter(true),
+        possible_values = &["waiting", "running", "finished", "failed", "canceled"]
+    )]
+    pub task_status: Vec<Status>,
```
Kobzol, line 53: This help string is getting repeated a lot, it would be annoying/error-prone to change, especially the `possible_values` field. What about creating a struct that will contain both task ID and task status selector and then using it in `job tasks`, `job cat` and possibly other commands that allow filtering tasks? You can use `#[clap(flatten)] ` to "inline" the struct into its parent.

crates/hyperqueue/src/common/cli.rs
```diff
@@ -1,31 +1,56 @@
...
+    fn from(id_selector_arg: IdSelectorArg) -> Self {
+        match id_selector_arg {
+            IdSelectorArg::Id(array) => IdSelector::Specific(array),
+            IdSelectorArg::Last => IdSelector::LastN(1),
+            IdSelectorArg::All => IdSelector::All,
+        }
+    }
+}
+
+impl From<Option<IdSelectorArg>> for IdSelector {
```
Kobzol, line 34: It was quite weird for me to see you create an `Option` and then call `into` it. It is not immediately obvious what these conversions do, so I would rather create explicit functions for them, instead of implementing `From`.

crates/hyperqueue/src/server/client/mod.rs
```diff
@@ -289,11 +296,11 @@ fn compute_job_detail(
     ToClientMessage::JobDetailResponse(responses)
 }
 
-fn get_job_ids(state: &State, selector: Selector) -> Vec<JobId> {
+fn get_job_ids(state: &State, selector: IdSelector) -> Vec<JobId> {
```
Kobzol, line 299: Let's take a reference here to avoid cloning the selector.

crates/hyperqueue/src/server/client/mod.rs
```diff
@@ -306,17 +313,17 @@ async fn compose_server_stats(_state_ref: &StateRef, backend: &Backend) -> ToCli
     ToClientMessage::StatsResponse(StatsResponse { stream_stats })
 }
 
-fn compute_job_info(state_ref: &StateRef, selector: Selector) -> ToClientMessage {
+fn compute_job_info(state_ref: &StateRef, selector: IdSelector) -> ToClientMessage {
```
Kobzol, line 316: Same here.

crates/hyperqueue/src/server/client/mod.rs
```diff
@@ -328,18 +335,18 @@ fn compute_job_info(state_ref: &StateRef, selector: Selector) -> ToClientMessage
 async fn handle_job_cancel(
     state_ref: &StateRef,
     tako_ref: &Backend,
-    selector: Selector,
+    selector: IdSelector,
```
Kobzol, line 338: And here.

crates/hyperqueue/src/server/job.rs
```diff
@@ -185,16 +187,65 @@ impl Job {
...
 
-    pub fn make_job_detail(&self, include_tasks: bool) -> JobDetail {
-        let tasks = if include_tasks {
-            self.tasks.values().cloned().collect()
-        } else {
-            Vec::new()
+    pub fn make_job_detail(
+        &self,
+        task_id_selector: IdSelector,
+        task_status_selector: StatusSelector,
```
Kobzol, line 193: Let's also avoid cloning the selectors here, this function can be called a lot of times and the selector can contain a vector.

crates/hyperqueue/src/server/job.rs
```diff
@@ -185,16 +187,65 @@ impl Job {
...
+    ) -> JobDetail {
+        let all_task_ids: Set<JobTaskId> = self.tasks.iter().map(|task| task.1.task_id).collect();
+        let mut task_ids_not_found: Vec<JobTaskId> = vec![];
+
+        // Select tasks by status
+        let mut tasks = match task_status_selector {
```
Kobzol, line 199: To avoid going through the tasks multiple times, I would suggest to just iterate through them once and then check if they are valid (by status and ID). If yes, add them to a result vec (or add their ID to the not found vec).
That way we will also avoid needless clones if e.g. something passes through status but then is filtered by ID.

crates/hyperqueue/src/transfer/messages.rs
```diff
@@ -92,12 +92,19 @@ pub struct SubmitRequest {
...
 #[derive(Serialize, Deserialize, Debug, Clone)]
-pub enum Selector {
+pub enum IdSelector {
     All,
     LastN(u32),
     Specific(IntArray),
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub enum StatusSelector {
```
Kobzol, line 102: Now that I look at it, `None` is probably not a good choice here. It would allow you to send e.g. `IdSelector::All` and `StatusSelector::None`, which doesn't really make much sense. We could optimize this situation in `make_job_detail`, but it shouldn't be possible to ask for this in the first place. And it wouldn't be clear if the client wants to warn about missing IDs in such case or not.

Let's introduce something like `TaskSelector`, which will contain both `IdSelector` and `StatusSelector` fields. Then remove the `None` variant from `StatusSelector` and add `Option<TaskSelector>` to `JobDetailRequest`. Not requiring tasks would then be modeled as passing `None` to it.

Kobzol (changes requested):

crates/hyperqueue/src/bin/hq.rs
```diff
@@ -271,10 +277,9 @@ async fn command_job_cat(gsettings: &GlobalSettings, opts: JobCatOpts) -> anyhow
         gsettings,
         &mut connection,
         opts.job_id,
-        opts.tasks.map(|selector| selector.into()),
+        get_task_selector(Option::from(opts.task_selector)),
```
Kobzol, line 280: `Option::from(...)` => `Some(...)` (happens on multiple places).

crates/hyperqueue/src/client/commands/job.rs
```diff
@@ -36,39 +34,33 @@ pub struct JobListOpts {
...
-    pub selector_arg: SelectorArg,
+    pub selector_arg: IdSelectorArg,
 }
 
 #[derive(Parser)]
 pub struct JobTasksOpts {
     /// Job ID
     pub job_id: u32,
+
+    /// Select specific task(s) by ids and state
```
Kobzol, line 45: Let's remove the documentation comments on `flatten`ed fields. The reader might think that they are shown in help, but they are not. The same in `JobCatOpts`.

crates/hyperqueue/src/client/commands/job.rs
```diff
@@ -129,11 +121,14 @@ pub async fn output_job_list(
...
-    selector: Selector,
+    selector: IdSelector,
 ) -> anyhow::Result<()> {
     let message = FromClientMessage::JobDetail(JobDetailRequest {
-        selector,
-        include_tasks: true,
+        job_id_selector: selector,
+        task_selector: Some(TaskSelector {
+            id_selector: IdSelector::All,
+            status_selector: StatusSelector::All,
```
Kobzol, line 130: The CLI printer currently only cares about failed jobs, but in JSON all tasks are printed. Maybe later we could "optimize" this by only downloading failed tasks if the CLI printer is used. But don't do that in this PR, it's not a problem currently.

crates/hyperqueue/src/common/cli.rs
```diff
@@ -1,31 +1,75 @@
...
+            IdSelectorArg::Id(array) => IdSelector::Specific(array),
+            IdSelectorArg::Last => IdSelector::LastN(1),
+            IdSelectorArg::All => IdSelector::All,
         }
     }
 }
+
+#[derive(Parser)]
+pub struct TaskSelectorArg {
+    /// Select task(s) outputs to view
```
Kobzol, line 37: This help now has to be general, e.g. in `hq job tasks` it should not mention outputs. 

Let's use something more general, like:
`Filter task(s) by ID.`
and
`Filter task(s) by status. You can use ...`

crates/hyperqueue/src/common/cli.rs
```diff
@@ -1,31 +1,75 @@
...
+        long,
+        multiple_occurrences(false),
+        use_delimiter(true),
+        possible_values = &["waiting", "running", "finished", "failed", "canceled"]
+    )]
+    pub task_status: Vec<Status>,
+}
+
+pub fn get_task_selector(opt_task_selector_arg: Option<TaskSelectorArg>) -> Option<TaskSelector> {
+    match opt_task_selector_arg {
```
Kobzol, line 53:
```rust
opt_task_selector_arg.map(|arg| TaskSelector {
        id_selector: get_id_selector(arg.tasks),
        status_selector: get_status_selector(arg.task_status),
    })
```

Kobzol (changes requested):

Kobzol:
@spirali Please take a look.

Kobzol (changes requested):

spirali (approved):

PR #352, opened 2022-02-01
Pytasks

spirali:


crates/pyhq/python/hyperqueue/client.py
```diff
@@ -1,14 +1,15 @@
...
 from typing import Optional
 
-from .ffi.ffi import connect_to_server, submit_job
+from .ffi.connection import Connection
 from .job import Job, JobId
 
 
 class Client:
-    def __init__(self, path: Optional[Path] = None):
+    def __init__(self, path: Optional[Path] = None, python_bin="python3"):
```
Kobzol, line 9: It will be useful to pass virtualenv here. Is it enough to pass e.g. `venv/bin/python`? Or does the environment need to be sourced?

crates/pyhq/python/hyperqueue/job.py
```diff
@@ -1,28 +1,54 @@
...
+            stdin=stdin,
+        )
         self.tasks.append(task)
         return task
 
-    def build(self) -> List[TaskConfig]:
-        return build_tasks(self.tasks)
+    def function(
+        self,
+        fn,
```
Kobzol, line 41:
```suggestion
        fn,
        *
```
?

crates/pyhq/python/hyperqueue/task/function.py
```diff
@@ -0,0 +1,47 @@
+from typing import Callable, Dict
+
+import cloudpickle
+
+from crates.pyhq.python.hyperqueue.ffi.protocol import TaskDescription
```
Kobzol, line 5: These imports should be relative.

Kobzol (approved):

crates/pyhq/python/hyperqueue/client.py
spirali: Good idea to pass venv. Just python is probably not enough. I will investigate it tomorrow

crates/pyhq/python/hyperqueue/task/function.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def _build(self, client, id_map: Dict[Task, TaskId]) -> TaskDescription:
+        depends_on = [id_map[dependency] for dependency in self.dependencies]
+        serialized_fn = serialize(self.fn)
+        return TaskDescription(
+            id=id_map[self],
+            args=[
+                client.python_bin,
+                "-c",
+                "import sys;import cloudpickle;import pickle;"
```
Kobzol, line 43: You can just use a multiline string :)

PR #357, opened 2022-02-03
Event parsing infrastructure for postprocessing analysis

fredykoci:


vizualization/hq_event_wrapper.py
```diff
@@ -0,0 +1,60 @@
+import subprocess as sp
+import os
+import json
+from events_type import *
```
Kobzol, line 4: Let's define `dataclasses` for each concrete event type, so that it is easier to work with, and then parse the individual events into the dataclasses. Like this:
```python
@dataclass
class Event:
    id: int
    time: datetime.datetime

@dataclass
class WorkerConnected(Event):
   worker: int

...
```

vizualization/hq_event_wrapper.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+class Events(list):
```
Kobzol, line 7: Please don't inherit from `list`.

vizualization/hq_event_wrapper.py
```diff
@@ -0,0 +1,60 @@
...
+            event_type = [event_type]
+        return [i for i in self if i['type'] in event_type]
+
+    def get_by_worker_id(self, id) -> list:
+        return [i for i in self.get_by_type(WorkerEvents) if i['id'] == id]
+
+
+class HQEventWrapper:
+
+    def __init__(self, hq_event_bin_path: str, cwd: str = "", hq_bin_path: str = ""):
```
Kobzol, line 22: Let's define a function that will receive a path to a JSON file and return the parsed events. For now we can just generate the JSON from CLI.

Kobzol (changes requested):

Kobzol:
Superseded by https://github.com/It4innovations/hyperqueue/pull/377.

Issue #317, opened 2022-01-15
Improve output of `hq submit --wait`


Kobzol:
When you submit a job and directly wait for its output and some tasks fail, only the number of failed tasks will be displayed. It would be nice to either:
- Display the errors of the failed tasks
- Show a hint to the user how they can display the errors (e.g. suggest `$ hq info job <job-id>`)

Related issue: https://github.com/It4innovations/hyperqueue/issues/311

PR #358, opened 2022-02-04
Show error msgs while submitting with wait

User3574:
Solves: Issue #317


crates/hyperqueue/src/client/commands/wait.rs
```diff
@@ -28,14 +29,32 @@ pub async fn wait_for_jobs(
...
+            selector: selector.clone(),
         }),
         ToClientMessage::WaitForJobsResponse(r) => r
     )
     .await?;
 
+    let detail = rpc_call!(
+        connection,
+        FromClientMessage::JobDetail(JobDetailRequest {
+            job_id_selector: selector.clone(),
```
Kobzol, line 41:
```suggestion
            job_id_selector: selector,
```
It doesn't matter here, but if it's the last usage within the function, you don't need to clone it and you can just move it.

crates/hyperqueue/src/client/commands/wait.rs
```diff
@@ -28,14 +29,32 @@ pub async fn wait_for_jobs(
...
+            task_selector: Some(TaskSelector {
+                id_selector: TaskIdSelector::All,
+                status_selector: TaskStatusSelector::Specific(vec![Status::Failed]),
+            })
+        }),
+        ToClientMessage::JobDetailResponse(r) => r
+    )
+    .await?;
```
Kobzol, line 49: Let's call this only if we actually see any `failed` tasks in `response`.

Kobzol (changes requested):

crates/hyperqueue/src/client/output/cli.rs
```diff
@@ -199,6 +199,34 @@ impl CliOutput {
             }
         }
     }
+
+    fn print_failed_tasks(&self, detail: &JobDetail, worker_map: &WorkerMap) {
```
Kobzol, line 203: This is basically the same thing as `print_task_summary`. These functions should do the same thing - display a summary of failed tasks. Could you try to unify them please?

Kobzol (changes requested):

Kobzol:
It looks fine, but since the wait messages (like X tasks have failed and wait finished in XXX) are logged to stderr, they are not synchronized with stdout, so the table with failed tasks can be displayed sooner than the log, which looks weird. Let's print the wait messages to stdout directly instead of logging them. But in that case they can only be logged when the output is CLI, so please move them to the CLI printer.

Issue #322, opened 2022-01-17
Implement `hq task list`


Kobzol:
The command should show all tasks of selected job(s):
```bash
$ hq task list <job-selector>
```

PR #359, opened 2022-02-09
HQ task list

User3574:
Displays task(s) of given job(s)
Closes: #322

### DIFF
--- a/crates/hyperqueue/src/bin/hq.rs
+++ b/crates/hyperqueue/src/bin/hq.rs
@@ -29,7 +29,8 @@ use hyperqueue::client::output::json::JsonOutput;
 use hyperqueue::client::output::outputs::{Output, Outputs};
 use hyperqueue::client::output::quiet::Quiet;
 use hyperqueue::client::status::Status;
-use hyperqueue::common::cli::{get_task_selector, IdSelectorArg};
+use hyperqueue::common::arraydef::IntArray;
+use hyperqueue::common::cli::{get_task_selector, IdSelectorArg, TaskSelectorArg};
 use hyperqueue::common::fsutils::absolute_path;
 use hyperqueue::common::setup::setup_logging;
 use hyperqueue::dashboard::ui_loop::start_ui_loop;
@@ -100,6 +101,8 @@ enum SubCommand {
     Server(ServerOpts),
     /// Commands for controlling HyperQueue jobs
     Job(JobOpts),
+    /// Commands for displaying task(s)
+    Task(TaskOpts),
     /// Submit a job to HyperQueue
     Submit(JobSubmitOpts),
     /// Commands for controlling HyperQueue workers
@@ -111,7 +114,7 @@ enum SubCommand {
     AutoAlloc(AutoAllocOpts),
     /// Event log management
     EventLog(EventLogOpts),
-    ///Commands for the dashboard
+    /// Commands for the dashboard
     Dashboard(DashboardOpts),
     /// Generate shell completion script
     GenerateCompletion(GenerateCompletionOpts),
@@ -223,6 +226,31 @@ pub struct JobProgressOpts {
     selector_arg: IdSelectorArg,
 }
 
+// Task CLI options
+
+#[derive(Parser)]
+pub struct TaskOpts {
+    #[clap(subcommand)]
+    subcmd: TaskCommand,
+}
+
+#[derive(Parser)]
+enum TaskCommand {
+    /// Displays task(s) associated with selected job(s)
+    List(TaskListOpts),
+}
+
+#[derive(Parser)]
+struct TaskListOpts {
+    /// Job id(s) selector
+    pub job_ids: IdSelectorArg,
+
+    /// Filter task(s) by status.
+    /// You can use multiple states separated by a comma.
+    #[clap(long, multiple_occurrences(false), use_delimiter(true), arg_enum)]
+    pub task_status: Vec<Status>,
+}
+
 #[derive(Parser)]
 struct GenerateCompletionOpts {
     /// Shell flavour for which the completion script should be generated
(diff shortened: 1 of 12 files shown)


crates/hyperqueue/src/bin/hq.rs
```diff
@@ -111,10 +111,37 @@ enum SubCommand {
...
+
+#[derive(Parser)]
+pub struct TaskOpts {
+    #[clap(subcommand)]
+    subcmd: TaskCommand,
+}
+
+#[derive(Parser)]
+enum TaskCommand {
+    /// Displays all tasks associated with selected job
```
Kobzol, line 132: `Display tasks associated with selected job(s)`

It does not have to be all tasks, since they can be filtered :)

crates/hyperqueue/src/bin/hq.rs
```diff
@@ -526,6 +567,9 @@ async fn main() -> hyperqueue::Result<()> {
         SubCommand::AutoAlloc(opts) => command_autoalloc(&gsettings, opts).await,
         SubCommand::EventLog(opts) => command_event_log(opts),
         SubCommand::GenerateCompletion(opts) => generate_completion(opts),
+        SubCommand::Task(TaskOpts {
```
Kobzol, line 570: Let's move this branch (and also the Task enum variant in the SubCommand enum) closer to the job branches/variants.

crates/hyperqueue/src/client/commands/job.rs
```diff
@@ -161,6 +161,37 @@ pub async fn output_job_tasks(
...
+    let message = FromClientMessage::JobDetail(JobDetailRequest {
+        job_id_selector,
+        task_selector,
+    });
+    let responses =
+        rpc_call!(connection, message, ToClientMessage::JobDetailResponse(r) => r).await?;
+
+    let jobs = responses
+        .into_iter()
+        .filter(|(job_id, opt_job)| match opt_job {
```
Kobzol, line 179: You can use `.filter_map` to handle filter and map in one go. You can then also get rid of the `unwrap`.

crates/hyperqueue/src/client/output/outputs.rs
```diff
@@ -44,6 +45,7 @@ pub trait Output {
     fn print_job_list(&self, jobs: Vec<JobInfo>, total_jobs: usize);
     fn print_job_detail(&self, job: JobDetail, worker_map: WorkerMap);
     fn print_job_tasks(&self, job: JobDetail, worker_map: WorkerMap);
+    fn print_jobs_tasks(&self, jobs: Vec<(JobId, JobDetail)>, worker_map: WorkerMap);
```
Kobzol, line 48: I would maybe call it just `print_tasks`.

crates/hyperqueue/src/client/output/json.rs
```diff
@@ -152,6 +152,8 @@ impl Output for JsonOutput {
         self.print(format_tasks(job.tasks, map));
     }
 
+    fn print_jobs_tasks(&self, _jobs: Vec<(JobId, JobDetail)>, _worker_map: WorkerMap) {}
```
Kobzol, line 155: I think that we could easily implement this. Let's create a dictionary keyed by job ID and use `format_tasks` for the values.

crates/hyperqueue/src/client/output/cli.rs
```diff
@@ -546,6 +546,75 @@ impl Output for CliOutput {
         self.print_horizontal_table(rows, header);
     }
 
+    fn print_jobs_tasks(&self, mut jobs: Vec<(JobId, JobDetail)>, worker_map: WorkerMap) {
```
Kobzol, line 549: This function is basically the same as `print_job_tasks`. Let's unify these two functions and only print the job ID column if there is more than one job.

Kobzol (changes requested):

crates/hyperqueue/src/bin/hq.rs
```diff
@@ -111,10 +112,37 @@ enum SubCommand {
...
+
+#[derive(Parser)]
+pub struct TaskOpts {
+    #[clap(subcommand)]
+    subcmd: TaskCommand,
+}
+
+#[derive(Parser)]
+enum TaskCommand {
+    /// Displays filtered task(s) associated with selected job(s)
```
Kobzol:
```suggestion
    /// Displays task(s) associated with selected job(s)
```

crates/hyperqueue/src/client/output/cli.rs
```diff
@@ -487,62 +487,84 @@ impl Output for CliOutput {
...
+            );
+        }
+
+        let mut header = match jobs_len {
+            1 => vec!["JobID".cell().bold(true)],
+            _ => vec![],
+        };
+
+        header.append(&mut vec![
+            "TaskID".cell().bold(true),
```
Kobzol, line 560:
```suggestion
            "Task ID".cell().bold(true),
```

crates/hyperqueue/src/client/output/cli.rs
```diff
@@ -487,62 +487,84 @@ impl Output for CliOutput {
...
+                        ]);
+
+                        job_rows
+                    })
+                    .collect(),
+            );
+        }
+
+        let mut header = match jobs_len {
+            1 => vec!["JobID".cell().bold(true)],
```
Kobzol, line 555:
```suggestion
            1 => vec!["Job ID".cell().bold(true)],
```

Kobzol (changes requested):

tests/test_task.py
```diff
@@ -0,0 +1,50 @@
+from .conftest import HqEnv
+from .utils import wait_for_job_state
+
+
+def test_task_list_single(hq_env: HqEnv):
+    hq_env.start_server()
+    hq_env.start_worker()
+
+    hq_env.command(
+        ["submit", "--array=0-3", "--", "bash", "-c", "echo $HQ_JOB_ID-$HQ_TASK_ID"]
```
Kobzol, line 10: The output is not important here. So let's use something simpler, like `hostname`, to not confuse the reader who might think that this complex output is important for the test :)

Kobzol (changes requested):

PR #360, opened 2022-02-09
Fix missing placeholder validation

Kobzol:
Do not warn that stdout/stderr of an array job misses the TASK_ID placeholder if it is contained in CWD and the path contains the CWD placeholder.


spirali (approved):
Update changelog please

PR #374, opened 2022-02-18
FIX: Set cancel state for tasks which dependency has failed

spirali:


crates/hyperqueue/src/server/state.rs
```diff
@@ -164,6 +167,15 @@ impl State {
     ) {
         log::debug!("Task id={} failed: {:?}", msg.id, msg.info);
 
+        for task_id in msg.cancelled_tasks {
+            log::debug!(
+                "Task id={} canceled because of task dependency fails",
+                task_id
+            );
+            let job = self.get_job_mut_by_tako_task_id(msg.id).unwrap();
```
Kobzol, line 175: The job can be looked-up only once (and merged with the lookup on line `179`).

tests/pyapi/test_dependencies.py
```diff
@@ -15,3 +15,24 @@ def test_single_dep(hq_env: HqEnv):
     job_id = client.submit(job)
 
     wait_for_job_state(hq_env, job_id, "FINISHED")
+
+
+def test_dep_failed(hq_env: HqEnv):
+    """
```
Kobzol, line 21: Text comment is wrong.

Kobzol (changes requested):

Kobzol (approved):

Issue #375, opened 2022-02-18
Shown makespan of canceled tasks grows in time.


spirali:
``hq job tasks <X>`` shows invalid makespan that grows in time
![image](https://user-images.githubusercontent.com/263349/154703906-04a58b3b-4a9c-4d3a-9cfd-d7d69a4f7e0e.png)

PR #376, opened 2022-02-18
FIX: tasks makespan time

User3574:
Closes: #375

### DIFF
--- a/crates/hyperqueue/src/client/output/cli.rs
+++ b/crates/hyperqueue/src/client/output/cli.rs
@@ -1066,7 +1066,8 @@ fn get_task_time(state: &JobTaskState) -> (Option<DateTime<Utc>>, Option<DateTim
     match state {
         JobTaskState::Canceled {
             started_data: Some(started_data),
-        } => (Some(started_data.start_date), None),
+            cancelled_date,
+        } => (Some(started_data.start_date), Some(*cancelled_date)),
         JobTaskState::Running { started_data, .. } => (Some(started_data.start_date), None),
         JobTaskState::Finished {
             started_data,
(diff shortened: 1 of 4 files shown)


crates/hyperqueue/src/client/output/cli.rs
```diff
@@ -1078,7 +1079,7 @@ fn get_task_time(state: &JobTaskState) -> (Option<DateTime<Utc>>, Option<DateTim
             end_date,
             ..
         } => (Some(started_data.start_date), Some(*end_date)),
-        JobTaskState::Canceled { started_data: None } | JobTaskState::Waiting => (None, None),
+        _ => (None, None),
```
Kobzol, line 1082: Let's remove the `_` case, so that if new states are added the compiler will tell us to update this function.

crates/hyperqueue/src/server/job.rs
```diff
@@ -357,11 +358,15 @@ impl Job {
...
                     started_data: Some(started_data.clone()),
+                    cancelled_date: Some(now),
                 };
                 self.counters.n_running_tasks -= 1;
             }
             JobTaskState::Waiting => {
-                *state = JobTaskState::Canceled { started_data: None };
+                *state = JobTaskState::Canceled {
+                    started_data: None,
+                    cancelled_date: None,
```
Kobzol, line 368: We always know the cancel date, so let's set it here and remove `Option` from `cancelled_date`.

Kobzol (changes requested):

tests/test_job.py
```diff
@@ -705,6 +705,20 @@ def test_job_tasks_table(hq_env: HqEnv):
...
 
 
+def test_job_tasks_makespan(hq_env: HqEnv):
+    hq_env.start_server()
+    hq_env.start_worker()
+
+    hq_env.command(["submit", "sleep", "10"])
+    hq_env.command(["job", "cancel", "1"])
+
+    table_1 = hq_env.command(["job", "tasks", "1"])
```
Kobzol, line 715: 1) Let's wait for the job to be cancelled before showing the task table.
2) Check that the makespan is not empty.

Kobzol (changes requested):

Issue #174, opened 2021-10-22
Display warning if autoalloc options seem suspicious.


Kobzol:
- Queue does not exist.
- Walltime is larger than the walltime limit of the given queue.
- Backlog size is larger than the max. queue limit of the given queue.
- Workers per allocation is larger than the max. number of workers of the given queue.

For PBS, you can find this with `qstat -Q <queue> -f -F json`. For Slurm, you can find it with `sinfo -O <...> -p <queue>`

PR #381, opened 2022-02-25
WIP: hq alloc checker

User3574:
Closes: #174


crates/hyperqueue/src/server/client/autoalloc.rs
```diff
@@ -238,6 +242,206 @@ fn create_queue(
...
 }
 
+struct QueueLimit {
+    walltime: Option<u32>,
+    backlog: Option<u32>,
+    workers: Option<u32>,
+}
+
+// Returns value of given arg, in formats: keyVALUE, key=VALUE, [key, VALUE]
+fn get_arg_by_key(key: String, args: &[String]) -> Option<String> {
```
Kobzol, line 252: It would be nice to write some Rust unit tests that check how this works for `-qqexp`, `-q qexp`, `-q=qexp` etc.

crates/hyperqueue/src/server/client/autoalloc.rs
```diff
@@ -238,6 +242,206 @@ fn create_queue(
...
+            .arg("-Q")
+            .arg(&queue_type)
+            .arg("-f")
+            .arg("-F")
+            .arg("json")
+            .output()?;
+
+        let output = String::from_utf8_lossy(&queue_info.stdout).into_owned();
+        let json: serde_json::Value =
+            serde_json::from_str(&*output).expect("JSON was not well-formatted");
```
Kobzol, line 298: Use `context` and `?` instead, this shouldn't crash the program.

crates/hyperqueue/src/server/client/autoalloc.rs
```diff
@@ -238,6 +242,206 @@ fn create_queue(
...
+        let queue_type = q_type.replace("-q", "");
+        let queue_info = Command::new("qstat")
+            .arg("-Q")
+            .arg(&queue_type)
+            .arg("-f")
+            .arg("-F")
+            .arg("json")
+            .output()?;
+
+        let output = String::from_utf8_lossy(&queue_info.stdout).into_owned();
```
Kobzol, line 296: `into_owned` is not needed, below `from_str(&output)` is enough.

crates/hyperqueue/src/server/client/autoalloc.rs
```diff
@@ -238,6 +242,206 @@ fn create_queue(
...
+        let json: serde_json::Value =
+            serde_json::from_str(&*output).expect("JSON was not well-formatted");
+        let queue = json
+            .get("Queue")
+            .and_then(|x| x.get(queue_type))
+            .context("can't find queue info")?;
+
+        let (walltime, backlog, workers) = match q_type.as_str() {
```
Kobzol, line 304: We cannot depend on the specific queue names. Instead just try to find `resources_available/nodes` first and if it's missing, try `resources_max/nodect`.

crates/hyperqueue/src/server/client/autoalloc.rs
```diff
@@ -238,6 +242,206 @@ fn create_queue(
...
+            ),
+            "qdgx" => (
+                "resources_max/walltime",
+                "max_queued",
+                "resources_available/nodes",
+            ),
+            _ => return Ok(None),
+        };
+
+        let (walltime, backlog, workers) = (
```
Kobzol, line 353: This is needlessly complicated, just create the variables one by one.

crates/hyperqueue/src/server/client/autoalloc.rs
```diff
@@ -238,6 +242,206 @@ fn create_queue(
...
+                }
+                Err(err) => {
+                    log::warn!("Can't find key {walltime}: {err}");
+                    None
+                }
+            },
+            match get_value_by_key_json(queue, String::from(backlog)) {
+                Ok(value) => Some(
+                    value
+                        .to_string()
```
Kobzol, line 368: Why is this `is_numeric` filter here? What data is returned by PBS here?

Kobzol (changes requested):
The check has to be performed on the server, not on the client, because the client might not even have PBS available (or it might be a different PBS than on the server). Therefore we also need to return the error in a different way rather than logging it.

Currently, returning errors is quite basic in the client/server RPC, it would be really nice to use e.g. `tarpc` or some better RPC approach to solve this first.

crates/hyperqueue/src/server/client/autoalloc.rs
User3574: Backlog has specific format in our qstat: "max_queued":"[u:PBS_GENERIC=5]"

crates/hyperqueue/src/server/client/autoalloc.rs
```diff
@@ -271,3 +464,45 @@ async fn try_submit_allocation(
 
     Ok(())
 }
+
+#[cfg(test)]
+mod tests {
+    use crate::server::client::autoalloc::get_arg_by_key;
```
Kobzol, line 470: Could you please add a complete parsing example, i.e. copy the output from PBS and put it into a test. This would also helps others to understand how does the PBS output look like, since it cannot be really seen from the code now.

Kobzol (changes requested):

PR #386, opened 2022-03-05
Dashboard: Changed TaskTimeline to JobTimeline

vyomkeshj:
The goal of this PR is to make available the data required for JobScreen.


crates/hyperqueue/src/server/state.rs
```diff
@@ -191,6 +203,7 @@ impl State {
 
     pub fn process_task_update(&mut self, msg: TaskUpdate, backend: &Backend) {
         log::debug!("Task id={} updated {:?}", msg.id, msg.state);
+        let (mut job_id, mut is_job_terminated): (JobId, bool) = (JobId::default(), false);
```
Kobzol, line 206: This should be `Option<JobId>`.

crates/hyperqueue/src/dashboard/ui/widgets/chart.rs
```diff
@@ -0,0 +1,217 @@
...
+    /// Clears the chart's plotters and data.
+    pub fn clear_chart(&mut self) {
+        self.chart_plotters.clear();
+        self.datasets.clear();
+    }
+
+    /// Changes the x-range of the chart to the new `end_time`
+    /// sets `self::is_live` to false.
+    pub fn set_end_time(&mut self, end_time: SystemTime) {
+        if self.is_live {
```
Kobzol, line 72: This condition doesn't have to be here.

crates/hyperqueue/src/dashboard/ui/widgets/chart.rs
```diff
@@ -0,0 +1,217 @@
...
+        let data_points: Vec<(_, _)> = times
+            .iter()
+            .flat_map(|time| {
+                self.chart_plotters
+                    .iter()
+                    .flat_map(|fetcher| fetcher.fetch_data_points(data, *time))
+            })
+            .collect();
+
+        data_points.iter().for_each(|(key, data)| {
```
Kobzol, line 115: Just use a for cycle if you converge to `for_each`.

crates/hyperqueue/src/dashboard/ui/widgets/chart.rs
```diff
@@ -0,0 +1,217 @@
...
+/// and defines how data will be fetched for each chart with `fetch_data_points`
+/// Multiple `ChartPlotters` can be used to add different charts to `DashboardChart`
+/// with each of the charts fetching their own data.
+pub trait ChartPlotter {
+    /// Declares the labels of the charts that will be plotted by this plotter and their style.
+    fn get_charts(&self) -> Map<String, PlotStyle>;
+
+    /// Fetches one set of points for each of the keys returned by `get_charts`
+    /// at given `time`
+    fn fetch_data_points(&self, data: &DashboardData, time: SystemTime) -> Map<String, (f64, f64)>;
```
Kobzol, line 49: It seems to me that a more efficient API would be to get a range of timepoints for a given key rather than get all keys for a timepoint. But until it becomes a bottleneck we don't need to change it.

crates/hyperqueue/src/dashboard/ui/screens/job/job_info_display.rs
```diff
@@ -0,0 +1,87 @@
...
+
+#[derive(Default)]
+pub struct JobInfoTable {
+    table: StatefulTable<JobInfoDataRow>,
+}
+
+#[derive(Default, Debug)]
+struct JobInfoDataRow {
+    pub label: &'static str,
+    pub data: String,
```
Kobzol, line 18: Since these rows get created frequently, I would suggest using `Cow<str>` instead of `String` for columns that may contain dynamic data.

crates/hyperqueue/src/dashboard/ui/screens/job/job_tasks_chart.rs
```diff
@@ -0,0 +1,155 @@
...
+}
+
+impl JobTaskChart {
+    /// Sets a new job id for the chart to plot data for.
+    pub fn set_job_id(&mut self, job_id: JobId) {
+        match self.job_id {
+            Some(chart_job_id) if chart_job_id == job_id => (),
+            _ => {
+                self.job_id = Some(job_id);
+                self.chart.clear_chart();
```
Kobzol, line 31: Can we just replace the whole chart? Unless it's really needed, it's nice to avoid such heavily mutable APIs (clear + fill).

crates/hyperqueue/src/dashboard/ui/screens/job/job_tasks_chart.rs
```diff
@@ -0,0 +1,155 @@
...
+    }
+}
+
+fn get_task_counts_for_job(
+    job_id: JobId,
+    data: &DashboardData,
+    at_time: SystemTime,
+) -> (usize, usize, usize) {
+    let num_running_tasks =
+        fetch_job_task_count_with_status(job_id, data, at_time, DashboardTaskState::Running);
```
Kobzol, line 130: This seems quite ineffective, since the query is invoked three times. Let's just invoke it once, iterate the items and split them by state.

Kobzol (changes requested):

crates/hyperqueue/src/dashboard/ui/widgets/chart.rs
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    /// The x-range of the chart is from (self::end_time - view_size) -> self::end_time
+    pub fn set_view_size(&mut self, view_size: Duration) {
+        self.view_size = view_size;
+    }
+
+    pub fn set_live_status(&mut self, live: bool) {
+        self.is_live = live;
+    }
```
Kobzol, line 77: These functions don't seem to be used anywhere.

Kobzol (approved):

PR #387, opened 2022-03-08
Rewrite autoallocator to avoid periodic polling

Kobzol:


crates/hyperqueue/src/server/autoalloc/process.rs
```diff
@@ -1,255 +1,646 @@
...
+            }
+            AutoAllocMessage::GetAllocations(queue_id, response) => {
+                let result = match autoalloc.get_queue(queue_id) {
+                    Some(queue) => Ok(queue.all_allocations().cloned().collect()),
+                    None => Err(anyhow::anyhow!("Queue {queue_id} not found")),
+                };
+                response.respond(result);
+                None
+            }
+            AutoAllocMessage::Stop(response) => {
```
spirali, line 124: Do we need explicit stop? Maybe just throw away the sending part?

crates/hyperqueue/src/server/autoalloc/process.rs
```diff
@@ -1,255 +1,646 @@
...
+) -> anyhow::Result<()> {
+    let remove_alloc_fut = {
+        let queue_state = autoalloc.get_queue_mut(id);
+
+        let fut = match queue_state {
+            Some(state) => {
+                let has_running_allocations =
+                    state.all_allocations().any(|alloc| alloc.is_running());
+                if has_running_allocations && !force {
+                    return Err(anyhow::anyhow!(
```
spirali, line 275: ``anyhow::bail!``

crates/hyperqueue/src/server/autoalloc/process.rs
```diff
@@ -1,255 +1,646 @@
...
+};
+
+pub async fn autoalloc_process(
+    state_ref: StateRef,
+    mut autoalloc: AutoAllocState,
+    mut receiver: RpcReceiver<AutoAllocMessage>,
+) {
+    let max_wait_for_message = Duration::from_secs(60 * 60);
+    loop {
+        match tokio::time::timeout(max_wait_for_message, receiver.recv()).await {
```
spirali, line 39: When there will be a steady stream of message then refresh is not called and old directory will not be deleted (remove_inactive_directories)?

crates/hyperqueue/src/worker/start.rs
```diff
@@ -504,8 +504,13 @@ fn try_get_pbs_info() -> anyhow::Result<ManagerInfo> {
     let manager_job_id =
         std::env::var("PBS_JOBID").expect("PBS_JOBID not found in environment variables");
 
-    let time_limit =
-        pbs::get_remaining_timelimit(&manager_job_id).expect("Could not get PBS timelimit");
+    let time_limit = match pbs::get_remaining_timelimit(&manager_job_id) {
+        Ok(time_limit) => Some(time_limit),
+        Err(error) => {
```
spirali, line 509: ``or_else``

spirali (approved):

crates/hyperqueue/src/server/autoalloc/process.rs
Kobzol: Good point.

crates/hyperqueue/src/server/autoalloc/process.rs
Kobzol: Refresh is called after each received message. I made this more explicit.

crates/hyperqueue/src/worker/start.rs
Kobzol: We need to convert `Result` to `Option`, so it's not that easy here. I would like to use `inspect`, but it's unstable ATM.

spirali: Ah, sorry I missed this.

PR #389, opened 2022-03-23
PyApi

spirali:


crates/pyhq/python/hyperqueue/task/function.py
```diff
@@ -13,19 +13,32 @@ def purge_cache():
...
+
+    try:
+        fn, a, kw = pickle.loads(sys.stdin.buffer.read())
+        fn(*a, **(kw if kw is not None else {}))
+    except Exception:
+        import os
+        import traceback
+
+        t = traceback.format_exc()
+        with open(os.environ["HQ_ERROR_FILENAME"], "w") as f:
```
Kobzol, line 34: If `task_dir` is false, then this environment variable won't exist?

crates/pyhq/python/hyperqueue/task/task.py
```diff
@@ -1,12 +1,21 @@
...
+from hyperqueue.ffi.protocol import ResourceRequest
 
 TaskId = int
 
 
 class Task:
-    def __init__(self, dependencies: Sequence["Task"] = ()):
+    def __init__(
+        self,
+        task_id: int,
```
Kobzol, line 11: `int` -> `TaskId`

crates/pyhq/python/hyperqueue/task/task.py
```diff
@@ -1,12 +1,21 @@
-from typing import Dict, Sequence
+from typing import Optional, Sequence
+
+from hyperqueue.ffi.protocol import ResourceRequest
```
Kobzol, line 3: This import should be relative.

tests/pyapi/test_dependencies.py
```diff
@@ -25,8 +25,8 @@ def test_dep_failed(hq_env: HqEnv):
     (job, client) = prepare_job_client(hq_env, with_worker=True)
 
     t1 = job.program(args=bash("exit 1"))
-    t2 = job.program(args=bash("echo 'hello' > foo1.txt"), depends=[t1])
-    job.program(args=bash("echo 'hello' > foo2.txt"), depends=[t2])
+    t2 = job.program(args=bash("echo 'hello' > foo1.txt"), dependencies=[t1])
+    job.program(args=bash("echo 'hello' > foo2.txt"), dependencies=[t2])
```
Kobzol, line 29: Later I would vote to shorten `dependencies` to `deps`, as it will probably be used quite a lot.

tests/pyapi/test_function.py
```diff
@@ -41,3 +43,32 @@ def body():
...
+
+
+def test_function_resources(hq_env: HqEnv):
+    (job, client) = prepare_job_client(hq_env)
+
+    job.function(lambda: 1, resources=ResourceRequest(cpus="1"))
+    job.function(lambda: 1, resources=ResourceRequest(cpus="2"))
+    job.function(lambda: 1, resources=ResourceRequest(cpus="all"))
+    job_id = client.submit(job)
+    time.sleep(2.0)
```
Kobzol, line 71: We can use the existing functions that wait for a specific task state here.

Kobzol (changes requested):

crates/pyhq/python/hyperqueue/task/function.py
spirali: It is true, but user cannot create Python task without task_dir

crates/pyhq/python/hyperqueue/task/task.py
Kobzol: There is `TaskId` on line 5 of this file, right? You also added another here: [ffi/connection.py](https://github.com/It4innovations/hyperqueue/pull/389/files/674a93bda9f2191f12cb0859c2f83a46fea8642b#diff-163d04ce001322a9df03dd530d5e3bfefadc19b4e03c6e1a4055266007e4b19b).

spirali: Oh, I was looking in a wrong branch

PR #392, opened 2022-03-30
Dashboard: Restructure and added screen tabs

vyomkeshj:
This PR aims to restructure dashboard so that:

1. A group of widgets is called a Fragment, 
2. A Fragment lives inside a Screen tab.
3. Fragments communicate with their screens using the enums `FromFragmentMessage` and `ToFragmentMessage`

It also renames dashboard_state to root_screen.


crates/hyperqueue/src/dashboard/ui/screens/overview_screen.rs
```diff
@@ -0,0 +1,63 @@
...
+use termion::event::Key;
+use tui::layout::Rect;
+
+pub struct OverviewScreen {
+    overview_fragments: Vec<(ScreenTab, Box<dyn Fragment>)>,
+    active_fragment: usize,
+}
+
+impl Screen for OverviewScreen {
+    fn get_tabs(&self) -> &(Vec<ScreenTab>, usize) {
```
Kobzol: This return type is weird, it would be better to use an explicit type.

crates/hyperqueue/src/dashboard/ui/screen/mod.rs
```diff
@@ -1,13 +1,43 @@
...
 
+pub trait Fragment {
+    fn draw(&mut self, in_area: Rect, frame: &mut DashboardFrame);
+    fn update(&mut self, data: &DashboardData, controller: &mut ScreenController);
+    fn handle_message(&mut self, message: ToFragmentMessage);
+    fn handle_key(
+        &mut self,
+        key: Key,
+        controller: &mut ScreenController,
+    ) -> Option<FromFragmentMessage>;
```
Kobzol, line 19: Why is this so general? I would expect that each fragment is "private" for a screen and it can generate different events. If we standardize the messages for all fragments, then the screens would have to support all possible fragment messages.

vyomkeshj: > Why is this so general? I would expect that each fragment is "private" for a screen and it can generate different events. If we standardize the messages for all fragments, then the screens would have to support all possible fragment messages.

I wanted to be able to re-use fragments across screens. I guess one possible solution to having separate messages for the screens would be using a generic in the Fragment trait for the type of message and having different messages for each of the screens, what do you think?

crates/hyperqueue/src/dashboard/ui_loop.rs
```diff
@@ -21,17 +21,17 @@ pub async fn start_ui_loop(gsettings: &GlobalSettings) -> anyhow::Result<()> {
 
     // TODO: When we start the dashboard and connect to the server, the server may have already forgotten
     // some of its events. Therefore we should bootstrap the state with the most recent overview snapshot.
-    let data = DashboardData::default();
+    let dashboard_data = WrappedRcRefCell::wrap(DashboardData::default());
```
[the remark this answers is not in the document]
vyomkeshj: My bad, I'll remove the wrapper

PR #395, opened 2022-04-06
Less scattering scheduler

spirali:
Make scheduler less "scattering" in the initial task placement.
It helps to turn off more workers because of idle timeout.
It will also help to find workers for multi-node tasks.


crates/tako/src/scheduler/state.rs
```diff
@@ -83,8 +92,20 @@ impl SchedulerState {
...
         taskmap: &TaskMap,
         worker_map: &Map<WorkerId, Worker>,
+        try_stick: bool,
     ) -> Option<WorkerId> {
-        self.tmp_workers.clear();
+
+        // Fast path
+        if try_stick && task.inputs.is_empty() {
+            if let Some(worker_id) = self.get_last() {
+                if worker_map[&worker_id].have_immediate_resources_for_rq(&task.configuration.resources) {
```
Kobzol, line 101: Why does this use `have_immediate_resources_for_rq`, but below `is_capable_to_run` is used? The new task might be different from the previous one, right?

crates/tako/src/scheduler/state.rs
```diff
@@ -83,8 +92,20 @@ impl SchedulerState {
         task: &Task,
         taskmap: &TaskMap,
         worker_map: &Map<WorkerId, Worker>,
+        try_stick: bool,
```
Kobzol, line 95: Please document this parameter, it's far from obvious.

crates/tako/src/scheduler/state.rs
```diff
@@ -248,6 +269,8 @@ impl SchedulerState {
 
         let ready_tasks = core.take_ready_to_assign();
         if !ready_tasks.is_empty() {
+            // Find worker with minimum of tasks but at least one, otherwise just return any worker
```
Kobzol, line 272: I'm not sure that I understand this comment.

Also, can we move these hardcoded constants (like `1000`) into something like `const MIN_READY_TASKS_TO_STICK = ...` and put it on top of the scheduler? So that it easier to see all the "knobs".

crates/tako/src/scheduler/state.rs
spirali: Added comment into the code

crates/tako/src/scheduler/state.rs
spirali: Renamed and removed forgotten invalid comment

Kobzol (approved):

PR #396, opened 2022-04-06
WIP: Enable Timeline Scrubbing

vyomkeshj:
This PR aims to add timeline scrubbing to the dashboard:

When it's done:

1. One will be able to press <- -> arrow keys to change the time for which the data is shown on the dashboard.
2. Press 'l' to restore dashboard back to live mode.
3. Press +, - to change the duration of data that is shown on the dashboard charts.

### DIFF
--- a/crates/hyperqueue/src/dashboard/data/mod.rs
+++ b/crates/hyperqueue/src/dashboard/data/mod.rs
@@ -29,6 +29,14 @@ pub struct DashboardData {
     job_timeline: JobTimeline,
     /// Tracks the automatic allocator events
     alloc_timeline: AllocationTimeline,
+    /// Tracks the time range for which data has been fetched for.
+    fetched_range: FetchedTimeRange,
+}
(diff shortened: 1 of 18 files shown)


Kobzol:
The concept looks nice so far. I wonder if we should first add reading log events from a file? It will require changes that will also be relevant for the timeline ("offline" vs "live" mode) and it could also be useful for testing this feature (so that you can pre-load an automatically generated list of events and then test the timeline).

vyomkeshj:
> The concept looks nice so far. I wonder if we should first add reading log events from a file? It will require changes that will also be relevant for the timeline ("offline" vs "live" mode) and it could also be useful for testing this feature (so that you can pre-load an automatically generated list of events and then test the timeline).

Yes, lets postmortem event loading first. I'll fix the problems with this one in parallel

Kobzol:
Closing for inactivity.

PR #397, opened 2022-04-06
OMP_NUM_THREADS set in tasks

spirali:


Kobzol:
Good idea. Since we know the corresponding core IDs, maybe we can do all the way and pin OpenMP to them? https://www.openmp.org/spec-html/5.0/openmpse53.html (`OMP_PLACES`).

spirali:
Good idea, the question what should be default. If we should directly set OMP_PLACES or introduce something like --pin=omp

Kobzol:
That actually sounds really useful! We could have `--pin=taskset` or `--pin=omp`. Using both is probably not very useful.

Kobzol (approved):

PR #398, opened 2022-04-07
Add OpenMP task pin mode

Kobzol:


crates/hyperqueue/src/transfer/messages.rs
```diff
@@ -36,10 +36,16 @@ pub enum FromClientMessage {
     MonitoringEvents(MonitoringEventRequest),
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub enum PinMode {
+    TaskSet,
+    OpenMP,
+}
```
spirali, line 43: Maybe introduce PinMode::None to avoid Optioning it all over the place? In some sense "None pinning" is a kind of pinning mode.

crates/hyperqueue/src/worker/start.rs
```diff
@@ -95,14 +95,14 @@ impl TaskLauncher for HqTaskLauncher {
...
                 task_dir,
                 job_id,
                 task_id,
             } = body;
 
-            if pin {
-                pin_program(&mut program, allocation);
+            if let Some(pin_mode) = pin_mode {
+                pin_program(&mut program, allocation, pin_mode);
                 program.env.insert(HQ_PIN.into(), "1".into());
```
spirali, line 106: Maybe storing a name of the mode into HQ_PIN variable?

crates/hyperqueue/src/transfer/messages.rs
Kobzol: Maybe I would be inclined to do that if we indeed did some crazy stuff with Options, but this value is just passed verbatim through 2 places and then one `if let` is used on it. There are no crazy `map` shenanigans or anything, so I would rather keep it as it is.

spirali: It still seems to me more easier to understand flat branching that 2 level branching. But ok, I do not have a strong opinion, Btw: It adds extra 4 bytes per task definition:)

Kobzol: `bincode` serialization:
```
PinMode::OpenMP -> 1 byte
Some(PinMode::OpenMP) -> 2 bytes
<Option<PinMode>>::None -> 1 byte
```

crates/hyperqueue/src/worker/start.rs
Kobzol: Good idea.

spirali (approved):

PR #401, opened 2022-04-19
Serialize MonitoringEvents in Chunks

vyomkeshj:
This will allow us to efficiently read a range of events from the file without reading the entire file.

Dashboard needs to query only the range of events that are currently visible on it.

### DIFF
--- a/crates/hyperqueue/src/server/event/log/read.rs
+++ b/crates/hyperqueue/src/server/event/log/read.rs
@@ -1,21 +1,34 @@
+use crate::server::event::log::write::EventChunkHeader;
 use crate::server::event::log::{canonical_header, LogFileHeader};
-use crate::server::event::MonitoringEvent;
+use crate::server::event::{MonitoringEvent, MonitoringEventId};
 use anyhow::anyhow;
 use flate2::read::GzDecoder;
 use rmp_serde::decode::Error;
 use std::fs::File;
-use std::io::ErrorKind;
+use std::io::{Seek, SeekFrom};
+use std::os::unix::fs::FileExt;
 use std::path::Path;
 
 /// Reads events from a file in a streaming fashion.
 pub struct EventLogReader {
-    source: GzDecoder<File>,
+    source: File,
+
+    event_chunks: Vec<(MonitoringEventId, EventChunk)>,
+    current_chunk_index: usize,
+
+    current_decoded_chunk: Vec<MonitoringEvent>,
+    index_in_decoded_chunk: usize,
+}
+
+struct EventChunk {
+    begin_index: usize,
+    size: usize,
 }
 
 impl EventLogReader {
     pub fn open(path: &Path) -> anyhow::Result<Self> {
-        let mut file = File::open(path)?;
-        let header: LogFileHeader = rmp_serde::from_read(&mut file)
+        let mut source = File::open(path)?;
+        let header: LogFileHeader = rmp_serde::from_read(&mut source)
             .map_err(|error| anyhow!("Cannot load HQ event log file header: {error:?}"))?;
 
         let expected_header = canonical_header();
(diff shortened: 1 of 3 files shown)


crates/hyperqueue/src/server/event/log/write.rs
```diff
@@ -1,18 +1,28 @@
...
 pub struct EventLogWriter {
-    file: GzipEncoder<File>,
+    file: File,
     buffer: Vec<u8>,
+
+    /// The ID of the first event in current buffer.
+    first_event_in_buffer: MonitoringEventId,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
```
Kobzol, line 19: Since this is also used for reading, I would put it into `mod.rs`.

crates/hyperqueue/src/server/event/log/write.rs
```diff
@@ -1,18 +1,28 @@
...
+pub struct EventChunkHeader {
+    pub first_event_id: MonitoringEventId,
+    pub chunk_length: usize,
```
Kobzol, line 22: We should document what does this length mean (is it compressed/uncompressed size? is it in bytes or event count?). Either by comment (preferred), variable name or type.

crates/hyperqueue/src/server/event/log/write.rs
```diff
@@ -21,17 +31,21 @@ impl EventLogWriter {
...
+        Ok(Self {
+            file,
+            buffer,
+            first_event_in_buffer: 0,
+        })
     }
 
     #[inline]
     pub async fn store(&mut self, event: MonitoringEvent) -> anyhow::Result<()> {
+        if self.buffer.is_empty() {
```
Kobzol, line 44: What happens if we flush/finish the writer without storing a single event? Then the chunk header would say that the first event ID is 0, which is wrong. We should model this as an enum, something like 
```rust
enum WriterState {
    EmptyChunk(Vec<u8>),
    ChunkInProgress {
        first_event_id: MonitoringEventId,
        buffer: Vec<u8>
    }
}
```
and define corresponding state transition methods between these two variants. Storing an event while empty chunk => go to chunk in progress. Flushing => if not empty chunk, go from chunk in progress to empty chunk.

The only reason to include the `Vec<u8>` in the `EmptyChunk` variant is so that we can reuse its allocated buffer. Otherwise we would need to recreate it each time we go from `EmptyChunk` to `ChunkInProgress`.

crates/hyperqueue/src/server/event/log/write.rs
```diff
@@ -52,14 +66,26 @@ impl EventLogWriter {
         Ok(())
     }
 
-    async fn write_buffer(&mut self) -> tokio::io::Result<()> {
-        self.file.write_all(&self.buffer).await?;
+    async fn write_buffer(&mut self) -> anyhow::Result<()> {
+        let mut buffer_gzipped =
+            GzipEncoder::with_quality(Vec::with_capacity(CHUNK_MAX_SIZE * 2), Level::Fastest);
```
Kobzol, line 71: We can also reuse the buffer for gzip, same as we do for msgpack.

crates/hyperqueue/src/server/event/log/write.rs
```diff
@@ -52,14 +66,26 @@ impl EventLogWriter {
...
+
+        let _ = buffer_gzipped.write_all(&self.buffer).await?;
```
Kobzol, line 73: No need for `let _`, just use `?`.

crates/hyperqueue/src/server/event/log/write.rs
```diff
@@ -52,14 +66,26 @@ impl EventLogWriter {
...
+        let _ = buffer_gzipped.flush().await;
```
Kobzol, line 74: This `flush` is not needed and just adds extra data by the compression codec. `shutdown`'s documentation claims that it does `flush` implicitly.

crates/hyperqueue/src/server/event/log/write.rs
```diff
@@ -52,14 +66,26 @@ impl EventLogWriter {
...
+        let _ = buffer_gzipped.shutdown().await;
+
+        let chunk_header = rmp_serde::encode::to_vec(&EventChunkHeader {
+            first_event_id: self.first_event_in_buffer,
+            chunk_length: buffer_gzipped.get_ref().len(),
+        })?;
+        self.file.write_all(&chunk_header).await?;
```
Kobzol, line 81: These two writes can be merged, as discussed.

crates/hyperqueue/src/server/event/log/read.rs
```diff
@@ -1,21 +1,24 @@
...
-use std::io::ErrorKind;
+use std::io::{Cursor, ErrorKind, Seek, SeekFrom};
+use std::os::unix::fs::FileExt;
 use std::path::Path;
 
 /// Reads events from a file in a streaming fashion.
 pub struct EventLogReader {
-    source: GzDecoder<File>,
+    source: File,
+    current_chunk: GzDecoder<Cursor<Vec<u8>>>,
```
Kobzol, line 15: We don't event need to copy the data from the file into a vec, we could just read them directly from the file (I guess). I think that an enum that would store something like `NoChunk`, `ReadingChunk` could be useful here.

Kobzol:
Closing for inactivity.

PR #402, opened 2022-04-22
First version of multi node tasks

spirali:


crates/tako/src/messages/worker.rs
```diff
@@ -63,12 +65,12 @@ pub struct TaskIdMsg {
 }
 
 #[derive(Serialize, Deserialize, Debug)]
-//#[serde(tag = "op")]
 pub enum ToWorkerMessage {
     ComputeTask(ComputeTaskMsg),
     StealTasks(TaskIdsMsg),
     CancelTasks(TaskIdsMsg),
     NewWorker(NewWorkerMsg),
+    Reservation(bool),
```
Kobzol, line 73: What does this bool mean? Maybe use an enum?

crates/tako/src/scheduler/state.rs
```diff
@@ -188,23 +206,53 @@ impl SchedulerState {
...
+            if worker.is_reserved() {
+                let unreserve = worker
+                    .mn_task()
+                    .map(|mn| {
+                        let task = task_map.get_task(mn.task_id);
+                        task.mn_root_worker().unwrap() == worker.id
+                    })
+                    .unwrap_or(true);
+                if unreserve {
+                    worker.set_reservation(false);
```
Kobzol, line 233: It would be nice to create a function that calls set_reservation + sends the change to the worker, so that we don't forget (it's in multiple places).

crates/tako/src/scheduler/state.rs
```diff
@@ -188,23 +206,53 @@ impl SchedulerState {
...
+                    comm.send_worker_message(worker.id, &ToWorkerMessage::Reservation(false));
                 }
-                let task = tasks.get_task(task_id);
-                comm.send_worker_message(worker_id, &task.make_compute_message());
             }
         }
     }
 
+    pub(crate) fn assign_mn(
```
Kobzol, line 240: I think that `multinode` is not that long that we need to shorten it to `mn` :)

crates/tako/src/worker/rpc.rs
```diff
@@ -358,6 +354,12 @@ async fn worker_message_loop(
                     .insert(msg.worker_id, msg.address)
                     .is_none())
             }
+            ToWorkerMessage::Reservation(on_off) => {
+                state.reservation = on_off;
+                if !on_off {
+                    state.last_task_finish_time = Instant::now();
```
Kobzol, line 360: Now that it's reset on more places, I would create a method like `reset_idle_timer()`.

crates/tako/src/lib.rs
```diff
@@ -1,5 +1,7 @@
 #![deny(clippy::await_holding_refcell_ref)]
 
+extern crate core;
```
Kobzol, line 3: Probably caused by some autoimport, should be deleted.

crates/tako/src/messages/worker.rs
spirali: It is just on/off flag.

Kobzol: It's not clear from the type alone what should this do. It would be nice to add a separate type or at least a comment. Also the other worker message are verbs (Cancel, Compute, Steal), but this one is not (it should also be fixed for `NewWorker`.

Maybe at least `SetReservation`?

spirali: ``SetReservation`` sounds good

PR #408, opened 2022-05-03
Benchmark updates: summary page, comparison

User3574:
- HTML pages bokeh format
- Comparison of multiple benches

### DIFF
--- a/benchmarks/main.py
+++ b/benchmarks/main.py
@@ -1,16 +1,23 @@
 import dataclasses
 import itertools
 import logging
+import typer
+
 from pathlib import Path
 from typing import List, Optional
 
-import typer
 from src.benchmark.identifier import BenchmarkDescriptor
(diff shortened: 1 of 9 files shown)


Kobzol:
Some notes (data was generated using `python main.py compare-zw`):
- The cluster report page should show memory as the second chart. Currently it is shown as the fourth chart (it has changed since the layout changed from grid to two columns).
- If there's no profiling data in comparison page, the corresponding tab should be hidden:
image
- The `summary compare` command leaves the second column blank:
image
I checked that the second benchmark contains some monitoring data. Maybe we should just delete this command, it's quite difficult to use manually anyway. And it currently pre-generates cluster report HTMLs for all benchmarks.
- The summary page somehow duplicates data:
summary

Kobzol:
Thanks a lot for working on this!

PR #421, opened 2022-05-23
Tako structure refactoring

spirali:


crates/tako/src/public/server.rs
```diff
@@ -0,0 +1 @@
+
```
Kobzol, line 1: This file seems to be useless at this moment?

crates/tako/src/public/worker.rs
```diff
@@ -0,0 +1 @@
+
```
Kobzol, line 1: Also this one.

crates/tako/src/lib.rs
```diff
@@ -1,12 +1,22 @@
...
+#[macro_use]
+pub mod internal;
+
+pub mod comm;
+pub mod gateway;
+pub mod hwstats;
+pub mod launcher;
+pub mod program;
+
+pub use crate::internal::common::index::{AsIdVec, ItemId};
```
Kobzol, line 16: This seems to be a bit duplicated with `public/mod.rs`?

crates/tako/src/lib.rs
```diff
@@ -16,8 +26,36 @@ define_id_type!(InstanceId, u32);
 pub type Priority = i32;
 pub type PriorityTuple = (Priority, Priority); // user priority, scheduler priority
 
-pub type Error = crate::common::error::DsError;
+pub type Error = internal::common::error::DsError;
 pub type Result<T> = std::result::Result<T, Error>;
 
-#[cfg(test)]
-pub mod tests;
+pub mod resources {
```
Kobzol, line 32: These reexports also seem to be duplicated with the `public` directory. Maybe the `public` directory is now useless?

crates/tako/src/lib.rs
```diff
@@ -1,12 +1,22 @@
 #![deny(clippy::await_holding_refcell_ref)]
 
+/*#[cfg(not(bench))]
```
Kobzol, line 3: We could add `#[doc(hidden)]`, but we don't have a lot of docs anyway :)

Kobzol (changes requested):

Issue #425, opened 2022-05-24
Incorrect number of autodetected GPUs


Nortamo:
This is not that urgent/important as manually specifying the gpu count is easy. Currently HQ will auto detect all GPUs on a node, instead of  just the once which we have access to. 

One could use `nvidia-smi` to get the available gpus, or then check it manually,
with something like (mock in python as I'm not that familiar with rust.):

```python
import os
import glob

num_total_gpus=len(glob.glob("/proc/driver/nvidia/gpus/*"))
for n in range(0,num_total_gpus):
    try:
        f= open("/dev/nvidia" + str(n),'r')
        print("Access to /dev/nvidia" + str(n))
    except IOError:
        print("No perm to /dev/nvidia" + str(n))
        continue
```

It's also possible that some additional changes are required then to get gpu placement/id correct and possible numa affinities setup (or is this something you are planning to support / is already supported,  the documentation was a bit scares on this point. )when running on a partial node

Kobzol:
Interesting. Is there some proper defined way of finding out which GPUs are actually enabled for the user? Trying to open `/dev/nvidia` seems a bit "fishy".

In general, we tried to make the detection as simple as possible (it's really a "best effort" attempt), mainly to avoid potential C dependencies. We could use https://docs.rs/nvml-wrapper/latest/nvml_wrapper/, but we would need to make sure that it is optional (I'm not sure how easy that is), as we are currently strongly opposed to depending on any external dependencies other than `(g)libc`.

Kobzol:
Could you please provide some more context? On which cluster does this happen, what is the Nvidia/CUDA configuration used there? Is there e.g. some environment variable that could be used to distinguish which GPUs are available?

Nortamo:
We have roughly the same setup on three clusters 
- Puhti (Finnish national system) (V100,   )
- Mahti (Finnish national system) (A100)
- LUMI (testing on A40:s currently )

All of which are Slurm cluster. Slurm has a setting
to constrain devices using device cgroups (this is the thing causing `EPERM` when opening `/dev/nvidia<num>`) ( https://slurm.schedmd.com/cgroup.conf.html ), other schedulers might have similar settings.  If not configured to use cgroups, slurm (and probably other schedulers) will just rely on CUDA_VISIBLE_DEVICES to direct users to the correct devices. Although if possible I think sites are moving away from this as the variable is user-writable. 

[...]

PR #428, opened 2022-05-25
Automatically detect available GPUs from `CUDA_VISIBLE_DEVICES`

Kobzol:
The env. variable has a precedence over procfs.

Related issue: https://github.com/It4innovations/hyperqueue/issues/425

### DIFF
--- a/crates/hyperqueue/src/worker/hwdetect.rs
+++ b/crates/hyperqueue/src/worker/hwdetect.rs
@@ -1,14 +1,15 @@
 use crate::common::parser::{consume_all, p_u32, NomResult};
 
-use tako::resources::{CpuId, CpusDescriptor, GenericResourceDescriptor, NumOfCpus};
-
+use crate::common::format::human_size;
 use nom::character::complete::{newline, space0};
 use nom::combinator::{map_res, opt};
 use nom::multi::separated_list1;
 use nom::sequence::{preceded, terminated, tuple};
 use nom::Parser;
 use nom_supreme::tag::complete::tag;
+use tako::format_comma_delimited;
 use tako::resources::cpu_descriptor_from_socket_size;
+use tako::resources::{CpuId, CpusDescriptor, GenericResourceDescriptor, NumOfCpus};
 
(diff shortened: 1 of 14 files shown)


crates/hyperqueue/src/worker/hwdetect.rs
```diff
@@ -58,12 +62,26 @@ pub fn detect_generic_resources() -> anyhow::Result<Vec<GenericResourceDescripto
     }
 
     if let Ok(mem) = read_linux_memory() {
-        log::debug!("Mem detected: {}", mem);
+        log::info!("Detected {mem}B of memory ({})", human_size(mem));
```
spirali, line 65: ``human_size`` already adds units

crates/hyperqueue/src/worker/hwdetect.rs
```diff
@@ -58,12 +62,26 @@ pub fn detect_generic_resources() -> anyhow::Result<Vec<GenericResourceDescripto
...
 
+/// Tries to detect available Nvidia GPUs from the `CUDA_VISIBLE_DEVICES` environment variable.
+fn detect_gpus_from_env() -> Option<GenericResourceDescriptor> {
+    if let Ok(devices) = std::env::var("CUDA_VISIBLE_DEVICES") {
+        if let Ok(devices) = parse_comma_separated_numbers(&devices) {
+            log::info!(
+                "Detected GPUs {} from `CUDA_VISIBLE_DEVICES`",
+                format_comma_delimited(&devices)
+            );
+            return Some(GenericResourceDescriptor::list(GPU_RESOURCE_NAME, devices));
```
spirali, line 79: It probably needs some kind of sanity check that indices are unique, otherwise it will break the allocator.
(As I am thinking about it, it should be also done for generic list defined resources)

spirali (changes requested):

crates/hyperqueue/src/worker/hwdetect.rs
Kobzol: The `B` is added to the raw number, not to the formatted `human_size` output.

PR #432, opened 2022-05-27
Stream log export into json

spirali:


crates/hyperqueue/src/client/commands/log.rs
```diff
@@ -53,6 +60,9 @@ pub enum LogCommand {
 
     /// Prints a raw content of one channel
     Cat(CatOpts),
+
+    /// Export log into JSON
+    Export(ExportOpts),
```
Kobzol, line 65: Maybe it would be better if we just outputted JSON when the user combines `cat` with `--output-mode=json`?

crates/hyperqueue/src/stream/reader/logfile.rs
```diff
@@ -181,6 +191,64 @@ impl LogFile {
...
+                .iter()
+                .map(|(&task_id, task_info)| (task_id, task_info.last_instance()))
+                .collect(),
+        })
+    }
+
+    pub fn export(&mut self, opts: &ExportOpts) -> anyhow::Result<()> {
+        let task_infos = Self::_gather_infos(&self.index, &opts.task)?;
+        let mut result = Vec::new();
+        self.file.seek(SeekFrom::Start(self.start_pos))?;
```
Kobzol, line 220: This could be extracted into a function, like `self.rewind_to_start()`.

crates/hyperqueue/src/stream/reader/logfile.rs
```diff
@@ -181,6 +191,64 @@ impl LogFile {
...
+        let mut last_pos: i64 = self.start_pos as i64;
+        let mut buffer = Vec::new();
+        for (task_id, instance) in &task_infos {
```
Kobzol, line 223: This seems to be duplicated with stream reading in `cat`. It would be nice to create an `Iterator` over the chunks.

PR #434, opened 2022-05-31
Server UID + placeholder %{SERVER_UID}

spirali:


CHANGELOG.md
```diff
@@ -16,7 +16,11 @@ directory using the `HQ_SERVER_DIR` environment variable.
 detect available GPU resources from the `CUDA_VISIBLE_DEVICES` environment variable.
 
 ### Stream log
-* Basic export of stream log into JSON (``hq log <log_file> export`)
+* Basic export of stream log into JSON (`hq log <log_file> export`)
+
+### Server
+* Server now generates a random UID string each time ``server start`` is called.
```
Kobzol, line 22:
```suggestion
* Server now generates a random unique ID (UID) string every time a new server is started (`$hq server start`).
```

crates/hyperqueue/src/common/serverdir.rs
```diff
@@ -144,6 +144,10 @@ pub struct AccessRecord {
     /// Version of HQ
     version: String,
 
+    /// UID of the server instance
+    #[serde(default)]
```
Kobzol, line 148: Do we need default? We don't guarantee backwards compatibility with older HQ versions for the access record, right?

crates/hyperqueue/src/server/bootstrap.rs
```diff
@@ -90,19 +92,27 @@ pub async fn get_client_connection(server_directory: &Path) -> anyhow::Result<Cl
...
 
-    if HqConnection::connect_to_server(&record).await.is_err() {
+    if ClientSession::connect_to_server(&record).await.is_err() {
         return Ok(ServerStatus::Offline(record));
     }
 
     Ok(ServerStatus::Online(record))
 }
 
+fn generate_server_uid() -> String {
```
Kobzol, line 108: We could also do something like `Docker` and generate short names like `religious-squirrel` or `bountiful-bear` or something like that :)

crates/hyperqueue/src/common/serverdir.rs
spirali: It ends later on an invalid version. I wanted to avoid some ugly message that access.json cannot be parsed while we already have a reasonable error message

crates/hyperqueue/src/server/bootstrap.rs
spirali: I want to use it in paths and I want to have rather shorter paths:)

PR #440, opened 2022-06-10
Streaming architecture image

spirali:


docs/jobs/streaming.md
```diff
@@ -15,6 +15,11 @@ network to the server, which will continuously append it to a single file called
 
     In this section, we refer to `stdout` and `stderr` as **channels**.
 
+
+<p align="center">
+<img width="600" src="../imgs/streaming.png">
```
Kobzol, line 20: This link is wrong, it should be `../../imgs/streaming.png`, since `mkdocs` adds `/` at the end of the URL.

Kobzol (changes requested):

Issue #426, opened 2022-05-24
Task output is still too wide in practice


spirali:
Task output is still too wide in practice

So make it less wide.

Random ideas:
* Error message on new row rather column? (This can be especially problematic with custrom errors, e.g. from Python API, that can be long and multiline)
* If stdout/stderr is relative to workdir, then just use some placeholder shortcut?
* Hostname column seems to deplete to much space

![image](https://user-images.githubusercontent.com/263349/170047844-dc255c3c-aa52-4fe9-b2cf-1ee430cc9d82.png)

Kobzol:
I'm not sure how to make the table less wide yet (maybe split each row into multiple rows), but I'd rather not introduce placeholders here. The goal of resolving placeholders in this table was to allow the user to easily copy the path and then read it/work with it. Introducing placeholders would make this harder.

spirali:
Unfortunatelly, it is still too wide in practise :)

![image](https://user-images.githubusercontent.com/263349/180240792-dae7f3aa-9df5-42b5-8107-b6908f4d06f2.png)

PR #444, opened 2022-06-24
Task list output verbosity

User3574:
Closes: #426

### DIFF
--- a/crates/hyperqueue/src/client/output/cli.rs
+++ b/crates/hyperqueue/src/client/output/cli.rs
@@ -34,6 +34,7 @@ use tako::resources::{
 };
 
 use crate::client::output::common::{resolve_task_paths, TaskToPathsMap};
+use crate::client::output::Verbosity;
 use crate::common::utils::str::{pluralize, select_plural};
 use crate::worker::hwdetect::MEM_RESOURCE_NAME;
 use crate::worker::start::WORKER_EXTRA_PROCESS_PID;
@@ -52,6 +53,7 @@ pub const TASK_COLOR_RUNNING: Colorization = Colorization::Yellow;
 pub const TASK_COLOR_INVALID: Colorization = Colorization::BrightRed;
 
 const TERMINAL_WIDTH: usize = 80;
+const ERROR_TRUNCATE_LENGTH: usize = 16;
 
 pub struct CliOutput {
     color_policy: ColorChoice,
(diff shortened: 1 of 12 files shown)


Kobzol:
For Unix tools it's common to select verbosity with short flags (`-v`, `-vv`, `-vvv`). I tried to look at how it can be done using `clap`, but it was a bit complicated, so I took the implementation from https://docs.rs/clap-verbosity-flag/latest/clap_verbosity_flag/ and implemented it here.

The default should be brief verbosity, otherwise it's not very useful, and it couldn't be disabled anyway with the short flags.

Could you please fix tests? They will need to use the verbose mode. Thanks.

Kobzol (changes requested):

Kobzol:
When the error is hidden, it might be confusing to the user, since we hide important information from him. Maybe it would be better if we always shown the error column, but if the output mode is not verbose, we truncate it. I suggest to truncate the error if it's longer than `X` (16) characters or if it contains a newline.

Then we should show some message that says "Use -v to display full error.", and place it either below the table or in the error column. @spirali thoughts?

spirali:
> When the error is hidden, it might be confusing to the user, since we hide important information from him. Maybe it would be better if we always shown the error column, but if the output mode is not verbose, we truncate it. I suggest to truncate the error if it's longer than `X` (16) characters or if it contains a newline.
> 
> Then we should show some message that says "Use -v to display full error.", and place it either below the table or in the error column. @spirali thoughts?

It seems like a good idea: hame some small fragment by default and full info by "-v".

crates/hyperqueue/src/client/output/cli.rs
```diff
@@ -632,6 +644,7 @@ impl Output for CliOutput {
         ]);
 
         self.print_horizontal_table(rows, header);
+        log::info!("Use -v to display full error.");
```
Kobzol, line 647:
```suggestion
        log::info!("An error message was truncated. Use -v to display the full error.");
```
+ only print this if at least one error was actually truncated.

crates/hyperqueue/src/client/output/cli.rs
```diff
@@ -603,8 +607,16 @@ impl Output for CliOutput {
...
                                 ("Stderr", stderr),
                             ]),
-                            match &task.state {
-                                JobTaskState::Failed { error, .. } => {
+                            match (verbosity, &task.state) {
+                                (Verbosity::Normal, JobTaskState::Failed { error, .. }) => {
+                                    let mut error_mut = error.clone();
+                                    if error_mut.len() >= ERROR_TRUNCATE_LENGTH {
+                                        error_mut.truncate(ERROR_TRUNCATE_LENGTH);
+                                        error_mut = error_mut.add("...");
```
Kobzol, line 615:
```suggestion
                                        error_mut.push_str("...");
```

Kobzol (changes requested):

Kobzol:
Great! :) Please rebase on master & squash commits and I'll merge it.

Kobzol (approved):

PR #445, opened 2022-06-24
Use chumsky and ariadne for better error messages

Kobzol:
This PR ports most user-facing parsers from `nom` to `chumsky`, with the main motivation to improve error messages. You can check the new output in tests.

I didn't port client resource requests in this PR, since they are pending on resource variants.

image


spirali (approved):
:heart:  Please put a note into CHANGELOG.

Issue #431, opened 2022-05-26
Kill all executing tasks on worker if worker crashes


Kobzol:
If a worker ends abruptly, its spawned children should be killed. Could be implemented using Linux process groups.

spirali:
It is also connected to #66, that a each task should be spawned into a process group and when it is canceled we should clean all processes

rostamn739:
@Kobzol How do you envision solving this with process groups? I could try to implement that in my spare time..

There was a nifty hack using DEATHSIG `prctl` in the child process but that won't help if the processes are forked from an interim thread

Kobzol:
I have been playing with this for a bit and it seems to be quite tricky. We'll probably do some best effort mechanism using process groups and setsid to kill the group/session leader when a sigterm is received on the worker.

I didn't know about DEATHSIG, I'll take a look, thanks.

Kobzol:
Fixed by https://github.com/It4innovations/hyperqueue/pull/514.

Issue #416, opened 2022-05-21
Send a signal to currently executing tasks when a worker ends


Kobzol:
When a worker ends abruptly and there are some tasks executing in it at that time, we should at least try to signal them that they will be stopped soon.

PR #447, opened 2022-06-24
Propagate signals from worker to its children

Kobzol:
This is an alternative to https://github.com/It4innovations/hyperqueue/pull/446.

Instead of using `DEATHSIG`, we create a background handler in the worker which propagates any received `SIGTERM`, `SIGINT` or `SIGHUP` signals to the worker children (=running tasks).

Advantages:
- Doesn't suffer from the performance regression caused by setting `DEATHSIG`.
- Propagates signals to the whole process group, which isn't done by `DEATHSIG` (that only handles the end of the parent process).

Disadvantages:
- A bit hacky (?).
- Doesn't handle SIGKILL.

Fixes: https://github.com/It4innovations/hyperqueue/issues/416
Fixes: https://github.com/It4innovations/hyperqueue/issues/431

### DIFF
--- a/tests/utils/wait.py
+++ b/tests/utils/wait.py
@@ -1,7 +1,7 @@
 import time
 from typing import List, Union
 
-DEFAULT_TIMEOUT = 5
+DEFAULT_TIMEOUT = 10
 
 
 class TimeoutException(BaseException):


Kobzol:
This has numerous issues and it's unclear whether it's a good idea at all. Putting it on ice for now.

Issue #443, opened 2022-06-23
Multiple workers-per-allocation not working


tsthakur:
So from what I understood `--workers-per-alloc <worker_count>` is used to run multiple workers on multiple nodes, but it doesn't seem to behave that way. For example if I want to run 5 calculations on one node, I use an automatic allocation with 1 worker per allocation which then launches 5 jobs (1 task per job) on 1 node. So this is working as expected. Now if I launch an allocation with 2 workers per allocation, I was expecting that my 10 calculations would run on 2 nodes, with each node having its own worker. But what happens is that  2 workers occupy 2 nodes, but any of the 10 calculations does not launch stating the following error - 

``srun: Job 748536 step creation temporarily disabled, retrying (Requested nodes are busy)``

Please note that the nodes are not in fact busy as there is one 'big' Slurm job running on these 2 nodes. But there is no process running on either nodes if I check with `top` or `htop`.

I may be completely misunderstanding the purpose of this `--workers-per-alloc <worker_count>` command. In which case, would it be possible to do what I am trying to do in some other way? Or is it a use case that is not planned to be supported?

Kobzol:
Hi :) The `--workers-per-alloc` specifies the maximum number of nodes which will be requested for a Slurm/PBS allocation. Unless you want to do something with multiple nodes, like running MPI code (which is not properly supported yet though in HQ) or there is some other reason why you want multi-node allocations, it shouldn't be really needed to use this flag.

That being said, if you do use it, it should enqueue a Slurm allocation with 2 nodes, and then create a worker on each of those nodes. It seems that in your situation, the second step (creating the workers) has failed. I'm not really sure why, I will try to investigate the error message that you have posted.

ramirezfranciscof:
> Hi :) The `--workers-per-alloc` flag specifies the maximum number of nodes which will be requested for a Slurm/PBS allocation. Unless you want to do something with multiple nodes, like running MPI code (which is not properly supported yet though in HQ) or there is some other reason why you want multi-node allocations, it shouldn't really be needed to use this flag.

Hey @Kobzol , just for clarification, you mean to say that running MPI is not supported across multiple nodes (i.e. worker communication) or is it not supported in multiple cores of a single node either? I have been trying to use it for this and came across some problems (like canceling jobs not cancelling the run of the underlying script-wrapped executable), but I'm not sure if it would make sense to report them if you don't support the use case yet.

Kobzol:
[...]

PR #448, opened 2022-06-27
Use `--overlap` flag when starting HQ workers in Slurm allocations.

Kobzol:
This should make it less likely that the resources allocated for these workers will
interfere with manual `srun` invocations executed by HQ tasks.

Fixes: https://github.com/It4innovations/hyperqueue/issues/443


spirali (changes requested):
Please update CHANGELOG; otherwise LGTM

spirali (approved):

Issue #306, opened 2022-01-14
Handle tasks that crash workers


Kobzol:
Currently, tasks will be retried if the worker on which they are computed crashes. Some tasks can actually be the source of the crash (e.g. they allocate too much memory, which causes the node of the worker to crash).

Such tasks should be identified and after they cause too many crashes, they should no longer be scheduled.

PR #449, opened 2022-06-28
Handle crashing tasks

User3574:
This PR adds a counter that remembers how many times was some task executing during a worker crash. This hints that the task may have been the source of the crash. If the counter reached too many hits, the task will be canceled to avoid endless worker-crash loops.

Closes: https://github.com/It4innovations/hyperqueue/issues/306

### DIFF
--- a/crates/hyperqueue/src/server/job.rs
+++ b/crates/hyperqueue/src/server/job.rs
@@ -18,6 +18,10 @@ use tako::Set;
 use tako::TaskId;
 use tokio::sync::oneshot;
 
+/// Threshold that is used to determine if task
+/// is causing crash of the worker(s).
+const CRASHING_TASK_THRESHOLD: u32 = 5;
+
 /// State of a task that has been started at least once.
 /// It contains the last known state of the task.
 #[derive(Serialize, Deserialize, Debug, Clone)]
@@ -116,6 +120,7 @@ pub struct Job {
     pub counters: JobTaskCounters,
 
     pub tasks: Map<TakoTaskId, JobTaskInfo>,
+    pub tasks_crash_counter: Map<TakoTaskId, u32>,
 
     pub log: Option<PathBuf>,
 
@@ -188,6 +193,7 @@ impl Job {
             completion_date: None,
             submit_dir,
             completion_callbacks: Default::default(),
+            tasks_crash_counter: Default::default(),
         }
     }
 
@@ -306,6 +312,20 @@ impl Job {
             for handler in self.completion_callbacks.drain(..) {
                 handler.send(self.job_id).ok();
             }
+        }
+    }
+
+    pub fn increment_task_crash_count(&mut self, tako_task_id: TakoTaskId) {
+        self.tasks_crash_counter
+            .entry(tako_task_id)
+            .and_modify(|count| *count += 1)
+            .or_insert(1);
+    }
+
+    pub fn is_often_crashing_task(&self, tako_task_id: TakoTaskId) -> bool {
+        match self.tasks_crash_counter.get(&tako_task_id) {
+            Some(&crashes) => crashes >= CRASHING_TASK_THRESHOLD,
+            None => false,
         }
     }
 
--- a/crates/hyperqueue/src/server/state.rs
+++ b/crates/hyperqueue/src/server/state.rs
@@ -252,7 +252,12 @@ impl State {
             .on_worker_added(msg.worker_id, msg.configuration);
     }
 
-    pub fn process_worker_lost(&mut self, msg: LostWorkerMessage) {
+    pub fn process_worker_lost(
+        &mut self,
+        state_ref: &StateRef,
+        tako_ref: &Backend,
+        msg: LostWorkerMessage,
+    ) {
         log::debug!("Worker lost id={}", msg.worker_id);
         let worker = self.workers.get_mut(&msg.worker_id).unwrap();
         worker.set_offline_state(msg.reason.clone());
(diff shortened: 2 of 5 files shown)


crates/hyperqueue/src/server/state.rs
```diff
@@ -118,6 +120,9 @@ impl State {
                 submission_date: job.submission_date,
             },
         );
+        for (task_id, _) in job.tasks.iter() {
```
Kobzol, line 123: Let's not pre-populate the map and make it sparse. In 99% of cases, there will be no entries in this map, so this would just waste memory. If an entry is not in the map, it means that there were no crashes.

tests/test_job.py
```diff
@@ -1249,3 +1249,22 @@ def test_zero_custom_error_message(hq_env: HqEnv):
         == "Error: Task created an error file, but it is empty"
     )
     # print(table)
+
+
+# def test_resubmit_lost_worker(hq_env: HqEnv):
```
Kobzol, line 1254: While it's nice to show a warning during resubmit, our main use-case here is to just stop re-running the task if it crashes worker too often. So if a task crashed a worker e.g. for 5 times, we should move it to the canceled/failed category, so that it's not started again.

It might be tricky to create a task that crashes workers in tests. Maybe we could create a bash script that is passed the PID of the worker that will execute the task (we know that beforehand in tests) and then the script could `kill` the PID to crash the worker.

crates/hyperqueue/src/server/state.rs
```diff
@@ -20,6 +20,8 @@ use tako::{define_wrapped_type, TaskId};
 pub struct State {
     jobs: crate::Map<JobId, Job>,
     workers: crate::Map<WorkerId, Worker>,
+    // Here we store how many times was each Task present at worker crash
+    crashing_task_ids: crate::Map<TakoTaskId, u32>,
```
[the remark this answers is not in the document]
spirali: My original idea was to put the counter into each task, but I like the idea of having a separate map as it would be mostly sparse in most cases (And in a case of mass fails, we probably consume more memory by storing error messages). 

I agree with @Kobzol that map should be stored in Job so we easily free memory related to Job once the job is removed.

crates/hyperqueue/src/server/job.rs
```diff
@@ -116,6 +116,7 @@ pub struct Job {
     pub counters: JobTaskCounters,
 
     pub tasks: Map<TakoTaskId, JobTaskInfo>,
+    pub crashing_tasks: Map<TakoTaskId, u32>,
```
Kobzol, line 119: `task_crash_counter` or something like that might be a better name.

crates/hyperqueue/src/server/job.rs
```diff
@@ -309,6 +311,22 @@ impl Job {
         }
     }
 
+    pub fn set_crashed_task(&mut self, tako_task_id: TakoTaskId) {
```
Kobzol, line 314: `increment_task_crash_count`

crates/hyperqueue/src/server/job.rs
```diff
@@ -309,6 +311,22 @@ impl Job {
...
+        self.crashing_tasks
+            .entry(tako_task_id)
+            .and_modify(|count| *count += 1)
+            .or_insert(1);
+    }
+
+    pub fn is_crashed_task(&self, tako_task_id: TakoTaskId) -> bool {
```
Kobzol, line 321: `is_often_crashing_task`

crates/hyperqueue/src/server/job.rs
```diff
@@ -309,6 +311,22 @@ impl Job {
...
+        match self.crashing_tasks.get(&tako_task_id) {
+            Some(&crashes) => {
+                crashes > 4
```
Kobzol, line 324: Let's create a constant from the `4` and put it somewhere on top of this file.

crates/hyperqueue/src/client/commands/submit/command.rs
```diff
@@ -687,6 +687,10 @@ pub struct JobResubmitOpts {
     /// Job that should be resubmitted
     job_id: u32,
 
+    /// Keep task(s) that were present at crash of the worker
+    #[clap(long)]
+    keep_problematic: bool,
```
Kobzol, line 692: Please remove the `resubmit` changes from this PR and focus this PR only on cancelling the tasks, that's the main objective.

crates/hyperqueue/src/server/state.rs
```diff
@@ -264,6 +264,12 @@ impl State {
         for task_id in msg.running_tasks {
             let job = self.get_job_mut_by_tako_task_id(task_id).unwrap();
             job.set_waiting_state(task_id);
+
+            // Increment crash counter for task and check
+            job.set_crashed_task(task_id);
+            if job.is_crashed_task(task_id) {
+                job.set_cancel_state(task_id, backend);
```
Kobzol, line 271: @spirali is it enough to call `set_cancel_state` here?

Kobzol (changes requested):

Kobzol (approved):
LGTM!

@spirali Please check the comment that I made about canceled state.

crates/hyperqueue/src/server/state.rs
spirali: No, it will just set the internal booking of HyperQueue but not inform Tako.
You should probably call ``cancel_tasks_from_callback`` from here.

crates/hyperqueue/src/server/state.rs
```diff
@@ -263,7 +268,12 @@ impl State {
 
         for task_id in msg.running_tasks {
             let job = self.get_job_mut_by_tako_task_id(task_id).unwrap();
-            job.set_waiting_state(task_id);
+            job.increment_task_crash_count(task_id);
+            if job.is_often_crashing_task(task_id) {
+                cancel_tasks_from_callback(state_ref, tako_ref, job.job_id, vec![task_id]);
```
spirali, line 273: For Tako scheduler it is better if task are cancelled atomically at once; hence I propose collecting task_ids for cancel first and then (if non-empty) trigger cancel_tasks_from_callback only once. But this is only a minor issue.

spirali (approved):

Kobzol:
@User3574 Would you try to add a changelog? :) You can check the existing changelog entries for inspiration.

Kobzol:
Thanks! Sadly a conflict was created in the meantime :sweat_smile: Please squash the commits into a single one, rebase over `main`, fix the conflict (just add both the entries to the changelog) and push it again.

Kobzol:
Thanks! Great PR.

PR #450, opened 2022-06-30
Fix: Resubmit empty task(s)

User3574:
Currently while trying to resubmit a job that has 0 tasks (filters) causes a panic. This PR handles this as a error.


Kobzol:
Thanks! I just added changelog.

PR #454, opened 2022-07-18
Update documentation for v0.11.0

lahwaacz:
The release v0.11.0 renamed "indices" to "range", see https://github.com/It4innovations/hyperqueue/releases/tag/v0.11.0

### DIFF
--- a/docs/deployment/allocation.md
+++ b/docs/deployment/allocation.md
@@ -74,13 +74,13 @@ creating a new allocation queue:
     === "PBS"
     
         ```bash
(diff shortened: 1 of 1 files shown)


lahwaacz:
Hmm, the last tip on the [gresources page](https://it4innovations.github.io/hyperqueue/stable/jobs/gresources/) seems to imply that `gpus` is special...

Kobzol:
Ah, you are correct :) There are some things done additionally for a resource named `gpus`. What I meant was that this wasn't really important for this specific documentation example, sorry. Anyway, using `gpus` there is better, so thanks for the change! :)

PR #459, opened 2022-07-21
PyAPI: Generic resources

spirali:


crates/pyhq/src/client/job.rs
```diff
@@ -25,6 +26,7 @@ use crate::{borrow_mut, run_future, ClientContextPtr, FromPyObject, PyJobId, PyT
 pub struct ResourceRequestDescription {
     n_nodes: NumOfNodes,
     cpus: String,
+    generic: Option<HashMap<String, u64>>,
```
Kobzol, line 29: Could this be just a dictionary, without the `Option`? It seems unnecessary. We can use `dataclass.field(default_factory=dict)` for initializing the `generic` field.

Kobzol (approved):

spirali:
Added also priorities because of LIGATE, however it probably conflicts with python doc PR.

tests/pyapi/test_job.py
```diff
@@ -157,3 +159,27 @@ def test_job_generic_resources(hq_env: HqEnv):
         "FINISHED",
         "WAITING",
     ]
+
+
+def test_task_priorities(hq_env: HqEnv):
+    (job, client) = prepare_job_client(hq_env)
+
+    priorities = list(range(-20, 20))
+    random.shuffle(priorities)
```
Kobzol, line 168: Could you maybe just put a doc comment here to say what does the test do? It's not obvious to me. I would also rather just fix the seed to have it more deterministic.

Kobzol (approved):

Issue #462, opened 2022-07-23
Add `hq worker wait` command


Kobzol:
Some use-cases start HQ workers and then wait until these workers are connected to the clusters. It would be useful to have a `hq worker wait <N>` command, which would wait until at least N workers are online.

PR #463, opened 2022-07-26
Wait on workers

User3574:
Closes: #462

### DIFF
--- a/crates/hyperqueue/src/client/commands/worker.rs
+++ b/crates/hyperqueue/src/client/commands/worker.rs
@@ -1,12 +1,13 @@
 use std::path::PathBuf;
 use std::time::Duration;
 
-use clap::Parser;
 use tako::resources::ResourceDescriptor;
-use tako::Map;
-use tempdir::TempDir;
-
 use tako::worker::{ServerLostPolicy, WorkerConfiguration};
+use tako::Map;
+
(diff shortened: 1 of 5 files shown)


crates/hyperqueue/src/client/commands/worker.rs
```diff
@@ -257,3 +258,42 @@ pub async fn stop_worker(session: &mut ClientSession, selector: IdSelector) -> c
...
+                Some(_) => offline += 1,
+            }
+        }
+        Ok((online, offline))
+    }
+
+    loop {
+        let (online, offline) = get_workers_status(session).await?;
+
+        if worker_count > online + offline {
```
Kobzol, line 288: This shouldn't be an error. The command should just wait until `N` workers are online. It doesn't care about offline workers at all.

Kobzol (changes requested):

Kobzol:
Good :) Please also add a changelog entry and I'll merge it.

Kobzol:
Uhh, there's a conflict now. It happens all the time lately, maybe we shouldn't write the changelogs directly in the PRs, what do you think @spirali?

@User3574 Please fix the conflict, thanks.

Kobzol (approved):

Kobzol:
Thanks! :)

Issue #458, opened 2022-07-20
Per task information


spirali:
When a job is submitted via Python interface, it may have different resource request per task. We are not providing a way how to inspect such informations. (e.g. ``hq task list <X>``) does not show any information about resource requests.

spirali:
The same problem is also with priorities.

spirali:
The priorities are still unsolved.

PR #464, opened 2022-07-27
hq task list, info

User3574:
Issue: #426, #458

This PR should serve as a base for additional (https://github.com/It4innovations/hyperqueue/issues/458) issues. Now it just solves a problem with a big task list output by moving some of the info into task info. Also the usage would be similar as is for a job (job list, job info).

### DIFF
--- a/crates/hyperqueue/src/client/task.rs
+++ b/crates/hyperqueue/src/client/task.rs
@@ -8,7 +8,7 @@ use crate::transfer::messages::{
     FromClientMessage, IdSelector, JobDetailRequest, TaskIdSelector, TaskSelector,
     TaskStatusSelector, ToClientMessage,
 };
-use crate::{rpc_call, JobTaskId};
+use crate::{rpc_call, JobId, JobTaskId};
 
 #[derive(clap::Parser)]
 pub struct TaskOpts {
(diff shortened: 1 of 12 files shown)


crates/hyperqueue/src/client/task.rs
```diff
@@ -32,7 +36,19 @@ pub struct TaskListOpts {
     pub verbosity: VerbosityFlag,
 }
 
-pub async fn output_job_tasks(
+#[derive(clap::Parser)]
+pub struct TaskInfoOpts {
+    /// Select specific job(s).
+    pub job_selector: JobSelectorArg,
```
Kobzol, line 42: We assume a single job here, right? So this should not allow passing multiple job IDs.

Kobzol (commented):
Could you please write a more detailed description on what this PR does/what it changes?

crates/hyperqueue/src/client/task.rs
User3574: Yes, you're right.
This PR should serve as a base for additional (#458) issues. Now it just solves a problem with a big task list output by moving some of the info into task info. Also the usage would be similar as is for a job (job list, job info).

spirali:
@Kobzol The goal is to provide detailed information per task (in the same fashion list/info as for jobs).

The change is trying to solve #458.

@User3574 LGTM. In task info, I propose vertical oriented table (as for job info) as more individual information per task would be needed.

Kobzol (approved):

PR #478, opened 2022-08-24
Added priorities to task info

User3574:
Closes: #458

### DIFF
--- a/tests/pyapi/test_job.py
+++ b/tests/pyapi/test_job.py
@@ -231,3 +231,17 @@ def test_resources_task(hq_env: HqEnv):
 
     table = hq_env.command(["task", "info", str(submitted_job.id), "0"], as_table=True)
     table.check_row_value("Resources", "cpus: 2 compact\nres0: 1")
+
+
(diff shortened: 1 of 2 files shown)


tests/pyapi/test_job.py
```diff
@@ -231,3 +231,17 @@ def test_resources_task(hq_env: HqEnv):
...
+
+def test_priority_task(hq_env: HqEnv):
+    (job, client) = prepare_job_client(hq_env)
+
+    job.program(
+        args=bash("echo Hello"),
+        priority=2,
+    )
+    submitted_job = client.submit(job)
+    wait_for_job_state(hq_env, 1, "WAITING")
```
Kobzol, line 244: Here the task can be executed, so we should wait for the `FINISHED` state (otherwise the task can finish too soon and we miss WAITING and the test fails, which is what happened on CI).

Kobzol (changes requested):

Kobzol:
Thanks!

Kobzol (approved):

Issue #480, opened 2022-08-24
Allow displaying more tasks in task info


Kobzol:
I think that it should be possible to select multiple tasks, e.g. `hq task info 1 2-3`. The task tables should just be printed below one another.

Also it would be nice if `JOB-ID` could be specified using the `last` placeholder. I don't think that we should enable querying across multiple jobs, since the task IDs might not be the same for them.

PR #481, opened 2022-08-25
Added multiple task selection for task info

User3574:
Closes: #480

### DIFF
--- a/crates/hyperqueue/src/client/task.rs
+++ b/crates/hyperqueue/src/client/task.rs
@@ -2,13 +2,13 @@ use crate::client::globalsettings::GlobalSettings;
 use crate::client::job::get_worker_map;
 use crate::client::output::{Verbosity, VerbosityFlag};
 use crate::common::arraydef::IntArray;
-use crate::common::cli::{JobSelectorArg, TaskSelectorArg};
+use crate::common::cli::{JobSelectorArg, SingleIdSelectorArg, TaskSelectorArg};
+use crate::rpc_call;
 use crate::transfer::connection::ClientSession;
 use crate::transfer::messages::{
-    FromClientMessage, IdSelector, JobDetailRequest, TaskIdSelector, TaskSelector,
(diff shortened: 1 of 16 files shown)


crates/hyperqueue/src/client/task.rs
```diff
@@ -87,13 +87,32 @@ pub async fn output_job_task_info(
     gsettings: &GlobalSettings,
     session: &mut ClientSession,
     job_id_selector: IdSelector,
```
Kobzol, line 89: Could you please create something like `SingleIdSelector` (and `SingleIdSelectorArg`) that only supports a single ID or Last? So that we have it checked in compile time and don't even offer users to enter invalid argument in CLI.

There are some other commands that take a single ID as e.g. u32. Could you please check if it would make sense to also use `SingleIdSelector` for some of them?

tests/test_task.py
```diff
@@ -42,3 +42,23 @@ def test_task_list_multi(hq_env: HqEnv):
...
+    hq_env.start_server()
+    hq_env.start_worker()
+
+    hq_env.command(["submit", "--array=5-7", "--", "bash", "-c", "hostname"])
+    wait_for_job_state(hq_env, 1, "FINISHED")
+
+    r = hq_env.command(["task", "info", "1", "5"], as_table=True)
+    assert r.get_row_value("Task ID") == "5"
+
+    r = hq_env.command(["task", "info", "1", "4-6"])
```
Kobzol, line 57: I think that it shouldn't be too hard to parse multiple tables? Could you try to take a look at the implementation of `as_table` and find if we can generalize it to multiple tables? Probably with some other function, to avoid forcing all users of `as_table` to treat the return value as a list.

Kobzol (changes requested):

Kobzol (commented):
It looks OK. Please rebase on `main`.

Kobzol (approved):
Great, thank you!

PR #487, opened 2022-08-29
Update

spirali:


scripts/print_vers.py
```diff
@@ -0,0 +1,19 @@
+# This scripts writes all places where version has to be updated when a new release is made
+
+import os
+
+ROOT_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+
+def check_file(path, prefix, case_sensitive=False):
```
Kobzol, line 7: It would be nicer if the scripts directly wrote the new version and also created a new git tag.

Kobzol (approved):

scripts/print_vers.py
spirali: Creating tag directly is not desirable. I want to first see that everything is working in CI and then create a tag.

But I agree that ideally it should directly update versions, but creating new version is still not that common that I want to spent a time to automate it fully:)

PR #491, opened 2022-09-02
New resource management

spirali:
* Some changes are summarized in docs and changelog
* ~~Benchmarks are not ported yet~~
* ~~PyHQ is not ported yet.~~


tests/test_resources.py
```diff
@@ -8,25 +8,26 @@
...
-        "1x4 cpus; fairy Sum(10001000); potato Range(1-12); shark List(1,2,3,5)"
+        "cpus 4x2; fairy 10001000; potato 12; shark 4"
     ]
 
     table = hq_env.command(["worker", "info", "1"], as_table=True)
+    print(table.get_row_value("Resources"))
     assert (
         table.get_row_value("Resources")
-        == "1x4 cpus\nfairy: Sum(10001000)\npotato: Range(1-12)\nshark: List(1,2,3,5)"
+        == "cpus: 4x2\nfairy: 10001000\npotato: 12\nshark: 4"
```
Kobzol, line 30: Why isn't `Range`/`Sum` printed here anymore?

docs/jobs/cresources.md
```diff
@@ -189,23 +206,20 @@ of configuration that you can specify:
...
-Several environment variables related to CPU management will be passed to tasks executed by
-HyperQueue:
 
-* `HQ_CPUS` - Comma-separated list of numeric core IDs assigned to the task.
-* `HQ_PIN` - It is set to `taskset` or `omp` (depending on the used pinning mode) if the task was
-  [pinned](#pinning).
-* `NUM_OMP_THREADS` - Set to the number of cores assigned for task. (For compatibility with OpenMP).
+### Disable Hyper Threading
+
+If you want to detect CPUs but ignores HyperThreading then ``--no-hyper-threading`` flag can be used.
```
Kobzol, line 218: `ignore`

docs/jobs/cresources.md
```diff
@@ -24,15 +25,35 @@ task would thus never be scheduled on a worker that has less than 8 CPUs.
 Note that this reservation exists on a logical level only. To ensure more direct mapping to physical
 cores, see [pinning](#pinning) below.
 
-### Requesting all CPUs
 
-You can also use the value `all` for the `--cpus` flag to ensure that the task will request all
-available CPUs of the worker where it is executed. This will make sure that there will not be any
-other tasks executing on the same worker at the same time.
+## CPUs are resource
```
Kobzol, line 29: CPUS are a resource

docs/jobs/resources.md
```diff
@@ -0,0 +1,207 @@
...
+You can specify the resource pools of a worker when you start it:
+
+```bash
+$ hq worker start --resource "<NAME1>=<DEF1>" --resource "<NAME2>=<DEF2>" ...
+```
+
+where `NAMEi` is a name (string ) of the `i`-th resource pool and `DEFi` is a definition of the
+`i-th` resource pool. You can define resource pools using one of the following formats:
+
+* `[<VALUE>, <VALUE>, ..., <VALUE>]` where `VALUE` is a non-negative integer. This will
```
Kobzol, line 64: A line is missing here.

docs/jobs/resources.md
```diff
@@ -0,0 +1,207 @@
...
+* `range(<START>-<END>)` where `START` and `END` are non-negative integers. This will create
+an indexed pool with numbers in the inclusive range `[START, END]`.
+* `sum(<SIZE>)` where `SIZE` is a positive integer. This will create a sum pool with the given
```
Kobzol, line 67: Same here (the entry is duplicated below).

crates/pyhq/python/hyperqueue/ffi/protocol.py
```diff
@@ -1,12 +1,24 @@
...
 
 
 @dataclasses.dataclass()
 class ResourceRequest:
     n_nodes: int = 0
-    cpus: str = "1"
-    generic: Dict[str, int] = dataclasses.field(default_factory=dict)
+    resources: Dict[str, Union[int, str]] = dataclasses.field(default_factory=dict)
+
+    def __init__(
```
Kobzol, line 10: Dataclasses shouldn't have an `__init__` method. I would turn this into a `@staticmethod` constructor. We can add an assert in `__post_init__` there there is a CPU resource.

Kobzol (commented):
It would be nice to have more doc comments on functions, as always :)

tests/test_resources.py
spirali: It is because full description of lists is enumeration of indices; that is long for Karolina (128 ids). So I would leave it as is now as default and add "-v" for some verbose mode that prints it in detail. For a user it is mostly important how many resources it have, not particular ids.

crates/pyhq/python/hyperqueue/ffi/protocol.py
spirali: I removed data class from resource request

tests/test_resources.py
Kobzol: Shouldn't we normalize these things anyway? Turn 0-127 into Range(128)? But it would have to work recursively for NUMA groups (Group(Range(16))), which is probably too annoying.

spirali: Maybe we can just do "ranging" only in the output, something like [[0-11], [12-24]] (it is actually implemented in task ids), but still I believe that for most user it is more important to see that it is "2x12", rather then a precise ID (at least as default view).

Kobzol: Ok, in most cases it will be range(N) anyway, i.e. numbers from 0 to N - 1.

Kobzol (approved):

PR #492, opened 2022-09-06
Web diff

fredykoci:
image
image

### DIFF
--- a/benchmarks/src/postprocessing/serve.py
+++ b/benchmarks/src/postprocessing/serve.py
@@ -10,10 +10,11 @@ from tornado import ioloop, web
 from tornado.ioloop import IOLoop
 import seaborn as sns
 from .monitor import create_page
-from .overview import pregenerate_entries, create_summary_page, create_comparer_page, render
+from .overview import pregenerate_entries, create_summary_page, create_comparer_page, render, summary_by_benchmark, \
+    two_level_summary
 from ..benchmark.database import Database
 from .report import ClusterReport
-from .common import create_database_df, groupby_workload
+from .common import create_database_df, groupby_workload, groupby_environment
 import os
 import matplotlib.pyplot as plt
 from PIL import Image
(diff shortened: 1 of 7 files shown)


benchmarks/src/postprocessing/serve.py
```diff
@@ -55,11 +62,104 @@ def get(self, key: str):
...
+            output = ""
+            tables[0].style.set_uuid("table0")
+            tables[1].style.set_uuid("table1")
+            output += tables[0].style.apply(color_0).set_uuid("table0").to_html()
+            output += tables[1].style.apply(color_1).set_uuid("table1").to_html()
+
+            plt.scatter(df['index'], df["duration"], s=10)
+            plt.xticks(list(range(max_index + 1)))
+            plt.ylabel("Duration [ms]")
+            plt.savefig("img.png")
```
Kobzol, line 87: This is a really hacky solution with storing the `img.png` file. The image should be either generated on demand or cached in memory.

benchmarks/src/postprocessing/serve.py
```diff
@@ -55,11 +62,104 @@ def get(self, key: str):
...
+            self.set_header("Content-type", "image/png")
+
+    class OverViewHandler(web.RequestHandler):
+        KEY_GROUPS = ["Grouped by workload:", "Grouped by environment:", "Grouped by benchmark:"]
+
+        def summary_reader(self):
+            output = {}
+            i = 0
+
+            with open("benchmark/zw/summary.txt") as fp:
```
Kobzol, line 110: We have a function that generates this summary, so we should reuse it and use the output directly without parsing this file, which might not even exist.

Kobzol (changes requested):

Kobzol:
Thanks.

Kobzol:
Formatting check failed, please reformat the code.

Kobzol (approved):

Issue #488, opened 2022-08-30
Automatic allocation behavior clarification / feature request


Nortamo:
Running `hyperqueue 0.12.0` (prebuilt downloaded from github) using slurm 21.08.7 on "Red Hat Enterprise Linux Server 7.9 "

Is it intended behavior that a HQ job submission will trigger a new allocation even tough there is 
sufficient resources ( running workers) already available, or is this some sort of bug?

I.e if I create an allocation with `--workers-per-alloc 2 --backlog 1 --max-worker-count 100 ` 
and submit a job, I will get an allocation each time I call `hq submit`. Possible that I might have miss-configured, I'm a bit uncertain about the backlog flag, and e.g setting that to 0 leads to a allocation which does not start any jobs. 

It might be a bit out of scope of what HQ is trying to do, but It would extremely useful if it were possible for 
new workers to only start when there are more jobs/tasks than currently "active" resources. This way HQ could be
used as an extremely dynamic scheduling engine sitting between the system batch scheduler and a very dynamic workflow manager. 
Kind of a perfect middle ground between workflow managers launching every task as separate batch job and running HQ + workflow manager inside a single batch allocation.

Kobzol:
Hi :) So there are a few things to unpack here.

> Possible that I might have miss-configured, I'm a bit uncertain about the backlog flag, and e.g setting that to 0 leads to a allocation which does not start any jobs.

Using a backlog of `0` should be a hard error, I will fix that, thanks.

>  I will get an allocation each time I call hq submit

This happens because submitting new tasks "wakes" the automatic allocator. But even if there was no submit, the new allocations would probably be submitted after some time anyway.

> Kind of a perfect middle ground between workflow managers launching every task as separate batch job and running HQ + workflow manager inside a single batch allocation.

This is definitely not out of scope for HQ, in fact this is one of the primary goals of HyperQueue, to be this kind of a tool. Automatic allocation is not required for that, but it's of course a big part of the ergonomics of using it.

However, being perfect here is very hard, since predicting the allocation requirements is basically an NP-hard problem where we have to somehow predict the future. The current implementation is actually quite basic, and doesn't do a lot of prediction, which is why it can sometimes allocate needlessly. On the other hand, there are situations where it's not really possible to do a better job.

[...]

PR #494, opened 2022-09-07
Improve behaviour of auto allocator when a new job arrives

Kobzol:
This is more of a hotfix than anything else, we'll need to change the behaviour of the autoallocator more substantially to resolve similar problems in a robust way.

Fixes: https://github.com/It4innovations/hyperqueue/issues/488


Kobzol:
I tested the fix manually, we need to build a different test infrastructure for testing these situations.

spirali (approved):

PR #498, opened 2022-09-16
Worker scheduler looks on uniqueness of a resource

spirali:


crates/tako/src/internal/worker/rqueue.rs
```diff
@@ -67,6 +127,15 @@ impl ResourceWaitQueue {
         task_map: &TaskMap,
         remaining_time: Option<Duration>,
     ) -> Vec<(TaskId, Allocation)> {
+        if self.need_recompute_resource_priorities {
```
Kobzol, line 130: Maybe we could do this when workers are added/removed and get rid of the bool? It would be slightly amortized here, but I would rather not put computation into the latency critical task starting when it's not needed. Worker connect/disconnect should be rare enough that we could just recompute it every time.

crates/tako/src/internal/worker/rqueue.rs
```diff
@@ -117,25 +186,30 @@ impl ResourceWaitQueue {
...
     use crate::internal::worker::test_util::ResourceQueueBuilder as RB;
     use crate::resources::{ResourceDescriptorItem, ResourceDescriptorKind};
-    use crate::{Map, PriorityTuple, TaskId};
+    use crate::{Map, Set, WorkerId};
 
     impl ResourceWaitQueue {
-        pub fn queues(
-            &self,
-        ) -> &Map<ResourceRequest, priority_queue::PriorityQueue<TaskId, PriorityTuple>> {
+        /*pub(crate) fn queues(&self) -> &Map<ResourceRequest, QueueForRequest> {
```
Kobzol, line 202: Please delete if unused.

Kobzol (commented):
LGTM except for the priority recomputation comment.

PR #499, opened 2022-09-27
Query into scheduler for autoalloc

spirali:
~~Note: Query is now work only for a single node tasks~~

### DIFF
--- a/crates/tako/src/gateway.rs
+++ b/crates/tako/src/gateway.rs
@@ -3,7 +3,7 @@ use serde::{Deserialize, Serialize, Serializer};
 use crate::internal::messages::common::TaskFailInfo;
 use crate::internal::messages::worker::WorkerOverview;
 use crate::internal::worker::configuration::WorkerConfiguration;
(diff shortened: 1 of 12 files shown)


crates/tako/src/gateway.rs
```diff
@@ -105,6 +105,36 @@ pub struct StopWorkerRequest {
...
 }
 
+#[derive(Serialize, Deserialize, Debug)]
+pub struct WorkerTypeQuery {
+    pub descriptor: ResourceDescriptor,
+    pub max_sn_workers: u32,            // For single-node tasks
+    pub max_worker_per_allocation: u32, // For multi-node tasks
+}
+
+/* Ask scheduler for the information about how
```
Kobzol, line 115: Could you please copy this comment or some shortened version also to the `compute_new_worker_query` function? I was quite lost until I got here during review.

Kobzol (commented):
I'm not sure yet if there won't be some issues when this is actually used in autoalloc, but for now it looks reasonable. I'll integrate it into autoalloc once it's merged.

PR #502, opened 2022-09-30
Send signal to task's process group when it is cancelled or timeouted

Kobzol:
This should clean up the task's process children, at least in the happy case.


spirali (approved):

Kobzol:
I checked benchmarks and it looks like this doesn't regress `sleep` nor zero worker. Which is a bit weird, since I would swear that using `pre-exec` by itself slowed things down a lot a few months ago. But maybe it was fixed in the meantime in Rust.

Issue #467, opened 2022-08-04
Add command to pause and resume automatic alloactor queue


Kobzol:
This would give the users a chance to pause new allocations momentarily, for example to change the submitted script or to avoid overloading the system.

PR #508, opened 2022-10-03
Add the ability to pause autoallocation queues

Kobzol:
Paused queues do not submit new allocations. This can be used for interactive experiments to avoid removing the queue.

Fixes: https://github.com/It4innovations/hyperqueue/issues/467

### DIFF
--- a/crates/hyperqueue/src/server/autoalloc/state.rs
+++ b/crates/hyperqueue/src/server/autoalloc/state.rs
@@ -19,7 +19,7 @@ use crate::Map;
 // Main state holder
 pub struct AutoAllocState {
     allocation_to_queue: Map<AllocationId, QueueId>,
-    queues: Map<QueueId, QueueState>,
+    queues: Map<QueueId, AllocationQueue>,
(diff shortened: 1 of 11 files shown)


crates/hyperqueue/src/server/autoalloc/state.rs
```diff
@@ -139,6 +152,18 @@ impl QueueState {
...
+    pub fn state(&self) -> &AllocationQueueState {
+        &self.state
+    }
+
+    pub fn pause(&mut self) {
+        self.state = AllocationQueueState::Paused;
+    }
+
+    pub fn resume(&mut self) {
+        self.state = AllocationQueueState::Running;
```
spirali, line 164: I probably do not understand this well, but I would expected here some kind of "wake up" that tests if we should not (almost) immediately ask for new allocations.

Kobzol: Good point. It's not strictly needed, but it's definitely useful. I added a refresh of the queue once it's resumed.

spirali (approved):

PR #510, opened 2022-10-05
Option --crash-limit

spirali:
Add `--crash-limit` flag to `hq submit`. It lets you control the maximum time a task might be present when its executing worker fails, before that task is cancelled.

### DIFF
--- a/crates/tako/src/internal/server/core.rs
+++ b/crates/tako/src/internal/server/core.rs
@@ -14,7 +14,7 @@ use crate::internal::server::taskmap::TaskMap;
 use crate::internal::server::worker::Worker;
 use crate::internal::server::workerload::WorkerResources;
 use crate::internal::server::workermap::WorkerMap;
-use crate::{TaskId, WorkerId};
+use crate::{Map, TaskId, WorkerId};
 
 pub(crate) type CustomConnectionHandler = Box<dyn Fn(ConnectionDescriptor)>;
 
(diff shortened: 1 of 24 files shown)


crates/tako/src/internal/server/core.rs
```diff
@@ -280,6 +286,7 @@ impl Core {
     /// it up.
     #[must_use]
     pub fn remove_task(&mut self, task_id: TaskId) -> TaskRuntimeState {
+        self.crash_counter.remove(&task_id);
```
Kobzol, line 285: Since this is in tako now, can't we just store the counter inside the `Task`?

crates/tako/src/internal/server/reactor.rs
```diff
@@ -133,10 +132,34 @@ pub(crate) fn on_remove_worker(
         worker_id,
         running_tasks
     );
-
     let _ = core.remove_worker(worker_id);
+
     comm.broadcast_worker_message(&ToWorkerMessage::LostWorker(worker_id));
+
+    // IMPORTANT: We have to announce error BEFORE we announce lost worker (+ running tasks)
```
Kobzol, line 139: Maybe we should fix HQ instead? As long as we pretend that Tako doesn't know anything about HQ :)

Kobzol: Ok. As long as the messages cannot go out of order, then I guess it's fine for this usecase.

crates/tako/src/internal/server/core.rs
Kobzol: It adds an additional hash map lookup after each task finishes though.. I would rather put it into `Task` and then optimize the size of `Task` more holistically somehow, I'll send a PR for that.

spirali: Ok, but this map is actually almost always empty, I would consider it as fast op.

spirali: crash counter was internalized

PR #525, opened 2022-11-01
Version changed to v0.13.0

spirali:


crates/pyhq/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "pyhq"
-version = "0.12.0"
+version = "0.13.0"
```
Kobzol, line 3: Maybe we should also move the version to a workspace property? I'm not sure why I haven't done that already, maybe there was some problem with it :thinking:

Kobzol (approved):

crates/pyhq/Cargo.toml
spirali: I agree. I will do it in the next version

Issue #540, opened 2023-01-11
Parse shebang line in shell scripts properly if they contain arguments.


ghuls:
Parse shebang line in shell scripts properly if they contain arguments, by splitting them and adding them as arguments.

`shebang_with_args.sh`:
```bash
#!/bin/bash -l

echo "${@}";
```

Now, it gives this error, as it tries to execute the whole line:
```
Error: Cannot execute "/bin/bash -l shebang_with_args.sh arg1 arg2": No such file or directory (os error 2) |                                                                                                                      
The program that you have tried to execute (`/bin/bash -l`) was not found.
```

Another example where an argument in the shebang is common, is in awk scripts:
`shebang_with_args_for_awk.sh`:
```awk
#!/usr/bin/awk -f

print "test";
```

And Perl scripts:
```
#! /usr/bin/perl -w
...
```

PR #542, opened 2023-01-12
Parse all shebang args in directive files

Kobzol:
Fixes: https://github.com/It4innovations/hyperqueue/issues/540


spirali (approved):
LGTM, please update changelog about this and previous fix.

PR #546, opened 2023-01-13
Stream bug

spirali:
It is hackish :see_no_evil: but it should work:)


Kobzol:
Indeed quite hacky :) Let's see if it works. You'll have to rebase on `main` first.

spirali:
> Indeed quite hacky :) Let's see if it works. You'll have to rebase on `main` first.

Rebased and it  seems to work

Kobzol (approved):

PR #556, opened 2023-02-13
Detect AMD GPUs from the environment

Kobzol:
Relevant issue: https://github.com/It4innovations/hyperqueue/issues/425

### DIFF
--- a/crates/hyperqueue/src/worker/hwdetect.rs
+++ b/crates/hyperqueue/src/worker/hwdetect.rs
@@ -86,24 +86,33 @@ pub fn detect_additional_resources(items: &mut Vec<ResourceDescriptorItem>) -> a
     Ok(())
 }
 
-/// Tries to detect available Nvidia GPUs from the `CUDA_VISIBLE_DEVICES` environment variable.
+pub const GPU_ENV_KEYS: &[&str; 3] = &[
+    "CUDA_VISIBLE_DEVICES",
+    "HIP_VISIBLE_DEVICES",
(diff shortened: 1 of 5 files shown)


crates/hyperqueue/src/worker/hwdetect.rs
```diff
@@ -86,24 +86,33 @@ pub fn detect_additional_resources(items: &mut Vec<ResourceDescriptorItem>) -> a
...
-            log::info!(
-                "Detected GPUs {} from `CUDA_VISIBLE_DEVICES`",
-                format_comma_delimited(&devices)
-            );
-
-            if !has_unique_elements(&devices) {
-                log::warn!("CUDA_VISIBLE_DEVICES contains duplicates ({})", devices_str);
+    GPU_ENV_KEYS.iter().find_map(|env_key| {
+        if let Ok(devices_str) = std::env::var(env_key) {
+            if let Ok(devices) = parse_comma_separated_values(&devices_str) {
```
spirali, line 99: What about ``if let Ok(devices) = std::env::var(env_key).map(parse_comma_separated_values)`` rather then two nested "if lets"?

spirali (approved):

crates/hyperqueue/src/worker/hwdetect.rs
Kobzol: I tried it now, but the error types are different, so it would require `map_err` or some other error conversions, it's not worth it.

PR #559, opened 2023-02-24
Retry connections to unreachable server with `localhost` if hostname matches

Kobzol:
This caused problems on LUMI, where the login node is not accessible through it's hostname.

### DIFF
--- a/crates/hyperqueue/src/transfer/connection.rs
+++ b/crates/hyperqueue/src/transfer/connection.rs
@@ -15,6 +15,7 @@ use tokio_util::codec::{Framed, LengthDelimitedCodec};
 
 use crate::common::error::error;
 use crate::common::serverdir::AccessRecord;
+use crate::common::utils::network::get_hostname;
 use crate::transfer::messages::{FromClientMessage, ToClientMessage};
 use crate::transfer::protocol::make_protocol_builder;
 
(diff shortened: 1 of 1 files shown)


crates/hyperqueue/src/transfer/connection.rs
```diff
@@ -134,6 +134,28 @@ impl ClientSession {
...
+    match TcpStream::connect(&address).await {
+        Ok(conn) => Ok(conn),
+        // 113 = EHOSTUNREACH on Linux. Replace with ErrorKind::HostUnreachable once it's stabilized
+        Err(error) if matches!(error.raw_os_error(), Some(113)) => {
+            let hostname = get_hostname(None);
+            if hostname == record.host() {
+                let address = format!("localhost:{}", record.server_port());
+                log::warn!(
+                    "Could not reach {address}. It's host matches the local hostname, retrying\
+with `address`.",
```
spirali, line 148: I do not understand the warning message. Do you mean retrying with ``localhost``?
(Btw: If it is common on some systems, maybe it should be just debug message to not spam a user that cannot do anything about it)

Kobzol: Oh, this should have been a format placeholder of course, fixed, thanks.
It only appears once per each client session, but you're right that it could be a bit annoying. Changed to debug log.

spirali (approved):

PR #567, opened 2023-03-25
Resource variants

spirali:
Resource variants + initial version of job file. Job file is just MVP and will be modified, it is included just for Resource Variants testing.

### DIFF
--- a/crates/tako/src/gateway.rs
+++ b/crates/tako/src/gateway.rs
@@ -42,12 +42,32 @@ impl Default for ResourceRequest {
     }
 }
 
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq, Hash, Clone)]
+pub struct ResourceRequestVariants {
+    pub variants: SmallVec<[ResourceRequest; 1]>,
+}
+
+impl ResourceRequestVariants {
+    pub fn min_time(&self) -> Duration {
+        self.variants
+            .iter()
+            .map(|rq| rq.min_time)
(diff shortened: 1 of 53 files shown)


crates/tako/src/gateway.rs
```diff
@@ -42,12 +42,32 @@ impl Default for ResourceRequest {
...
+pub struct ResourceRequestVariants {
+    pub variants: SmallVec<[ResourceRequest; 1]>,
+}
+
+impl ResourceRequestVariants {
+    pub fn min_time(&self) -> Duration {
+        self.variants
+            .iter()
+            .map(|rq| rq.min_time)
+            .min()
```
Kobzol, line 69: I wonder if this shouldn't be max(), at least for autoalloc usage.

Kobzol (approved):
General comment: jobfiles and the Python interface probably has some duplication currently, I suppose that it's a known issue :)

crates/tako/src/gateway.rs
spirali: It depends how autoalloc is using it. What about requests (e.g. for gpus) that needs 10mins OR 1h (for cpus). Then it should be runnable on gpu queue that has timelimit >10min.

spirali:
> General comment: jobfiles and the Python interface probably has some duplication currently, I suppose that it's a known issue :)

I know, but I will create a separate serie of PRs with job file.

PR #571, opened 2023-03-31
Job Definition File

spirali:
Support for Job definition (without dependencies)


crates/hyperqueue/src/client/commands/submit/defs.rs
```diff
@@ -162,14 +148,64 @@ pub struct TaskDef {
...
+#[serde(deny_unknown_fields)]
+pub struct JobDef {
+    #[serde(default)]
+    pub name: String,
+
+    pub max_fails: Option<JobTaskCount>,
+
+    #[serde(default)]
+    #[serde(rename = "task")]
+    pub tasks: Vec<TaskDef>,
```
Kobzol, line 196: Shouldn't we define these mutually exclusive options with enums? To avoid the manual errors checks. Or did you do it to provide better error messages?

crates/tako/src/internal/common/resources/request.rs
```diff
@@ -151,9 +151,21 @@ impl ResourceRequestVariants {
...
+            .iter()
+            .enumerate()
+            .find_map(|(i, rq)| {
+                if allocator.is_capable_to_run(rq) {
+                    Some((i, rq.sort_key(allocator)))
+                } else {
+                    None
+                }
+            })
+            .unwrap();
```
Kobzol, line 165: This makes me a bit nervous. Is there always going to be at least a single variant for which `is_capable_to_run` returns true? And why do we skip the variants before it? It would be great to at least add a comment here.

docs/python/index.md
```diff
@@ -2,16 +2,6 @@
 To provide greater flexibility and support use-cases that are difficult to express using the CLI
 (such as dependencies between tasks), HyperQueue also offers a Python API for submitting complex jobs.
 
-There are several benefits to using the Python API, which are not easily available in the command-line
-interface:
-
-- [**Task dependencies**](dependencies.md) You can specify dependencies between individual
```
Kobzol, line 8: Well dependencies are still not in jobfile, right? :) Same as with Python tasks.

docs/jobs/resources.md
```diff
@@ -232,3 +232,8 @@ $ hq submit --cpus=2 --array=1-10
 ```
 
 then each task will require two cores.
+
+## Resource variants
+
+A task may have attached more resource requests. There are no command line interface for
```
Kobzol, line 238: There are no -> There is no

docs/jobs/jobfile.md
```diff
@@ -0,0 +1,153 @@
...
+## Minimal example
+
+First, we create file with the following content:
+
+```toml
+[[task]]
+command = ["sleep", "1"]
+```
+
+Let us assume that we have named this file as ``myjob``,
```
Kobzol, line 23: I would name it with the `.toml` extension. Also, the name is not the same as in the command (`myjob` vs `myfile`).

docs/jobs/jobfile.md
```diff
@@ -0,0 +1,153 @@
...
+command = ["sleep", "2"]
+
+[[task]]
+id = 2
+command = ["sleep", "3"]
+```
+
+## Task arrays
+
+If you want to create uniform tasks you can define task array (similar to ``--array`):
```
Kobzol, line 104: Missing second backtick (you can also just wrap things with single backticks :)).

docs/jobs/jobfile.md
```diff
@@ -0,0 +1,153 @@
...
+    Options `entries` and `ids` can be used together.
+
+## Task dependencies
+
+TODO
+
+## Resource variants
+
+More resource configurations may be defined for a task.
+In this case, HyperQueue will takes into account all these configurations during scheduling.
```
Kobzol, line 131: will take

crates/tako/src/internal/common/resources/request.rs
spirali: I will add comment in the code.

spirali:
@Kobzol Dependencies added + docs fixed

Kobzol (approved):

Issue #577, opened 2023-04-20
hq job info output mode differences


svatosFZU:
Hi,
I am using "hq job info" to get some basic information about a job. I am using --output-mode json to make further processing easy (and because it contains start and finish timestamps). Unfortunately, there are some values available in cli mode which are not in json mode (and vice versa). To be specific, I miss the Resources (number of cores, etc.) and Workers (worker node name) info which is in cli but not in json. Would it be possible to add them also into the json mode? Also, if it would be possible to add ID of PBS/SLURM allocation into the info, that would be great.

```
hq job info --output-mode cli 7000
+----------------------+-------------------------+
| ID                   | 7000                    |
| Name                 | bash                    |
| State                | FINISHED                |
| Tasks                | 1; Ids: 0               |
| Workers              | cn601.karolina.it4i.cz  |
| Resources            | cpus: 32 compact        |
| Priority             | 0                       |
| Command              | bash                    |
| Stdout               | <None>                  |
| Stderr               | <None>                  |
| Environment          |                         |
| Working directory    | /home/svatosm           |
| Task time limit      | None                    |
| Crash limit          | 1                       |
| Submission date      | 2023-04-17 03:24:00 UTC |
| Submission directory | /home/svatosm           |
| Makespan             | 6h 22m                  |
+----------------------+-------------------------+
```

```
hq job info --output-mode json 7000
{
  "crash_limit": 1,
  "finished_at": "2023-04-17T09:46:56.088693387Z",
  "info": {
    "id": 7000,
    "name": "bash",
    "task_count": 1,
    "task_stats": {
      "canceled": 0,
      "failed": 0,
      "finished": 1,
      "running": 0,
      "waiting": 0
    }
  },
  "max_fails": null,
  "pin_mode": "None",
  "priority": 0,
  "program": {
    "args": [
      "bash"
    ],
    "cwd": "/home/svatosm",
    "env": {},
    "stderr": "Null",
    "stdout": "Null"
  },
  "resources": {
    "min_time": 0.0,
    "n_nodes": 0,
    "resources": []
  },
  "started_at": "2023-04-17T03:23:59.989407980Z",
  "submit_dir": "/home/svatosm",
  "task_dir": false,
  "tasks": [
    {
      "cwd": "/home/svatosm",
      "finished_at": "2023-04-17T09:46:56.088693387Z",
      "id": 0,
      "started_at": "2023-04-17T03:23:59.994031765Z",
      "state": "finished",
[...]

PR #579, opened 2023-04-20
Add more information to JSON output

Kobzol:
Fixes: https://github.com/It4innovations/hyperqueue/issues/577

I didn't add a changelog, because JSON output is still unstable.


Kobzol:
I think that consumers of the JSON output would prefer a machine readable output, they don't have the HQ parser available. They can always generate the text syntax from the machine readable one easily, but not necessarily vice-versa.

spirali:
> I think that consumers of the JSON output would prefer a machine readable output, they don't have the HQ parser available. They can always generate the text syntax from the machine readable one easily, but not necessarily vice-versa.

Ok. I do not have a strong opinion about it.

PR #582, opened 2023-04-21
Deps updates

spirali:


spirali:
hasbrown & dev dependencies not updated (I do not have energy to fix it:)

crates/hyperqueue/src/worker/parser.rs
```diff
@@ -564,7 +564,7 @@ mod test {
     #[test]
     fn test_parse_resource_def_numeric_value() {
         insta::assert_snapshot!(expect_parser_error(parse_resource_definition_inner(), "x=1"), @r###"
-        Unexpected end of input found while attempting to parse resource kind, expected _ or x:
```
Kobzol, line 567: This is an error regression :(

Kobzol (approved):

crates/hyperqueue/src/worker/parser.rs
spirali: Do you want to look at? Otherwise I will just revert to the older version.

Kobzol: Nah, let's just leave it as it is, the error messages are quite imprecise in other locations too, anyway.

spirali: Leave it in the updated version version?

Kobzol: Yeah let's keep 0.9, it should be faster to compile and includes some nice improvements.

Issue #584, opened 2023-05-01
resubmitted streaming jobs 


arashbm:
let's say I submit a job with multiple tasks, and some of those tasks fail. If I try to resubmit the failed tasks, the resubmitted job instantly fails.

```con
$ hq submit --array=0-4 --log=output.log bash some_script.sh
$ # a while later...
$ hq job resubmit 1 --filter=failed,canceled
```

The server logs:
```
ERROR Stream connection ended with: Job 2 is not registered for streaming
```

Expected behaviour, to me at least, is that the new job just appends to the original log.

Kobzol:
Hi! Thanks for posting this issue. The combination of output streaming and resubmission is not currently implemented at the moment. However, as you can see, the error message is not very good.

As a short term solution, we will improve the error message to let users know that this is not currently supported. In longer term, we plan to implement support for it (we were basically waiting for the first use-case that would need it, seems like you're the first one :) ).

Kobzol:
This is no longer relevant, because we have since removed resubmit completely.

spirali:
Let me note that the mechanism that is recommended now for resubmitting do not suffer with this kind error. However, as appending to log is not implemented, resubmit of the task will overwrite the log file.

PR #585, opened 2023-05-03
Output an explicit error message when resubmit is attempted for a job with a log

Kobzol:
Related issue: https://github.com/It4innovations/hyperqueue/issues/584

### DIFF
--- a/tests/test_job.py
+++ b/tests/test_job.py
@@ -765,10 +765,24 @@ def test_job_resubmit_empty(hq_env: HqEnv):
     hq_env.start_workers(2, cpus=1)
     wait_for_job_state(hq_env, 1, "FINISHED")
 
-    try:
-        hq_env.command(["job", "resubmit", "--filter=canceled", "1"], as_table=True)
-    except Exception as e:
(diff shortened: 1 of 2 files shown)


tests/test_job.py
```diff
@@ -771,6 +771,23 @@ def test_job_resubmit_empty(hq_env: HqEnv):
...
+        ["submit", "--array=1-10", "--log", "foo.bin", "--", "/bin/nonexisting"]
+    )
+    hq_env.start_workers(1)
+    wait_for_job_state(hq_env, 1, "FAILED")
+
+    try:
+        hq_env.command(["job", "resubmit", "1"], as_table=True)
+    except Exception as e:
+        assert (
+            "Resubmit is not currently supported when output streaming (`--log`) is used"
```
spirali, line 786: You can use 
```python
hq_env.command(..., expect_fail="Resubmit is not currently supported when output streaming (`--log`) is used")
```

Kobzol: Thanks, I changed it.

spirali (approved):

PR #613, opened 2023-08-01
Add automatic file cleanup after task success

Kobzol:
If a task succeeds, often we do not care about its stdout/stderr output. This PR adds a "file closing mode" to stdout/stderr, which makes HQ automatically remove the files from disk if their task succeeds.

Documentation and changelog isn't done yet, I first want to bakeshed the various names and flags :)

### DIFF
--- a/crates/hyperqueue/src/client/commands/submit/command.rs
+++ b/crates/hyperqueue/src/client/commands/submit/command.rs
@@ -6,6 +6,9 @@ use std::{fs, io};
 
 use anyhow::{anyhow, bail};
 use bstr::BString;
+use chumsky::primitive::{filter, just};
+use chumsky::text::TextParser;
+use chumsky::Parser as ChumskyParser;
 use clap::{ArgMatches, Parser};
 use smallvec::smallvec;
 use tako::gateway::{
@@ -23,6 +26,7 @@ use crate::client::resources::{parse_allocation_request, parse_resource_request}
 use crate::client::status::Status;
 use crate::common::arraydef::IntArray;
 use crate::common::cli::OptsWithMatches;
+use crate::common::parser2::{all_consuming, CharParser, ParseError};
 use crate::common::placeholders::{
     get_unknown_placeholders, parse_resolvable_string, StringPart, CWD_PLACEHOLDER,
     JOB_ID_PLACEHOLDER, TASK_ID_PLACEHOLDER,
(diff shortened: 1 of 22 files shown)


crates/hyperqueue/src/client/commands/submit/command.rs
```diff
@@ -878,4 +947,75 @@ mod tests {
...
+        )
+    }
+
+    #[test]
+    fn stdio_path_and_mode() {
+        assert_eq!(
+            parse_stdio_def("foo/bar/baz.txt:cleanup").unwrap(),
+            StdioDefInput::File {
+                path: Some(PathBuf::from("foo/bar/baz.txt")),
+                on_close: FileOnCloseBehavior::DeleteIfSuccess
```
spirali, line 976: Why not DeleteIfFinished? "Success" is not name of the task state

crates/hyperqueue/src/client/commands/submit/defs.rs
```diff
@@ -345,4 +398,39 @@ mod test {
...
+        );
+    }
+
+    #[test]
+    fn test_parse_stdio_struct() {
+        let r = JobDef::parse(
+            r#"
+        [[task]]
+        command = ["sleep", "1"]
+        stdout = { path = "foo", on_close = "delete-if-success" }
```
spirali, line 424: Again, why not delete-if-finished?

spirali (changes requested):
* Doc and changelog is missing
* I do not like inconsistency between "cleanup" and "delete-if-success". How you want to name keyword for removing empty files, as it will be also kind of "cleanup"?
* I do not like name delete-if-success, I propose delete-if-finished as finished is an existing state.

Kobzol:
> I do not like inconsistency between "cleanup" and "delete-if-success". How you want to name keyword for removing empty files, as it will be also kind of "cleanup"?

I agree! But what name should be used on the CLI? Isn't `delete-if-finished` a bit too long? Maybe it's fine…

spirali:
Brainstorming:
* Shorten "delete" to just "del", i.e `del-if-finished` `del-if-empty`
* Consider the field always a delete rule, so we can just write `/my/path/stdout:finished` `/my/path/stdout:empty`
* Emphasis that file will be not there by using "no" prefix: `/my/path/stdout:no-finished` `/my/path/stdout:no-empty`
* Variant with "rm": `/my/path/stdout:rm-finished` `/my/path/stdout:rm-empty`
* Alternatively "rm-if": `/my/path/stdout:rm-if-finished` `/my/path/stdout:rm-if-empty`

spirali:
Looking at my examples, I'm starting to slightly prefer the last one.

Kobzol:
Renamed to `rm-if-finished` and added documentation and changelog.

docs/jobs/jobs.md
```diff
@@ -337,6 +337,27 @@ If the message is longer than 2KiB, then it is truncated to 2KiB.
...
+## Automatic file cleanup
+If you create a lot of tasks and do not use [output streaming](streaming.md), a lot of `stdout`/`stderr`
+files can be created on the disk. In certain cases, you might not be interested in the contents of these files, especially
+if the task has finished successfully, and you instead want to remove them as soon as they are not needed.
+
+For that, you can use a file cleanup mode when specifying `stdout` and/or `stderr` to choose what should happen with
+the file when its task finishes. The mode is specified as a name following a colon (`:`) after the file path.
+Currently, one cleanup mode is implemented:
+
+- Remove the file if the task has finished (successfully):
```
spirali, line 349: Maybe explicitly mention state "FINISHED" (+link) rather just word "finished". Or explicitly claim that it will be NOT deleted when task fails or is canceled.

Kobzol: Good point! Added both.

spirali (approved):

PR #626, opened 2023-10-02
Non-integer amount of resources

spirali:


crates/hyperqueue/src/client/resources.rs
```diff
@@ -79,85 +63,82 @@ mod test {
...
+            expect_parser_error(all_consuming(p_resource_request()), "a"),
+            "Unexpected end of input found, expected =:\n  a\n   |\n   --- Unexpected end of input\n",
         );
-        check_parse_error(
-            p_resource_request,
-            "a=x",
-            "Parse error\nexpected Resource amount at character 2: \"x\"\n  expected one of the following 2 variants:\n    expected \"all\" at character 2: \"x\"\n    or\n    expected integer at character 2: \"x\"",
+        assert_eq!(
+            expect_parser_error(all_consuming(p_resource_request()), "a=x"),
+            "Unexpected token found, expected all:\n  a=x\n    |\n    --- Unexpected token `x`\n",
```
Kobzol, line 141: The parse errors have regressed somewhat, as e.g. `ResourceAmount` isn't mentioned anymore. But let's not resolve that here.

crates/tako/src/internal/common/resources/amount.rs
```diff
@@ -0,0 +1,115 @@
+use derive_more::{Add, AddAssign, Sub, SubAssign, Sum};
+use serde::{Deserialize, Serialize};
+
+pub type ResourceUnits = u32;
+pub type ResourceFractions = u32;
+
+pub const FRACTIONS_PER_UNIT: ResourceFractions = 10_000;
+pub const FRACTIONS_MAX_DIGITS: usize = 4; // = log10(FRACTIONS_PER_UNIT)
```
Kobzol, line 8:
```rust
pub const FRACTIONS_MAX_DIGITS: usize = FRACTIONS_PER_UNIT.ilog10() as usize
```
?

crates/hyperqueue/src/common/parser2.rs
```diff
@@ -230,12 +231,24 @@ pub fn parse_u32() -> impl CharParser<u32> {
     })
 }
 
-/// Parse 8-byte integer.
-pub fn parse_u64() -> impl CharParser<u64> {
-    parse_integer_string().try_map(|p, span| {
-        p.parse::<u64>()
-            .map_err(|_| ParseError::custom(span, "Cannot parse as 8-byte unsigned integer"))
-    })
+pub fn parse_resource_amount() -> impl CharParser<ResourceAmount> {
```
Kobzol, line 234: Could you please add a comment above this function that explains what does this parse, along with an example of how could a valid input look like?

crates/hyperqueue/src/common/parser2.rs
```diff
@@ -283,6 +296,40 @@ mod tests {
...
+        );
+        assert_eq!(
+            parse_resource_amount().parse_text("100.9").unwrap(),
+            ResourceAmount::new(100, 9000)
+        );
+        assert_eq!(
+            parse_resource_amount().parse_text("0.346").unwrap(),
+            ResourceAmount::new(0, 3460)
+        );
+        dbg!(parse_resource_amount().parse_text("0.12345"));
```
Kobzol: Forgotten `dbg!`.

crates/hyperqueue/src/common/parser2.rs
```diff
@@ -283,6 +296,40 @@ mod tests {
...
+        assert!(parse_resource_amount().parse_text("0.12345").is_err());
```
Kobzol, line 329: It would be nice to use `insta` to see the actual error produced here.

crates/tako/src/internal/worker/resources/allocator.rs
```diff
@@ -120,103 +115,45 @@ impl ResourceAllocator {
...
+                    return true;
+                }
+                if *amount > max_alloc {
                     return false;
                 }
-                if *amount == 0 {
-                    return true;
+                let (units, fractions) = amount.split();
+                if fractions > 0 {
+                    todo!() // Fractions and ForceCompact is not supported in the current version
```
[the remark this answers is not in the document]
spirali: There is, it is part of validation of resource request.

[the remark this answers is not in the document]
Kobzol: Then this should be an assert instead of todo.

crates/tako/src/internal/worker/resources/allocator.rs
```diff
@@ -5,23 +5,23 @@ use crate::internal::common::resources::{
...
 pub struct ResourceAllocator {
     pools: ResourceVec<ResourcePool>,
-    free_resources: ResourceCountVec,
+    free_resources: ConciseFreeResources,
     remaining_time: Option<Duration>,
 
     blocked_requests: Vec<BlockedRequest>,
     higher_priority_blocked_requests: usize,
-    running_tasks: Vec<ResourceCountVec>, // TODO: Rework on multiset?
+    running_tasks: Vec<Rc<Allocation>>, // TODO: Rework on multiset?
```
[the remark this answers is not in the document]
spirali: Originaly there was two types of allocation "just description" (e.g.: 5 cpus) and full allocation (indices: [0,10,5,1,4]). Allocator returned full allocation and remembers only the light one. With fraction resources, allocator is no longer two-phased (first do a light allocation and after that full) [technicly it is possible, but the description 
of allocation cannot be just a number with fractional resources and it almost look like full alocation, so description allocations were removed].
So allocation is now single phased. Allocator now returns an allocation but also remembers the full alocation. To avoid coping, it is wrapped to to Rc. (It ale makes faster the dealocation since to find the right allocation, we need to just check pointers and not full allocations.

crates/tako/src/internal/server/workerload.rs
```diff
@@ -246,16 +249,20 @@ impl WorkerLoad {
...
                 return 0;
             }
             let load = self
                 .n_resources
                 .get(entry.resource_id)
                 .copied()
-                .unwrap_or(0);
-            result += ((load * 512) / (count * 512)) as u32;
+                .unwrap_or(ResourceAmount::ZERO);
+            result += (load.total_fractions() * 512 / count.total_fractions()) as u32;
```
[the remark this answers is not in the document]
spirali: It would be possible but here it is more like recomputing to a different precision fixed point number. So dividing resource by resource would be slightly consfusing (maybe).

Kobzol (approved):

PR #649, opened 2023-12-08
New "resubmit" mechanism

spirali:
The preparation for removing resubmit.

It allows "resubmit" via:
```commandline
hq submit --array=`hq job task-ids --filter=filter` ... 
```

(This PR does not removes resubmit yet)

### DIFF
--- a/crates/hyperqueue/src/client/commands/submit/command.rs
+++ b/crates/hyperqueue/src/client/commands/submit/command.rs
@@ -681,7 +681,7 @@ pub(crate) async fn send_submit_request(
         )
         .await?;
     } else if progress {
-        wait_for_jobs_with_progress(session, vec![info]).await?;
+        wait_for_jobs_with_progress(session, &[info]).await?;
     }
     Ok(())
 }
--- a/crates/hyperqueue/src/client/commands/submit/command.rs
+++ b/crates/hyperqueue/src/client/commands/submit/command.rs
@@ -714,7 +714,7 @@ fn get_ids_and_entries(opts: &JobSubmitOpts) -> anyhow::Result<(IntArray, Option
                     })
                     .collect(),
             );
-            IntArray::from_ids(id_set.iter().copied())
+            IntArray::from_sorted_ids(id_set.into_iter())
         } else {
             array.clone()
         }
--- a/crates/hyperqueue/src/client/commands/submit/command.rs
+++ b/crates/hyperqueue/src/client/commands/submit/command.rs
@@ -1,3 +1,4 @@
+use std::collections::BTreeSet;
 use std::io::{BufRead, Read};
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
@@ -246,14 +247,14 @@ pub struct SubmitJobConfOpts {
     // Parameters for creating array jobs
     /// Create a task array where a task will be created for each line of the given file.
     /// The corresponding line will be passed to the task in environment variable `HQ_ENTRY`.
-    #[arg(long, conflicts_with("array"), value_hint = clap::ValueHint::FilePath)]
+    #[arg(long, value_hint = clap::ValueHint::FilePath)]
     each_line: Option<PathBuf>,
 
     /// Create a task array where a task will be created for each item of a JSON array stored in
     /// the given file.
     /// The corresponding item from the array will be passed as a JSON string to the task in
     /// environment variable `HQ_ENTRY`.
-    #[arg(long, conflicts_with("array"), conflicts_with("each_line"), value_hint = clap::ValueHint::FilePath)]
+    #[arg(long, conflicts_with("each_line"), value_hint = clap::ValueHint::FilePath)]
     from_json: Option<PathBuf>,
 
     /// Create a task array where a task will be created for each number in the specified number range.
@@ -687,7 +688,7 @@ pub(crate) async fn send_submit_request(
 }
 
 fn get_ids_and_entries(opts: &JobSubmitOpts) -> anyhow::Result<(IntArray, Option<Vec<BString>>)> {
-    let entries = if let Some(ref filename) = opts.conf.each_line {
+    let mut entries = if let Some(ref filename) = opts.conf.each_line {
         Some(read_lines(filename)?)
     } else if let Some(ref filename) = opts.conf.from_json {
         Some(make_entries_from_json(filename)?)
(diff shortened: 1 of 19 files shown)


crates/hyperqueue/src/client/output/quiet.rs
```diff
@@ -130,6 +131,8 @@ impl Output for Quiet {
     ) {
     }
 
+    fn print_task_ids(&self, _job_task_ids: Vec<(JobId, IntArray)>) {}
```
Kobzol, line 134: I think that here we could write one ID per line? This mode should really be renamed from quiet to something like `bash-processable` :)

crates/hyperqueue/src/client/commands/wait.rs
```diff
@@ -72,18 +73,23 @@ pub async fn wait_for_jobs(
...
         log::warn!("There are no jobs to wait for");
     } else {
-        let total_tasks: JobTaskCount = jobs.iter().map(|info| info.n_tasks).sum();
-        let mut remaining_job_ids: Set<JobId> = jobs.into_iter().map(|info| info.id).collect();
+        let total_tasks: JobTaskCount = jobs
+            .iter()
+            .filter(|info| !is_terminated(info))
+            .map(|info| info.n_tasks)
+            .sum();
+        let mut remaining_job_ids: BTreeSet<JobId> = jobs
```
Kobzol, line 86: Any particular reason why `Set` was switched to `BTreeSet` and why the `Vec` was changed to a slice?

crates/hyperqueue/src/common/arraydef.rs
```diff
@@ -38,19 +38,22 @@ impl IntArray {
         IntArray { ranges }
     }
 
-    pub fn from_ids(ids: Vec<u32>) -> IntArray {
+    // ids has to be sorted!
```
Kobzol, line 41: Let's add an assert for this :)

crates/hyperqueue/src/common/arraydef.rs
```diff
@@ -38,19 +38,22 @@ impl IntArray {
...
+        let mut last_id = None;
         for id in ids {
-            if let Some(pos) = ranges.iter().position(|x| id == (x.start + x.count)) {
-                ranges[pos].count += 1;
+            if last_id.map(|last_id| last_id + 1 == id).unwrap_or(false) {
+                ranges.last_mut().unwrap().count += 1;
             } else {
                 ranges.push(IntRange::new(id, 1, 1));
             }
+            last_id = Some(id)
```
Kobzol, line 51: Missing semicolon.

crates/hyperqueue/src/common/arraydef.rs
```diff
@@ -104,7 +107,11 @@ impl fmt::Display for IntArray {
                 ));
             }
         }
-        write!(f, "{}", &str[0..str.len() - 2])
+        if str.len() >= 2 {
```
Kobzol, line 110: We could use `enumerate` and only add `, ` if we're not in the last iteration, that's way less hacky.

crates/hyperqueue/src/common/cli.rs
```diff
@@ -320,6 +320,8 @@ pub enum JobCommand {
     Wait(JobWaitOpts),
     /// Interactively observe the execution of a job
     Progress(JobProgressOpts),
+    /// Print tasks Ids for given job
```
Kobzol, line 323: `tasks` -> `task`

crates/hyperqueue/src/server/client/submit.rs
```diff
@@ -194,7 +194,7 @@ pub async fn handle_resubmit(
 
                         ids.sort_unstable();
                         JobDescription::Array {
-                            ids: IntArray::from_ids(ids),
+                            ids: IntArray::from_ids(ids.iter().copied()),
```
Kobzol, line 197: `into_iter()`

tests/test_job.py
```diff
@@ -1476,3 +1476,22 @@ def check_child_process_exited(hq_env: HqEnv, stop_fn: Callable[[subprocess.Pope
     parent, child = pids
     wait_for_pid_exit(parent)
     wait_for_pid_exit(child)
+
+def test_job_task_ids(hq_env: HqEnv):
+    hq_env.start_server()
+    hq_env.command(["submit", "--array=2,7,9,20-30", "--", "python", "-c", "import os; assert os.environ['HQ_TASK_ID'] not in ['25', '26', '27', '28']"])
+    hq_env.start_workers(1, cpus=1)
+    wait_for_job_state(hq_env, 1, "FAILED")
```
Kobzol, line 1484: We should wait until all tasks have finished :thinking: Otherwise there could be a race condition in the test.

docs/jobs/arrays.md
```diff
@@ -100,3 +100,19 @@ If `items.json` contained this content:
 ```
 then HyperQueue would create two tasks, one with `HQ_ENTRY` set to `{"batch_size": 4, "learning_rate": 0.01}`
 and the other with `HQ_ENTRY` set to `{"batch_size": 8, "learning_rate": 0.001}`.
+
+### Combining with `--each-line`/`--from-json` with `--array`
+
+Option `--each-line` or `--from-json` can be combined with option `--array`.
+In such case, only a subset of lines/json will be submited.
```
Kobzol, line 107: `submited` -> `submitted`

docs/jobs/failure.md
```diff
@@ -1,29 +1,42 @@
...
-parameters. You can achieve this using **resubmit**:
+## Resubmitting array jobs
+When a job fails or is canceled, you can submit it again. 
+However, in case of [task arrays](arrays.md), different tasks may end in different states, and often we want to 
+recompute only tasks with a specific status (e.g. failed tasks).
 
-```bash
-$ hq job resubmit <job-id>
+By following combination of commands you may recompute only failed tasks. Let us assume that we want to recompute
+all failed jobs in job 5:
```
Kobzol, line 10: `jobs` -> `tasks`

crates/hyperqueue/src/client/commands/submit/command.rs
```diff
@@ -687,18 +688,38 @@ pub(crate) async fn send_submit_request(
...
+                    .filter_map(|(id, value)| {
+                        if id_set.contains(&(id as u32)) {
+                            Some(value)
+                        } else {
+                            None
+                        }
+                    })
+                    .collect(),
+            );
+            IntArray::from_ids(id_set.iter().copied())
```
Kobzol, line 717: `into_iter()`

Kobzol (commented):
This is a pretty cool idea, I like it!

tests/test_job.py
spirali: It is not a race. FAILED state (of a job) is when there are no running & waiting tasks.

Kobzol: Oh, you're right, sorry. Nevermind :)

crates/hyperqueue/src/common/arraydef.rs
spirali: The method is quite ugly but I have tried to do a minimal touches originally. I have rewritten it in the last commit.

crates/hyperqueue/src/client/commands/wait.rs
spirali: Because we need to random removes & sorted iterator for creating IntArray.

crates/hyperqueue/src/client/output/quiet.rs
spirali: I have changed it. if there are 2 or more jobs, it prints also job ids.

Kobzol (approved):

PR #679, opened 2024-02-23
Dependencies updated

spirali:


.github/workflows/build.yml
```diff
@@ -46,7 +46,7 @@ jobs:
         uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: 1.72.0
+          toolchain: 1.76.0
```
Kobzol, line 49: We use `1.72.0` as our minimum (https://github.com/It4innovations/hyperqueue/blob/main/Cargo.toml#L14), so if we change it in CI, we should also update it in `Cargo.toml`.

Kobzol (changes requested):

.github/workflows/build.yml
spirali: Fixed. Btw: I am updating environment because one of our dependencies needs at least 1.74.

Kobzol (approved):

PR #680, opened 2024-02-23
HQ_HOST_FILE and HQ_NUM_NODES introduced

spirali:
Some modifications to run multinode tasks on Karolina more easily.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,17 @@
+# Dev
+
+## New features
+
+* `HQ_NUM_NODES` for multi-node tasks introduced. It contains the number of nodes assigned to task.
+  You do not need to manually count lines in `HQ_NODE_FILE` anymore.
+
+## Changes
+
+* Node file generated for multi-node tasks now contains only short hostnames
+  (e.g. if hostname is "cn690.karolina.it4i.cz", only "cn690" is written into node list)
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,17 @@
...
+
+## New features
+
+* `HQ_NUM_NODES` for multi-node tasks introduced. It contains the number of nodes assigned to task.
+  You do not need to manually count lines in `HQ_NODE_FILE` anymore.
+
+## Changes
+
+* Node file generated for multi-node tasks now contains only the first part of hostname
+  if it contains character "." (e.g. if hostname is "cn690.karolina.it4i.cz", only "cn690" is written into node list)
```
Kobzol, line 11: I wonder how universal this is. Is it a hotfix to make this work on Karolina? :)

spirali: It not only problem of Karolina but it problem on all IT4I systems  (e.g. Quake on OpenMPI on Barbora has the same problem) and on all systems that uses dots in hostnames. 

As it seems, at least SLURM and OpenMPI removes everything after first dot. 

As I understand the problem: having "." in hostnames is a bad practice as it mess with DNS. This cutting of dots is common (??) practice how to solve the problem.

Kobzol: I see. Ok, then please just include this reasoning as a comment in the code, where the split is implemented.

Kobzol (approved):

PR #706, opened 2024-05-17
Resuming server state from journal + live event streaming

spirali:
* Journal can be loaded and server state can be resumed
* Event format is changed
  - New events/data was added for resuming server state needs
  - old format used Tako Ids, new format is Tako independant 
  - events are serialized via bincode (not msgpack)
* Server no longer remembers past events in memory.
  If history is needed, it is loaded from journal file
* Some copying of events removed
* Events can be dynamically streamed into client, message pulling is removed
* This change completely breaks Dashboard (mainly because of events format (TakoIds) and event streaming)
* Default overview_interval is set to 5s.


crates/hyperqueue/src/server/event/log/write.rs
```diff
@@ -1,66 +1,44 @@
...
-        file.flush().await?;
-
-        // Keep buffer capacity larger than max size to avoid reallocation if we overflow
-        // the buffer.
-        let buffer = Vec::with_capacity(BUF_MAX_SIZE * 2);
-        Ok(Self { file, buffer })
+    pub fn create_or_append(path: &Path) -> anyhow::Result<Self> {
+        let mut file = if !path.exists() {
+            let mut file = BufWriter::new(File::create(path)?);
+            if file.stream_position()? == 0 {
```
Kobzol: Is it possible for this to not be `0`?

crates/hyperqueue/src/server/event/log/stream.rs
```diff
@@ -54,17 +54,17 @@ async fn streaming_process(
     loop {
         tokio::select! {
             _ = flush_fut.tick() => {
-                writer.flush().await?;
+                writer.flush()?;
```
Kobzol, line 66: Why is this no longer async? That could block the main HQ thread. Especially if we're writing to LUSTRE...

crates/hyperqueue/src/server/event/log/read.rs
```diff
@@ -28,21 +33,38 @@ impl EventLogReader {
         if hq_version != HQ_VERSION {
             bail!("Version of journal {hq_version} does not match with {HQ_VERSION}");
         }
-        Ok(Self { source: file })
+        Ok(Self {
+            position: file.stream_position().unwrap(),
```
Kobzol, line 37: `unwrap()` -> `?`

crates/hyperqueue/src/server/event/log/read.rs
```diff
@@ -4,19 +4,24 @@ use crate::HQ_VERSION;
...
-use std::io::BufReader;
 use std::io::Read;
+use std::io::{BufReader, Seek};
 use std::ops::Deref;
 use std::path::Path;
 
 /// Reads events from a file in a streaming fashion.
 pub struct EventLogReader {
     source: BufReader<File>,
+    position: u64,
```
Kobzol, line 15: Could you please document how the restoration and truncation works?

crates/hyperqueue/src/client/commands/event/mod.rs
```diff
@@ -30,12 +38,37 @@ struct ExportOpts {
...
+        .send(FromClientMessage::StreamEvents)
+        .await?;
+    let stdout = std::io::stdout();
+    let stdout = stdout.lock();
+    let mut stdout = BufWriter::new(stdout);
+    while let Some(event) = connection.connection().receive().await {
+        let event = event?;
+        if let ToClientMessage::Event(e) = event {
+            writeln!(stdout, "{}", format_event(e))?;
+            stdout.flush()?;
```
Kobzol, line 62: Maybe we could flush only every once in a while? Flushing after every event kind of defeats the purpose of the `BufWriter`.

crates/hyperqueue/src/server/event/log/stream.rs
```diff
@@ -68,6 +80,13 @@ async fn streaming_process(
                             break
                         }
                     }
+                    Some(EventStreamMessage::RegisterListener(tx)) => {
```
Kobzol, line 80: How exactly does this work? It reads the events.. from the file system log file? And also blocks this `async fn`?

crates/hyperqueue/src/common/utils/time.rs
```diff
@@ -67,7 +67,7 @@ pub mod mock_time {
     use std::time::Instant;
 
     thread_local! {
-        static MOCK_TIME: RefCell<Option<Instant>> = RefCell::new(None);
+        static MOCK_TIME: RefCell<Option<Instant>> = const { RefCell::new(None) };
```
Kobzol, line 70: Isn't `inline const` only stabilized in Rust `1.79`?

crates/tako/src/internal/common/trace.rs
```diff
@@ -41,39 +35,39 @@ macro_rules! trace_time {
...
-            w,
-            "{}",
-            SystemTime::now()
-                .duration_since(SystemTime::UNIX_EPOCH)
-                .unwrap()
-                .as_micros()
-        )
-    }
-}
+// struct FileGuard(Arc<Mutex<BufWriter<std::fs::File>>>);
```
Kobzol, line 38: What's up with this?

docs/deployment/server.md
```diff
@@ -57,7 +59,39 @@ The server will quit when your SSH session ends, because it will receive a SIGHU
...
+
+```bash
+$ hq server start --journal /path/to/journal
+```
+
+!!! warning
+
+    This functionality resumes the state of jobs and auto allocation queues,
+    not worker connections.
+    In the current version, new workers has to be connected to the server
```
Kobzol, line 84: `has` -> `have`

Kobzol (commented):
Left a bunch of comments. Huge PR :)

Cargo.toml
```diff
@@ -55,3 +55,6 @@ inherits = "release"
 lto = true
 codegen-units = 1
 debug = "line-tables-only"
+
+[profile.dev]
```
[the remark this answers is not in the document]
Kobzol: Yeah, please revert this. This is not great:
image

crates/hyperqueue/src/common/utils/time.rs
Kobzol: It's fine to upgrade, but it hasn't been released yet, that's why I was surprised :laughing: I realized now that this is actually some built-in functionality of the `thread_local!` macro (https://doc.rust-lang.org/std/macro.thread_local.html).

crates/tako/src/internal/common/trace.rs
Kobzol: Ah, it is unused. Let's delete this.

crates/hyperqueue/src/server/event/log/write.rs
```diff
@@ -13,8 +13,12 @@ pub struct EventLogWriter {
...
 impl EventLogWriter {
-    pub async fn create(path: &Path) -> anyhow::Result<Self> {
-        let mut file = File::create(path).await?;
+    pub async fn create_or_append(path: &Path) -> anyhow::Result<Self> {
+        let mut file = OpenOptions::new()
+            .create(true)
+            .append(true)
+            .open(path)
+            .await?;
         let header = rmp_serde::encode::to_vec(&canonical_header())?;
```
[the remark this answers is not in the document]
spirali: This code was removed in later commits. In the final version header is written only once.

crates/hyperqueue/src/server/restore.rs
```diff
@@ -0,0 +1,75 @@
...
+use tako::ItemId;
+
+struct RestorerJob {
+    job_info: JobInfo,
+}
+
+impl RestorerJob {
+    pub fn restore(self, job_id: JobId, state: &mut State) {
+        let tako_base_id = state.new_task_id(self.job_info.task_ids.len() as JobTaskCount);
+        let job = Job::new(
```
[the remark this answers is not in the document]
spirali: What should be no-op?

crates/hyperqueue/src/server/event/log/stream.rs
spirali: Yes it blocks the thread, but only the thread that is responsible for writing events into the file. We need this to avoid modification of file while we are reading from it.

tests/test_restore.py
```diff
@@ -17,15 +21,103 @@ def test_restore_fully_finished(hq_env: HqEnv, tmp_path):
...
+    table.check_columns_value(["ID", "State", "Name"], 1, ["2", "WAITING", "sleep"])
+
+    hq_env.start_worker()
+    wait_for_job_state(hq_env, [1, 2], "FINISHED")
+
+
+def test_restore_partially_finished_task(hq_env: HqEnv, tmp_path):
+    journal_path = os.path.join(tmp_path, "my.journal")
+    hq_env.start_server(args=["--journal", journal_path])
+    hq_env.command(["submit", "--array=0-4", "--", "python", "-c",
```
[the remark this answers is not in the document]
spirali: Where I can find it?

crates/hyperqueue/src/server/event/log/write.rs
```diff
@@ -1,66 +1,44 @@
...
-        let header = bincode_config().serialize(&canonical_header())?;
-        file.write_all(&header).await?;
-        file.flush().await?;
-
-        // Keep buffer capacity larger than max size to avoid reallocation if we overflow
-        // the buffer.
-        let buffer = Vec::with_capacity(BUF_MAX_SIZE * 2);
-        Ok(Self { file, buffer })
+    pub fn create_or_append(path: &Path) -> anyhow::Result<Self> {
+        let mut file = if !path.exists() {
```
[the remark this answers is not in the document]
spirali: It was just a temporary version. Not in the final one.
On the other hand, when some other process will modify the journal, it will probably goes horribly wrong as we read the file, potentially truncating it, and then appending to it. If someone modify the file in between it it is quite wild what may happen with the consistency of the file

crates/hyperqueue/src/server/event/log/write.rs
spirali: Not in the final version.

crates/hyperqueue/src/server/event/log/stream.rs
spirali: No, all operations over the file that runs in its own thread. And it actually do only one thing at the time. It is async only because we need to communicate via async queues form main thread.

crates/hyperqueue/src/client/commands/event/mod.rs
spirali: BufWriter still have purpose as an event may be written in several writes (especially large event "submit").
However, I consider this mostly as command for testing. As we do not have non-testing users for this, it is hard to find the right flush policy.

crates/hyperqueue/src/server/restore.rs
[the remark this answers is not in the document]
Kobzol: Outdated comment. Too many commits at once :)

tests/test_restore.py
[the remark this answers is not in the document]
Kobzol: `tests/utils/cmd.py`

crates/hyperqueue/src/server/event/log/stream.rs
Kobzol: I understand it now. But please add some comments about this to the code, it's not obvious how this works and that it depends on blocking the writer thread.

crates/hyperqueue/src/client/commands/event/mod.rs
Kobzol: Ok, fair enough.

PR #728, opened 2024-07-18
Update build.yml

theely:
Builds Python binding on MacOS ARM

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -133,7 +133,10 @@ jobs:
           path: ${{ steps.archive.outputs.archive-name }}
   build-python-binding:
     needs: [ set-env ]
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.image }}
+    strategy:
+            matrix:
+                image: [ubuntu-latest, macos-latest]
     steps:
       - name: Checkout sources
         uses: actions/checkout@v3


Kobzol:
Hi, thanks for the PR! I'm still a bit uneasy about providing these packages when we officially don't support them :sweat_smile:, but I guess there's no harm in doing that when we explicitly state that it's a best effort approach.

It looks like `HQ_BUILD_VERSION` is not propagated into maturin, probably because it is set using `docker-options`, and Docker isn't used on macOS for the maturin build. Could you try adding `HQ_BUILD_VERSION: ${{ needs.set-env.outputs.version }}` [here](https://github.com/it4innovations/hyperqueue/blob/c1a0c9d723c54109540a4f3cfd5dd2ea1e272258/.github/workflows/build.yml#L138)?

Kobzol:
Right, so the macOS Python package was built, but most of the Python API tests (expectedly) fail with it. I'm really unsure whether we want to release packages are unsupported and also fail most tests, especially when interested users can just build the package locally.. @spirali What do you think?

theely:
Indeed this is less than ideal. Conceptually it makes sense to provide client libs (Python package) built for multiple architectures even if the server binary only runs on x64.
Unfortunately, I don't see a way to build a client on one architecture and test it against a server running on a different one.

Kobzol:
Wel, the client should *probably* work on macOS (not 100% sure though), so it is a valid use-case (although most HQ users just run the client on a supercomputer). But as you said, providing testing of a macOS client and e.g. a Linux/x64 server/worker is highly non-trivial and we do not have the necessary testing infrastructure prepared for that.

spirali:
I like the idea that we are distributing MacOs client, but without at least partial automatic testing it is quite thin ice and it will be broken somewhere in the future.

Does GitHub CI allow something like docker compose? So we can start linux & macos instance together? Or can be we start linux docker container inside a macos CI instance?

Kobzol:
Starting a (virtualized) Linux Docker container inside a macOS is probably possible, but I'm not sure if that works on GitHub Actions, I didn't find much information about it online.

PR #747, opened 2024-09-12
Worker streaming

### LINKED ISSUE #716: "Invalid entry type: 32"  in hq log when force removing a submission queue
Hi

I had HQ tasks that were failing but could not debug their error messages because the HQ log file was reporting this:
```
hq log hq-processing.log summary
```
was giving me this
```
Invalid entry type: 32
```
with any of the log commands 'summary', 'show' or 'cat'.

This is how I'm creating the allocation queue:
```
hq alloc add slurm --name processing \
    --workers-per-alloc 1 --max-worker-count $WORKERS --backlog $WORKERS \
    --idle-timeout 2m --time-limit 30m \
    -- -p debug -A $SBATCH_ACCOUNT \
    --cpus-per-task 128 --ntasks 1 --mem-per-cpu 1750 \
    -o "$SLURM_LOGS_DIR/hq-worker-%x.out" -e "$SLURM_LOGS_DIR/hq-worker-%x.err"
```
then I'm creating a tsv with all the tasks parameters and submitting an array of tasks like this:
```
hq submit --each-line $entries \
    --cpus 128 \
    --progress --log=$SLURM_LOGS_DIR/hq-processing.log \
    --max-fails=5 --crash-limit=5 \
    bash 10.processing
```

After the job is finished and the `hq submit` command stops waiting, I'm removing the queue with `hq alloc remove --force $qid`. I noticed that, without `--force`, the logs were showing correctly and I had the tasks log messages there. But to automatically remove the queue after `hq submit` finishes, I need to submit less tasks than `--max-fails`. So, I guess the problem here is force-removing the queue while it has a running job (even if a blocking `hq submit --progress` has finished and the remaining tasks won't run because the job has reached the `--max-fails` limit).

Is it possible that this is a bug and I could get the log messages (or at least the ones that have been captured, as it states in the documentation) even if I run `hq alloc remove --force`?

HyperQueue version: 0.19.0
Slurm version: 22.05.10

Thanks in advance,
Jaume

spirali:
Adds worker streaming & removes server streaming.

It also fixes #716


crates/hyperqueue/src/client/commands/reader.rs
```diff
@@ -45,16 +53,22 @@ pub struct CatOpts {
...
     pub task: Option<IntArray>,
 }
 
 #[derive(Parser)]
-pub enum LogCommand {
+pub enum StreamCommand {
     /// Prints summary of log file
     Summary(SummaryOpts),
 
+    /// Prints content of log ordered by time
```
Kobzol, line 69: Is the comment correct?

crates/hyperqueue/src/common/cli.rs
```diff
@@ -199,7 +199,7 @@ pub enum SubCommand {
     /// Commands for controlling HyperQueue workers
     Worker(WorkerOpts),
     /// Operations with log
-    Log(LogOpts),
```
Kobzol, line 202: Should this be called `Stream`? `Read` seems quite general, we might want to introduce more things that could be read by the server in the future.

crates/hyperqueue/src/stream/reader/streamdir.rs
```diff
@@ -0,0 +1,464 @@
...
+    }
+
+    pub fn instance(&self, instance_id: InstanceId) -> Option<&InstanceInfo> {
+        self.instances
+            .iter()
+            .find(|info| info.instance_id == instance_id)
+    }
+}
+
+pub struct StreamDir {
```
Kobzol, line 73: Could you please add some comments either to this struct or to the whole file, to provide some high-level explanation of what is this good for and (mainly) how does it work?

crates/hyperqueue/src/stream/reader/streamdir.rs
```diff
@@ -0,0 +1,464 @@
...
+        let mut paths = Vec::new();
+        let mut server_uids: Set<String> = Set::new();
+        let mut found = false;
+        for entry in std::fs::read_dir(path)? {
+            let entry = entry?;
+            let path = entry.path();
+            if path
+                .extension()
+                .and_then(|e| e.to_str())
+                .map(|s| s == "hqs")
```
Kobzol, line 105: Could we please put `"hqs"` into some global constant? It is probably used on multiple places.

crates/hyperqueue/src/stream/reader/streamdir.rs
```diff
@@ -0,0 +1,464 @@
...
+            return Err(HqError::GenericError("No log files found".to_string()));
+        }
+
+        if server_uids.len() > 1 {
+            use std::fmt::Write;
+            let mut msg = "Found streams from multiple server instances:\n".to_string();
+            for server_uid in server_uids {
+                writeln!(msg, "{server_uid}").unwrap();
+            }
+            msg.push_str("Use --server-uid=... to specify server uid");
```
Kobzol, line 143: This might be a bit annoying for users. I assume that in most cases, they will want "read the files for the latest server UID". As a future improvement, we could provide this functionality somehow.

crates/hyperqueue/src/stream/reader/streamdir.rs
```diff
@@ -0,0 +1,464 @@
...
+                }
+                _ => Err(error.into()),
+            },
+        }
+    }
+
+    fn create_index(
+        paths: &[PathBuf],
+    ) -> crate::Result<BTreeMap<JobId, BTreeMap<JobTaskId, TaskInfo>>> {
+        let mut index: BTreeMap<JobId, BTreeMap<JobTaskId, TaskInfo>> = BTreeMap::new();
```
Kobzol, line 171: Type alias `StreamIndex = BTreeMap<JobId, BTreeMap<JobTaskId, TaskInfo>>`?

crates/hyperqueue/src/transfer/stream.rs
```diff
@@ -26,25 +25,27 @@ pub struct DataMsg {
...
+    pub instance: InstanceId,
+    pub channel: ChannelId,
+    pub size: u64,
 }
 
-#[derive(Serialize, Deserialize, Debug)]
-pub enum ToStreamerMessage {
-    Error(String),
-    EndResponse(EndTaskStreamResponseMsg),
+pub enum StreamerMessage {
```
Kobzol, line 45: This should probably be in `src/worker/streamer.rs`? It does not implement serde traits and it's currently only used in that file.

crates/hyperqueue/src/worker/bootstrap.rs
```diff
@@ -97,28 +97,29 @@ pub async fn initialize_worker(
             server_dir.access_filename()
         )
     })?;
-    let server_address = format!("{}:{}", record.worker.host, record.worker.port);
-    log::info!("Connecting to: {}", server_address);
 
     std::fs::create_dir_all(&configuration.work_dir)?;
     std::fs::create_dir_all(&configuration.log_dir)?;
```
Kobzol, line 102: Found this by accident, but it seems to me like `log_dir` isn't actually used by anything in HQ?

crates/hyperqueue/src/worker/start/program.rs
```diff
@@ -210,18 +210,18 @@ async fn resend_stdio(
...
-    stream: Rc<StreamSender>,
+    stream: StreamSender,
 ) -> tako::Result<()> {
     if let Some(mut stdio) = stdio {
-        log::debug!("Starting stream {}/{}/{}", job_id, job_task_id, channel);
+        log::debug!("Resending stream {}/{}/{}", job_id, job_task_id, channel);
         loop {
             let mut buffer = vec![0; STDIO_BUFFER_SIZE];
             let size = stdio.read(&mut buffer[..]).await?;
+            buffer.truncate(size);
```
Kobzol, line 220: Buffer with size zero is used to signal the end of the stream? It would be nice to leave an explicit comment about this. Or even an explicit enum variant in `StreamerMessage` :)

crates/hyperqueue/src/worker/streamer.rs
```diff
@@ -1,314 +1,205 @@
...
+    let mut buffer: Vec<u8> = Vec::new();
+    {
+        let streamer = streamer_ref.get();
+        let header = StreamFileHeader {
+            server_uid: Cow::Borrowed(&streamer.server_uid),
+            worker_id: streamer.worker_id,
+        };
+        bincode_config().serialize_into(&mut buffer, &header)?;
+    };
+    file.write_all(&buffer).await?;
```
Kobzol, line 182: Why is this not written to the `file` directly?

docs/jobs/streaming.md
```diff
@@ -8,104 +8,101 @@ $ hq submit --array=1-10000 my-computation.sh
...
-## Redirecting output to the log
+## Redirecting output to the stream
+
 You can redirect the output of `stdout` and `stderr` to a log file and thus enable output streaming by passing a path
-to a filename where the log will be stored with the `--log` option:
+to a filename where the log will be stored with the `--stream` option:
 
 ```
-$ hq submit --log=<log-path> --array=1-10000 ...
+$ hq submit --log=<stream-path> --array=1-10_000 ...
```
Kobzol, line 28: `--log` -> `--stream`

docs/jobs/streaming.md
```diff
@@ -8,104 +8,101 @@ $ hq submit --array=1-10000 my-computation.sh
...
+### Stream summary
 
-## Inspecting the log file
-HyperQueue lets you inspect the data stored inside the log file using various subcommands. All `log` subcommands have the
-following structure:
+You can display a summary of a log file using the `summary` subcommand:
 
 ```bash
-$ hq log <log-file-path> <subcommand> <subcommand-args>
+$ hq log <stream-path> summary
```
Kobzol, line 88: `hq read`? Same below.

tests/test_jobfile.py
```diff
@@ -22,10 +23,11 @@ def test_job_file_submit_minimal(hq_env: HqEnv, tmp_path):
 def test_job_file_submit_maximal(hq_env: HqEnv, tmp_path):
     hq_env.start_server()
     hq_env.start_workers(3, cpus=4, args=["--resource", "gpus=[0,1]"])
+    os.mkdir("output")
     tmp_path.joinpath("job.toml").write_text(
         """
 name = "test-job"
-stream_log = "output.log"
+stream = "output"
```
Kobzol, line 30: Documentation in `docs/jobs/jobfile.md` should be updated (`stream_log` ->` stream`) - it also contains a `--log` reference.

crates/hyperqueue/src/stream/reader/streamdir.rs
spirali: It is not clear what is the last server uid. Moreover servers may run in parallel and write to the same folder.

crates/hyperqueue/src/worker/streamer.rs
spirali: Because serde does not support direct writing to async files

crates/hyperqueue/src/common/cli.rs
spirali: I did not want name it just "stream" as it is a verb for creating a stream. But "read-stream" seems quite long for me so I have left just "read".

Kobzol: Our existing commands use `<noun> <verb>`:
- `server start` (noun verb)
- `job submit` (noun verb)
There are some cases that are something between a noun and a verb (summary, info, list), but it's mostly consistent.

So to keep it consistent, this should be `hq stream read`, IMO.

CHANGELOG.md
```diff
@@ -4,6 +4,15 @@
 
 * It is now possible to dynamically submit new tasks into an existing job (we call this concept "Open jobs").
   See [Open jobs documentation](https://it4innovations.github.io/hyperqueue/stable/jobs/openjobs/)
+* Worker streaming. Instead, streaming to server, worker can now write combined io stream files directly.
+  It creates more files than original solution (that was exactly 1 file), but the number of files stays
+  small and independent on the number of executed tasks. The new architecture also parallel write and
+  storing more job streams into one stream handle.
```
Kobzol, line 10:
```suggestion
* Worker streaming. Before, you could stream task stderr/stdout to the server over the network using the `--log` parameter of `hq submit`. This approach had various issues and was not scalable. Therefore, we have replaced this functionality with worker streaming, where the streaming of task output to a set of files on disk is performed by workers instead.
  This new streaming approach creates more files than original solution (where it was always one file per job), but the number of files stays
  small and independent on the number of executed tasks. The new architecture also allows parallel I/O writing and storing of multiple job streams in one stream handle. You can use worker streaming using the `--stream` parameter of `hq submit`. Check out the documentation for more information.
```

crates/hyperqueue/src/common/cli.rs
spirali: I do not like it as we are not planning any other `hq stream <something>`. If you think that noun is better than I propose `hq stream-reader` .

Kobzol: That sounds surprising, as we have several operations with the log. I checked the CLI help now and I think that we have unnecessarily moved one layer down.

Now there is `hq read cat` , `hq read export`, `hq read jobs`, that is really weird. I understand that you don't want `hq stream read cat`, but if we just get rid of `read`, I think that it will solve itself. So we should have just `hq stream export`, `hq stream cat` etc.

spirali: Sorry for briefness. I wanted to propose `hq stream-reader cat ...`.  I just want to avoid name "stream" as it seems to me very misleading, I want to signal that is just a reading end of the interface, not "streaming" one.

spirali: Renamed to `hq outout-log`

Kobzol (approved):
Great job! :)

PR #765, opened 2024-10-07
Removed some very old and obsolete code from Tako

spirali:


Kobzol (approved):
Removing code is always good :)

PR #766, opened 2024-10-07
Deps updated

spirali:


Kobzol:
I guess we could update our minimum version to `1.80.0`.

Kobzol (approved):

PR #774, opened 2024-10-18
Add `Serialized` wrapper to make it harder to accidentally deserialize the wrong type

Kobzol:
While restoring the dashboard, I realized that the comment about the type of the `Submit::serialized_desc` was wrong - it was not storing `JobDescription`, but `SubmitRequest`. There was also one place in the code that was deserializing this incorrect type. I created a type wrapper to make sure that this doesn't happen again.

We should use the same thing for `SerializedTaskContext`, however it currently uses different bincode options (it uses fixint encoding and doesn't allow trailing bytes), so I did not use it there yet.

### DIFF
--- a/crates/hyperqueue/src/client/commands/journal/output.rs
+++ b/crates/hyperqueue/src/client/commands/journal/output.rs
@@ -1,8 +1,7 @@
 use crate::client::output::json::format_datetime;
 use crate::server::event::payload::EventPayload;
-use crate::server::event::{bincode_config, Event};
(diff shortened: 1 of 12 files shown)


spirali:
Nice idea.

Both serialization of SubmitRequest and SerializedTaskContext should not allow trailing spaces. I think that it is ok to synchronize both on the same serializer. My first guess would be variable encoding but I did do any benchmarking so also fixint is probably ok.

spirali (approved):

PR #778, opened 2024-10-24
Task task id refactoring

spirali:
This PR makes Tako's TaskId as u64, this allows to encode JobId and JobTaskId directoly into TakoId. It allows to remove translation lookups between TakoId and JobTaskId/TakoId.  It also allows to have keys in the map of tasks in Job directly JobTaskIds (and not TakoTaskId), that allows to remove some creating ad-hoc sets of JobTaskIds and directly do a lookup into tasks.

### DIFF
--- a/crates/hyperqueue/src/server/job.rs
+++ b/crates/hyperqueue/src/server/job.rs
@@ -7,7 +7,7 @@ use crate::transfer::messages::{
     TaskStatusSelector,
 };
 use crate::worker::start::RunningTaskContext;
-use crate::{JobId, JobTaskCount, JobTaskId, Map, TakoTaskId, WorkerId};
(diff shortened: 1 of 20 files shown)


crates/hyperqueue/src/server/job.rs
```diff
@@ -160,20 +161,40 @@ impl Job {
 
     pub fn make_job_detail(&self, task_selector: Option<&TaskSelector>) -> JobDetail {
         let (mut tasks, tasks_not_found) = if let Some(selector) = task_selector {
-            match &selector.id_selector {
-                TaskIdSelector::All => (
+            match (&selector.id_selector, &selector.status_selector) {
```
Kobzol, line 164: Do we have a test for this?

Kobzol (approved):
Nice cleanup! Left one nit.

PR #781, opened 2024-10-30
Fix alloc resume message

StHagel:
Currently resuming a paused alloc prints the same message as pausing it. This is probably a remnant of some copy-and-pasting.

This commit fixes the message to properly show the word "resumed" instead of "paused".


Kobzol:
Indeed :) Thank you very much!

PR #786, opened 2024-11-16
New journal operations

spirali:
Adds:
`hq journal flush`
`hq journal prune`
`hq start server --journal-flush-period=...`


crates/hyperqueue/src/server/event/journal/stream.rs
```diff
@@ -83,11 +93,28 @@ async fn streaming_process(
...
-                        let mut reader = JournalReader::open(log_path)?;
+                        let mut reader = JournalReader::open(journal_path)?;
                         for event in &mut reader {
                             tx.send(event?).unwrap()
                         }
-                    }
+                    },
+                    Some(EventStreamMessage::PruneJournal { live_jobs, live_workers, callback })  => {
+                        writer.flush()?;
+                        let mut tmp_path: OsString = journal_path.into();
```
Kobzol, line 104: It would be safer to create the journal file in a temporary directory, to avoid potential race conditions. Although we don't know where /tmp is located, so maybe try to atomically create a unique file in the directory of the journal?

crates/hyperqueue/src/server/event/journal/stream.rs
```diff
@@ -83,11 +93,28 @@ async fn streaming_process(
...
+                        tmp_path.push(".tmp");
+                        let tmp_path: PathBuf = tmp_path.into();
+                        {
+                            let mut reader = JournalReader::open(journal_path)?;
+                            let mut writer = JournalWriter::create_or_append(&tmp_path, None)?;
```
Kobzol, line 109: Shouldn't it be an error if the `tmp_path` file already existed before?

crates/hyperqueue/src/server/event/journal/stream.rs
```diff
@@ -83,11 +93,28 @@ async fn streaming_process(
...
+                            if let Err(e) = prune_journal(&mut reader, &mut writer, &live_jobs, &live_workers) {
+                                remove_file(&tmp_path)?;
+                                return Err(e.into())
+                            }
+                        }
+                        rename(&tmp_path, &journal_path)?;
+                        writer = JournalWriter::create_or_append(journal_path, None)?;
```
Kobzol, line 116: Shouldn't we drop the writer before doing the rename? Otherwise the `writer` will be closed while the file it deals with was already overwritten. It's probably fine, but one never knows with file descriptors...

crates/hyperqueue/src/client/commands/journal/mod.rs
```diff
@@ -30,6 +30,9 @@ enum JournalCommand {
 
     /// Connect to a server and remove completed tasks and non-active workers from journal
     Prune,
+
+    /// Connect to a server and forces to flush a journal
+    Flush,
```
Kobzol, line 35: What's the motivation for this command? It seems weird to have a command that affects the internal implementation details of HQ and essentially flushes a buffer.. I hope it wasn't only added for the Python test xD

Kobzol (commented):
Left a few comments.

Kobzol:
I don't understand the https://github.com/It4innovations/hyperqueue/pull/786/commits/c38892a389dd65d3dda1e666d0dcd3c4f7c71607 commit. You don't need `Option`, you can simply `drop(writer)` and then overwrite it a few lines below.

spirali:
> I don't understand the [c38892a](https://github.com/It4innovations/hyperqueue/commit/c38892a389dd65d3dda1e666d0dcd3c4f7c71607) commit. You don't need `Option`, you can simply `drop(writer)` and then overwrite it a few lines below.

You are right, I have realized it during the night and do not have time fix it yet :)

Kobzol (approved):

PR #804, opened 2025-01-24
Deps updated

spirali:


Kobzol (approved):

Kobzol:
This adds 13 new packages :( But we don't need to block on that I guess. I can take a look later if we can deduplicate something.

PR #806, opened 2025-01-30
Fix hypershell desc

glentner:
Hey folks. 👋🏻 

This PR first and foremost was simply to fix the broken links in the comparison overview page on the website with regard to [hypershell](https://hypershell.org). I've also updated parts of the description. Overall I don't have any complaints, just offering my perspective as the author of HyperShell.

A couple things:
* I love the overview page and thought the matrix comparison was great. I think you did a good job on that.
* I think the comparison to GNU Parallel is appropriate and kept that in the description. The primary goal of the project was to build a better, more feature-rich utility beyond GNU Parallel that was more approachable (usable) for researchers / HPC users.
* I'm trying to keep the terminology strait between your project and how I typically talk about workflow topology in high-thoughput computer. In your documentation I believe "allocation" refers to a job allocation (i.e., Slurm job). Just so everyone understands, we do not bundle one task per allocation by default as your original wording states. The whole point of workflow tools like HQ, HS, and anything a user would sling together is to avoid individual jobs per task. When HyperShell creates bundles that doesn't strictly associate the task in that group of tasks it merely is a preferred size for the communication channels (i.e., database transactions). Regardless of task aggregation ("bundling"), they are executed asynchronous and will "return to the server" even if one of the other tasks originally in the bundle gets held up.

I appreciate inclusion in your list of other tools. I want to invite you to participate in an effort ongoing on this front, maybe not really related to this PR. I've been partnering with a bunch of HPC centers, national labs, universities, in the USA and am working on creating a online community "encyclopedia of workflow tools" for researchers to understand the full landscape of frameworks, tools, and platforms and how they relate to each other. In a nutshell, disambiguating foundational things like CWL (or DSL generally), parallel vs distributed, library vs tool, DAG or bag-of-tasks, etc.

If you're interested in being a part of that let me know.

### DIFF
--- a/docs/other-tools.md
+++ b/docs/other-tools.md
@@ -31,13 +31,7 @@ Below we also provide **opinionated**[^opinionated] descriptions of selected tas
 [GNU Parallel](https://www.gnu.org/software/parallel/) is a command-line utility for executing many tasks in parallel on a set of computational nodes. It does not offer many advanced task runtime features, but it does one thing well; it enables a parallelized and even distributed execution of a set of programs with a single command invocation. HyperQueue takes inspiration from this approach, as it offers a CLI that can be used to execute task graphs with many tasks and complex resource requirements with a single command.
 
 ## HyperShell
-[HyperShell](https://hypershell.readthedocs.io) is primarily designed for
-executing many homogeneous tasks using the command-line. It does introduce several useful features
-on top of [GNU Parallel](#gnu-parallel), such as automatic task re-execution when a task fails and storing the task state in a database, which enables users to observe the history of executed workflows. It also provides a simple autoscaling functionality that automatically submits allocations. However, tasks in HyperShell are strictly tied to allocations; by default, one task is submitted in a single
-allocation. It does provide the option to [bundle](https://hypershell.readthedocs.io/cli/cluster.html) several tasks together, but users
-have to specify the maximum bundle size explicitly, which makes load balancing inflexible.
-HyperShell does not support task dependencies; therefore, it cannot be used to execute general
-task graphs.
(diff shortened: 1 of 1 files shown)


Kobzol:
Hi! Thanks a lot for reaching out and sending this PR.

I have to say that every broad comparison of tools that deal with similar problems is necessarily being biased (since I'm the author of the comparison, and of one of the tools), not 100% informed (since I didn't use all of the tools for production use-cases) and simplistic/misleading (because tools are just too different to put into specific categories). Nevertheless, I had to compare some tools for my PhD thesis, so I had to somehow massage a bunch of tools together into a table :)

Sorry for misrepresenting the behavior of hypershell in our docs! My understanding was that the bundle size (which is 1 by default, if I understand it correctly) was used to determine how many tasks will be mapped to an allocation (e.g. a single SLURM job). But it seems like it's more of a performance thing, to determine how many tasks to bundle together when saving task statutes or when exchanging task information over the network?

In general, having a bundle size default of one isn't bad in any way, and it's nice that it can be overridden in HyperShell. I was trying to distinguish two general approaches, one is that users somehow split/slice the task graph into specific subgraphs manually or semimanually (by marking groups of tasks or specifying how large should an automatically created group be), and tools such as HyperQueue, which try to put all this burden to the scheduler, and perform this mapping fully dynamically at runtime, without forcing the user to think about it at all. It seems that based on your description, HyperShell also schedules tasks to different allocations without the user having to provide the mapping in any way, and the bundle size is just an optimization parameter (?).

In any case, I'm happy to merge your PR as is! I just need to fix CI, because some GitHub actions got deprecated and we'll need to upgrade them. I remember that there were some breaking changes in the new versions of these actions, so it might take me a few days, I'm a bit busy now.

glentner:
No worries, I understand. There are lots of other tools not even mentioned here (e.g., Launcher, TaskFarmer), NERSC has documentation for their users that also has a good overview page.

HyperShell is built of many concurrent threads that each operate as their own agent. The scheduler thread selects the next _N_ new tasks from the database (the "bundle" size) and serializes the bundle and offers them up to clients over a socket. Clients pull these bundles and unpack them locally for their concurrent executor threads to take and execute.  It is a very good thing to allow the default bundle size to be 1 because in 95% of cases I've run into with researchers there is no need for this optimization and we get more responsiveness in the system at typical small workflow sizes. If you have 10M identical single-core tasks though it helps to set `-b 128` (or however many cores per node). 

Feel free to reach out (glentner@purdue.edu) and I'll share about the efforts with this workflow encyclopedia.

Kobzol (approved):
Thank you!

PR #812, opened 2025-02-06
Update output streaming documentation

Kobzol:
It was a bit outdated in a few places.

This PR also creates the streaming directory if it does not exist, instead of returning an error, to match the documentation. I'm not sure if the creation was problematic for some reason, if so, we should add a comment explaining why that is.


spirali (approved):
The directory creation was omitted to detect typo in path more easily. But I think that creation of the directory is also ok

PR #822, opened 2025-02-19
Fix docs formatting

lahwaacz:
This should fix formatting in https://it4innovations.github.io/hyperqueue/stable/jobs/jobs/#time-management and below, which seems to be broken since 34a368b5efad0fd9c9e666539aa8aa8529ff322b


Kobzol (approved):
Thank you! :)

Issue #832, opened 2025-02-23
`hq alloc add slurm` does not fully understand Slurm time format


lahwaacz:
Hi, I'm sorry for reporting another issue 👼 

```
$ hq alloc add slurm --time-limit 1-00:00:00
error: invalid value '1-00:00:00' for '--time-limit <TIME_LIMIT>': Could not parse PBS walltime. Use either `HH:MM:SS` or humantime format (2hours): InvalidCharacter(1)

For more information, try '--help'.
```

From [sbatch manual](https://slurm.schedmd.com/sbatch.html#OPT_time):

> Acceptable time formats include "minutes", "minutes:seconds", "hours:minutes:seconds", "days-hours", "days-hours:minutes" and "days-hours:minutes:seconds".

Moreover, the error message says "PBS walltime" while this is a `slurm` subcommand.

Kobzol:
Hi, no problem, keep them coming :)

The `--time-limit` flag has custom syntax, which is printed in the help message. It's either a "humantime" format, or the PBS format. It does not accept Slurm syntax.

lahwaacz:
Can you clarify the supported formats in the `--help` output?

```
$ hq alloc add slurm --help
...
  -t, --time-limit <TIME_LIMIT>
          Time limit (walltime) of PBS/Slurm allocations
...
$ hq worker start --help
...
      --time-limit <TIME_LIMIT>
          Worker time limit. Worker exits after given time
...
```

Ideally it would include something like the sbatch manual, but with some examples of the "humantime" format.

Kobzol:
https://github.com/It4innovations/hyperqueue/pull/836 this unifies the syntax everywhere to use the same formats, and adds examples to the CLI help output.

PR #836, opened 2025-02-25
Unify format of all durations in CLI and add examples to them

Kobzol:
This PR unifies all duration parsers in the CLI to support both `HH:MM:SS` and humantime formats (to be consistent), and also adds a shared documentation string to them, to provide examples.

Fixes: https://github.com/It4innovations/hyperqueue/issues/832

### DIFF
--- a/crates/hyperqueue/src/client/commands/mod.rs
+++ b/crates/hyperqueue/src/client/commands/mod.rs
@@ -6,3 +6,24 @@ pub mod server;
 pub mod submit;
 pub mod wait;
 pub mod worker;
+
+/// Helper macro for generating CLI help for a `Duration` (or `Option<Duration>`) value
+/// that can be specified either using the HMS or humantime formats.
+macro_rules! duration_doc {
+    ($text:expr) => {
+        concat!(
+            $text,
+            "\n\n",
+            r#"You can use either the `HH:MM:SS` format or a "humantime" format.
+For example:
+- 01:00:00 => 1 hour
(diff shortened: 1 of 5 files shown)


crates/hyperqueue/src/client/commands/mod.rs
```diff
@@ -6,3 +6,23 @@ pub mod server;
...
+        concat!(
+            $text,
+            "\n\n",
+            r#"You can use either the `HH:MM:SS` format or a "humantime" format.
+For example:
+- 01:00:00 => 1 hour
+- 02:05:10 => 2 hours, 5 minutes, 10 seconds
+- 1h => 1 hour
+- 2h5m10s => 2 hours, 5 minutes, 10 seconds
+- "3h 10m 5s" => 3 hours, 10 minutes, 5 seconds (with spaces, wrapped in quotes)"#
```
spirali, line 23: Quotes around time specification containing spaces depends on shell, isn't it?
I can also write "--heartbeat=3h 2s", but I cannot write '--heartbeat="3h 2s"'
Therefore claim that time specification with space has to be contain quotes is misleading.

Btw: You can also use full names of time units:
"2 hours 3 minutes"

Kobzol: Ok, fair., removed the quotes from the example and removed the whole parentheses section. I hope that people will realize that they need to quote the time somehow when it contains spaces :)

spirali (approved):

PR #841, opened 2025-03-07
Adds flag --history-only to "hq journal stream"

spirali:
Streaming only past events from a live server without waiting for new ones.


Kobzol (commented):
I would modify the CLI so that we have:
- `replay` - only streams the history
- `stream` - only streams new events
- `stream --from-start` - streams both history and new events (`--from-start` is bikesheddable)

Otherwise looks fine.

spirali:
I am not sure how much just live stream without history is useful. But I like the name "replay" for history-only command.

Kobzol:
Ok, no problem with just having replay and stream.

Kobzol (approved):

Issue #830, opened 2025-02-22
hq alloc add does not respect --output-mode json


lahwaacz:
```
$ hq --output-mode json alloc add slurm --time-limit 1:00:00
2025-02-22T15:41:40Z INFO A trial allocation was submitted successfully. It was immediately canceled to avoid wasting resources.
2025-02-22T15:41:40Z INFO Allocation queue 2 successfully created
```

Since `--output-mode` is a global option, it should be respected by all subcommands.

Kobzol:
This is just a log printed to stderr, right? This command doesn't have any output (that would be printed to stdout in JSON), its only response is the process exit code.

lahwaacz:
Hmm, right. Unfortunately an empty string is not a valid JSON so I was expecting that it would give `{}` or `[]`.

tdadadavid:
Hi, @Kobzol I would love to contribute/fix this, can this be assigned to me?
I might need some help, thanks

spirali:
Thanks for help. Feel free to ask.

tdadadavid:
Hi @spirali I am trying to push my work (not done though) but I having permission error, 
```markdown
ERROR: Permission to It4innovations/hyperqueue.git denied to tdadadavid.
```
Please what am I missing, thanks.

spirali:
Hi,

Fork the repository by clicking on:

![Image](https://github.com/user-attachments/assets/91d6d766-22ee-433b-8abb-105433562a92)

Then push your changes into your fork of the repository.

Then Github will offer you to create a PR to our repository from your branch.

tdadadavid:
Ohh silly me thanks I appreciate

PR #844, opened 2025-03-08
Feat:  Add empty json output mode

tdadadavid:
[✅] Add `print_empty` to the `Output`trait
[✅] Implemented `print_empty` for `JSON`, `Cli` and `Quiet` 

Fixes: #830

### DIFF
--- a/crates/hyperqueue/src/client/output/cli.rs
+++ b/crates/hyperqueue/src/client/output/cli.rs
@@ -1177,6 +1177,8 @@ impl Output for CliOutput {
         output_file(&path);
         Ok(())
     }
+
+    fn finalize_output(&self) {}
 }
 
 struct AllocationTimes {
--- a/crates/hyperqueue/src/client/output/json.rs
+++ b/crates/hyperqueue/src/client/output/json.rs
@@ -1,3 +1,4 @@
+use std::cell::Cell;
 use std::collections::HashMap;
 use std::path::Path;
 use std::time::Duration;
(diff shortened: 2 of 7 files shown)


Kobzol (commented):
Hi, thanks for the PR! The `{}` should be only printed for the JSON printer, for the normal printer there's no point in printing an additional empty line.

Actually, it should be easy to solve this fully generally for all commands with this approach:
- Remember in the JSON printer if anything was printed.
- Add a `finalize_output` function to the printer, which will be called just before HQ exits (even if an error has occurred).
- Implement `finalize_output` for the JSON printer in a way where it will output `{}` if nothing was previously printed.

tdadadavid:
my apologies for the late reply, I am currently writing exams, I will be done next week and pick this up. Thanks!

crates/hyperqueue/src/client/output/json.rs
```diff
@@ -251,6 +251,10 @@ impl Output for JsonOutput {
     fn print_error(&self, error: Error) {
         self.print(json!({ "error": format!("{error:?}") }))
     }
+
+    fn finalize_output(&self) {
+        self.print(json!({}));
```
Kobzol, line 256: This should only be printed if we did not print anything in the JSON printer. Otherwise we will print multiple JSON objects, which is invalid JSON.

crates/hyperqueue/src/client/output/cli.rs
```diff
@@ -1034,6 +1034,10 @@ impl Output for CliOutput {
             job_id.to_string().color(colored::Color::Green),
         );
     }
+
+    fn finalize_output(&self) {
+        println!()
```
Kobzol, line 1039: We don't need to print anything here.

tdadadavid: Thanks will fix this.

Kobzol:
Hi, are you still interested in working on this issue? If not, I would finsih it.

tdadadavid:
Good day @Kobzol, I am not available to work on this. Apologies for not stating that earlier.

PR #847, opened 2025-03-14
Authentication and encryption can be disabled

spirali:
This code removes the previous ugly benchmarking hack that bypass security function very deep in the auth & encryption code. So even if you had a key, you could get insecure connection. It also used "HQ_" prefixed env variable in Tako code.

This PR removes the hack and when keys should not be used, it is not present in the memory at all.
You can directly tell from access record if connection will be secured. 

It introduces flags `--disable-client-authentication-and-encryption` and `--disable-worker-authentication-and-encryption` that allows to disable connection security (Note: it is intentionally very long)

You will also get warnings when server/worker/client is started and the connection is not secured.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 ### New features
 
 * Added `hq doc` command for accessing documentation about various HQ features from the command-line.
+* `hq journal replay` added. It similar to `hq journal stream` but it will not wait for new events.
+* More robust initialization of dashboard
+* Authentication and encyrption of client/worker connection can be disabled. It is mostly for testing 
+  and benchmarking purpose. Do not use if you are not in 100% safe environment.
 
 ### Breaking change
 * The Python API now requires Python 3.9, up from Python 3.6.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,10 +5,11 @@
 * Added `hq doc` command for accessing documentation about various HQ features from the command-line.
 * `hq journal replay` added. It similar to `hq journal stream` but it will not wait for new events.
 * More robust initialization of dashboard
-* Authentication and encyrption of client/worker connection can be disabled. It is mostly for testing 
+* Authentication and encryption of client/worker connection can be disabled. It is mostly for testing
   and benchmarking purpose. Do not use if you are not in 100% safe environment.
 
 ### Breaking change
+
 * The Python API now requires Python 3.9, up from Python 3.6.
 
 ## v0.21.1
@@ -20,7 +21,10 @@
 ## v0.21.0
 
 ### Breaking change
-* Pre-built HyperQueue releases available from our GitHub repository are now built with GLIBC `2.28`, instead of `2.17`. If you need to run HyperQueue on a system with an older GLIBC version, you might need to recompile it from source on your system. If you encounter any issues, please let us know.
+
+* Pre-built HyperQueue releases available from our GitHub repository are now built with GLIBC `2.28`, instead of `2.17`.
+  If you need to run HyperQueue on a system with an older GLIBC version, you might need to recompile it from source on
+  your system. If you encounter any issues, please let us know.
 
 ### Changes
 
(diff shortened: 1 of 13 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,10 @@
 ### New features
 
 * Added `hq doc` command for accessing documentation about various HQ features from the command-line.
+* `hq journal replay` added. It similar to `hq journal stream` but it will not wait for new events.
+* More robust initialization of dashboard
+* Authentication and encyrption of client/worker connection can be disabled. It is mostly for testing
```
Kobzol, line 8:
```suggestion
* Authentication and encryption of client/worker connection can be disabled. It is mostly for testing 
```

crates/hyperqueue/src/client/commands/server.rs
```diff
@@ -104,6 +104,18 @@ struct ServerStartOpts {
     /// Path to access file that is used for configuration of secret keys and ports
     #[arg(long)]
     access_file: Option<PathBuf>,
+
+    /// If set, client connection will NOT be AUTHENTICATED and ENCRYPTED.
+    /// ANYONE CAN CONNECT TO THE SERVER AS CLIENT!
+    /// USE ON YOUR OWN RISK.
```
Kobzol, line 110:
```suggestion
    /// USE AT YOUR OWN RISK.
```

crates/hyperqueue/src/client/commands/server.rs
```diff
@@ -104,6 +104,18 @@ struct ServerStartOpts {
...
+    #[arg(long)]
+    disable_client_authentication_and_encryption: bool,
+
+    /// If set, worker connection will NOT be AUTHENTICATED and ENCRYPTED.
+    /// ANYONE CAN CONNECT TO THE SERVER AS WORKER!
+    /// USE ON YOUR OWN RISK.
```
Kobzol, line 116:
```suggestion
    /// USE AT YOUR OWN RISK.
```

crates/hyperqueue/src/common/serverdir.rs
```diff
@@ -82,20 +82,28 @@ fn resolve_active_directory(path: &Path) -> PathBuf {
 }
 
 fn serde_serialize_key<S: Serializer>(
-    key: &Arc<SecretKey>,
+    key: &Option<Arc<SecretKey>>,
```
Kobzol, line 85: nit: it's more idiomatic to pass `Option<&` than `&Option<`.

crates/hyperqueue/src/common/serverdir.rs
```diff
@@ -82,20 +82,28 @@ fn resolve_active_directory(path: &Path) -> PathBuf {
...
     serializer: S,
 ) -> Result<S::Ok, S::Error> {
-    let str = serialize_key(key);
-    serializer.serialize_str(&str)
+    if let Some(key) = key {
+        let str = serialize_key(key);
+        serializer.serialize_str(&str)
```
Kobzol, line 90: Shouldn't this be `serialize_some`? :thinking:

crates/tako/src/internal/server/rpc.rs
```diff
@@ -95,10 +93,6 @@ pub(crate) async fn worker_authentication(
         &mut reader,
     )
     .await?;
```
Kobzol, line 95: The `assert` wasn't useful?

spirali:
I am aware of the circumstances :)

crates/hyperqueue/src/common/serverdir.rs
spirali: I think it does not matter, but you are right serializing an Option should be probably more elegant.

crates/tako/src/internal/server/rpc.rs
Kobzol: That if you have `Some(key)`, `do_authentication` returned `Some(sealer)`. But it's probably not needed.

crates/hyperqueue/src/common/serverdir.rs
Kobzol: If it works without it, then it's fine.

PR #849, opened 2025-03-14
Add command for deploying workers using SSH

Kobzol:
This was way more complicated than I originally thought, mainly because of the passthrough command-line handling, pff.

### DIFF
--- a/crates/hyperqueue/src/common/cli.rs
+++ b/crates/hyperqueue/src/common/cli.rs
@@ -262,7 +262,7 @@ pub enum WorkerCommand {
     ///
     /// Note that to use this command, an OpenSSH-compatible `ssh` binary
     /// has to be available on the local node.
-    DeploySSH(DeploySSHOpts),
+    DeploySsh(DeploySshOpts),
 }
 
 #[derive(Parser)]
@@ -309,7 +309,7 @@ pub struct HwDetectOpts {
 }
 
 #[derive(Parser, Debug)]
-pub struct DeploySSHOpts {
+pub struct DeploySshOpts {
     /// Show log output of the spawned worker(s).
     #[arg(long)]
     pub show_output: bool,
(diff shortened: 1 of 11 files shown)


crates/hyperqueue/src/common/cli.rs
```diff
@@ -303,6 +308,25 @@ pub struct HwDetectOpts {
     pub no_hyper_threading: bool,
 }
 
+#[derive(Parser, Debug)]
+pub struct DeploySSHOpts {
```
spirali, line 312: I believe that in Rust naming convention it should `DeploySshOpts` (for example `TcpListener`).

crates/hyperqueue/src/common/cli.rs
```diff
@@ -303,6 +308,25 @@ pub struct HwDetectOpts {
...
+    /// Show log output of the spawned worker(s).
+    #[arg(long)]
+    pub show_output: bool,
+
+    /// Path to a nodefile that contains the hostnames to which should
+    /// HQ workers be deployed to.
+    /// Each line in the file should correspond to one hostname address.
```
spirali, line 319: Can I add a hostname with port in case that SSH is not running on a standard port? (e.g. "myserver:1234")

docs/deployment/worker.md
```diff
@@ -72,6 +72,30 @@ Example submission script:
...
 
+#### Deploying a worker using SSH
+
+If you have an OpenSSH-compatible `ssh` binary available in your environment, HQ can deploy workers to a set of hostnames using the `deploy-ssh` command:
+
+```bash
+$ hq worker deploy-ssh <nodefile> <worker-args>
+```
+
+For example, if we had the following content stored in file called `nodefile.txt`:
```
spirali, line 83: The documentation should probably explicitly say that `node1` is a hostname. I am not sure that it is a common knowledge what "nodefile" means.

spirali: Also maybe calling it `hostfile` would be more descriptive.

Kobzol: Changed documentation and renamed it to `hostfile`.

crates/hyperqueue/src/common/cli.rs
Kobzol: Good idea, added support for that.

spirali (approved):

Issue #852, opened 2025-03-16
Expose `--crash-limit` in Python API


lahwaacz:
The `hq submit` subcommand has a [--crash-limit parameter](https://github.com/It4innovations/hyperqueue/blob/1fd4290e7c3eb002f76375f1c2f22d4618306849/docs/jobs/jobs.md?plain=1#L452-L463), but I can't find an equivalent in the Python API. It seems that only [max_fails](https://it4innovations.github.io/hyperqueue/stable/python/apidoc/hyperqueue.job.Job.html#f___init__) is exposed in the `Job.__init__` method.

Kobzol:
The Python API has blind spots, this is one of them. I'll add it.

Kobzol:
https://github.com/It4innovations/hyperqueue/pull/856 exposes the crash limit in the Python API. Note that it is actually a property of a *task*, not a *job*. It's just that `hq submit` can only set a given task property for all tasks at once.

PR #856, opened 2025-03-21
Add crash limit to Python API

Kobzol:
Fixes: https://github.com/It4innovations/hyperqueue/issues/852


spirali (approved):

lahwaacz:
Thanks! I see that `crash_limit` has the type `u32` in rust, but Python `int` has [unlimited precision](https://docs.python.org/3/library/stdtypes.html#numeric-types-int-float-complex). Do you handle the overflow somehow?

Kobzol:
You will get a very loud warning, it will crash :)

lahwaacz:
> You will get a very loud warning, it will crash :)

OK, that sounds good :laughing:

PR #858, opened 2025-03-25
Protection against OOM when reading corrupted bincode

spirali:


Kobzol (commented):
Thanks! Do you think that we can [write a test](https://github.com/Kobzol/kobzol.github.io/pull/38) for this? :D

spirali:
It is actually quite difficult, it seems that we need to write a large number on position where size of collection is written.
I can craft manually the file, but it would need to be be updated whenever the journal format changes. 

I have put on my roadmap migration to bincode 2.0, so I will think about it again when I will do the update, as journal format would also need an update.

Kobzol:
I think that we could use a custom serialize impl that would write a bogus size and then just small data.

Kobzol:
Also, when we load 2 (or N) *identical* events in a row, we should probably report an error (or at least a warning), because it's highly unlikely that several identical events (including identical timestamps) will be stored in the journal.

Issue #863, opened 2025-03-29
Memory not released by hq-server after running hq job forget


bguo068:
Hi `HyperQueue` team,

First off, thank you for the excellent work on `HyperQueue`. It has been working great for me -- especially when used with a compatible version of `Nextflow`. `HyperQueue` has really improved our workflow pipeline by efficiently managing many small tasks, which is important for us since our cluster penalizes large numbers of small jobs submitted individually.

I’ve encountered a potential issue regarding memory usage on the `hq server`. Over time, as more jobs are submitted, the server’s memory usage keeps increasing. For example, our current hq-server process is using over 4.6 GB of RAM on the login node, which accounts for nearly 30% of the node’s total memory.

I tried running:
```sh
hq job forget --filter finished,failed,canceled all
```

This appears to help temporarily -- newly submitted jobs don’t cause a further increase in memory usage for a while. However, it seems that the memory already used by the hq-server process is not released after the forget operation.

I’m wondering if this is expected behavior or a potential area for improvement. Conceptually, if job metadata is stored in a structure like `Vec<JobInfo>`, maybe the vector is being cleared without calling `shrink_to_fit()` or similar? That could explain why memory usage doesn’t drop even after forgotten jobs are removed.

Would it be feasible to add **an option** to explicitly release unused memory after forgotten jobs are removed? I understand memory management can be complex and sometimes OS-level caching is involved, but any insight would be appreciated.

Thanks again for all your hard work on this project!

Best,
Bing

Kobzol:
Hi, thanks for the issue report. HQ uses the `jemalloc` allocator by default (unless you recompile HQ yourself without it), which might be a bit reluctant in returning memory the OS, indeed.

That being said, we weren't calling `shrink_to_fit` before, and it sounds like a good idea. Implemented that in https://github.com/It4innovations/hyperqueue/pull/865.

bguo068:
Thank you for your reply. I believe the additional changes and suggested recompilation make sense. I will try the allocator suggestion and see how it works.

spirali:
Btw: How many workers were connected in your case? We are offering forgetting the tasks but not workers. These data are not big, but they still may grow.

@Kobzol Maybe we should provide some command "forget everything what is not running/waiting" that would also forget the workers.

Kobzol:
@bguo068 Could you please check if this issue still persists with the latest [nightly release](https://github.com/It4innovations/hyperqueue/releases/tag/nightly)? Thank you!

[...]

PR #865, opened 2025-03-31
Try to release memory after forgetting jobs

Kobzol:
I tried to write a test for this, but it is super slow to create a large number of jobs in the Python tests (especially since we test HQ in debug mode), and it is also very tricky to make any assertions about RSS.

Related issue: https://github.com/It4innovations/hyperqueue/issues/863


spirali:
I have extend your PR by also shrinking Tako's structures. If tasks are submitted at once, Tako's internal may also allocate large maps memory.

spirali:
If you do not find any problem in my code, merge as you wish.

PR #866, opened 2025-03-31
Extended worker info

spirali:
`hq worker info` now shows state and runtime info (a information from the scheduler about assigned and running tasks).

Idea occured in #784.

### DIFF
--- a/crates/hyperqueue/src/client/output/cli.rs
+++ b/crates/hyperqueue/src/client/output/cli.rs
@@ -1,12 +1,12 @@
 use cli_table::format::{Justify, Separator};
-use cli_table::{print_stdout, Cell, CellStruct, Color, ColorChoice, Style, Table, TableStruct};
+use cli_table::{Cell, CellStruct, Color, ColorChoice, Style, Table, TableStruct, print_stdout};
 
 use std::fmt::{Display, Write};
 use std::io::Write as write;
 
 use crate::client::job::WorkerMap;
-use crate::client::output::outputs::{Output, OutputStream, MAX_DISPLAYED_WORKERS};
-use crate::client::status::{get_task_status, job_status, Status};
+use crate::client::output::outputs::{MAX_DISPLAYED_WORKERS, Output, OutputStream};
+use crate::client::status::{Status, get_task_status, job_status};
 use crate::common::env::is_hq_env;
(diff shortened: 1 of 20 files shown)


Kobzol (commented):
This is great, thanks! I think that two useful pieces of information would also be:
- When was the last time a task was attempted to be computed? When you have a worker that repeatedly starts tasks, but they fail (or complete quickly), the worker might not have any tasks at a given moment, but seeing when it started the last task could be useful to debug why it is not idle.
- When was the last time a task was finished?

spirali:
Added time and task id of the last started task. Because of some complications I did not add last completed tasks; I am postponing it for larger refactoring of gateway connection (planned after merging datalayer).

crates/hyperqueue/src/client/output/cli.rs
```diff
@@ -363,7 +344,42 @@ impl Output for CliOutput {
                     .unwrap_or("N/A")
                     .cell(),
             ],
+            vec![
+                "Last task started".cell().bold(true),
+                last_task_started
+                    .map(|t| {
+                        format!("Job: {}; Task: {}; Time: {}", t.job_id, t.task_id, t.time).cell()
```
Kobzol, line 351:
```suggestion
                        format!("Job: {}; Task: {}; Time: {}", t.job_id, t.task_id, format_datetime(t.time)).cell()
```

crates/hyperqueue/src/client/output/cli.rs
```diff
@@ -363,7 +344,42 @@ impl Output for CliOutput {
...
+                    running_tasks,
+                    assigned_tasks,
+                    is_reserved,
+                } => {
+                    write!(s, "Assigned tasks: {}", assigned_tasks).unwrap();
+                    if running_tasks > 0 {
+                        write!(s, "; Running tasks: {}", running_tasks).unwrap();
+                    }
+                    if is_reserved {
+                        write!(s, "; reserved").unwrap();
```
Kobzol, line 369:
```suggestion
                        write!(s, "; reserved for a multi-node task").unwrap();
```
`reserved` on its own isn't super clear :)

Kobzol (commented):
Thanks, looks great! Left two nits.

PR #867, opened 2025-04-02
Deps updated

spirali:


Kobzol (approved):
7 new dependencies added for `cargo build` :/ I can try to use https://crates.io/crates/cargo-duplicates later to try to remove some of them.

spirali:
We are keeping older version of psutils and nom, that may depends on some older versions. Also we are using rand in very basic way, so disabling some features could help.

Issue #864, opened 2025-03-30
Only send worker overviews on demand by default


Kobzol:
Currently, workers gather and send HW statistics by default to the server. This costs a bit of performance and network bandwidth  and if journalling is enabled, also a lot of disk space.

We should disable the overviews by default, and make them opt-in. To make dashboard work more easily, when dashboard is connected, the server should tell workers to temporarily start gathering the HW statistics. When all dashboards disconnect, the server should disable it again.

PR #871, opened 2025-04-04
Reconstruct events from current server state when streaming without a journal

Kobzol:
I wanted to work on https://github.com/It4innovations/hyperqueue/issues/864, but as usually, I got sidetracked.. When I opened the dashboard, I realized that it does not show past events if the server doesn't use a journal. This PR improves that.

### DIFF
--- a/crates/hyperqueue/src/server/client/mod.rs
+++ b/crates/hyperqueue/src/server/client/mod.rs
@@ -82,8 +82,16 @@ async fn stream_history_events<Tx: Sink<ToClientMessage, Error = HqError> + Unpi
     mut history: mpsc::UnboundedReceiver<Event>,
 ) {
     log::debug!("Resending history started");
-    while let Some(e) = history.recv().await {
-        if tx.send(ToClientMessage::Event(e)).await.is_err() {
+
+    let mut events = Vec::with_capacity(1024);
+    let capacity = events.capacity();
+    while history.recv_many(&mut events, capacity).await != 0 {
+        let events = std::mem::replace(&mut events, Vec::with_capacity(capacity));
+        if tx
+            .send_all(&mut futures::stream::iter(events).map(|e| Ok(ToClientMessage::Event(e))))
+            .await
+            .is_err()
+        {
             return;
         }
     }
--- a/crates/hyperqueue/src/server/client/mod.rs
+++ b/crates/hyperqueue/src/server/client/mod.rs
@@ -4,22 +4,22 @@ use std::sync::Arc;
 
 use futures::{Sink, SinkExt, Stream, StreamExt};
 use orion::kdf::SecretKey;
-use tokio::net::{TcpListener, TcpStream};
-use tokio::sync::{Notify, mpsc};
-
 use tako::gateway::{CancelTasks, FromGatewayMessage, StopWorkerRequest, ToGatewayMessage};
 use tako::{Set, TaskGroup};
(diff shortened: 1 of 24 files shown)


crates/hyperqueue/src/client/commands/submit/jobfile.rs
```diff
@@ -140,6 +141,7 @@ fn build_job_submit(jdef: JobDef, job_id: Option<JobId>) -> SubmitRequest {
             max_fails: jdef.max_fails,
         },
         submit_desc: JobSubmitDescription {
+            created_at: Utc::now(),
```
spirali, line 144: This timestamp should be filled at the server, because clock on the server is the ground truth for event time.

So it should probably be a part of the `Job` in the server, not `JobSubmitDescription` (as something that is coming from the client).

spirali (changes requested):

crates/hyperqueue/src/server/client/mod.rs
```diff
@@ -226,6 +238,143 @@ pub async fn client_rpc_loop<
...
+        tx.send(Event::at(
+            worker.started_at(),
+            EventPayload::WorkerConnected(*id, Box::new(worker.configuration().clone())),
+        ))?;
+
+        // Worker end
+        let info = worker.make_info(None);
+        if let Some(ended) = info.ended {
+            tx.send(Event::at(
+                ended.ended_at,
```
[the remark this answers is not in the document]
Kobzol: Right, I added sorting in the dashboard to get around this, but you're right that other clients might depend on the ordering. The problem with buffering is that.. we have to buffer everything in memory :) But hopefully when we reconstruct the state, it should take that much memory.

crates/hyperqueue/src/client/commands/submit/jobfile.rs
Kobzol: Hmm, good point. A bit annoying that this is *per submit*, so I'll need to store something like `SmallVec<[(Time, JobSubmitDescription); N]` on the server.

crates/hyperqueue/src/server/client/mod.rs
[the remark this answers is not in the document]
spirali: If client can sort, then server can also sort :)

In case of problem we can also implement "partial sorts". Gather all workers data and then have 2 queues for started/finished tasks and do a merge like in mergesort. But it is probably not worth it.

crates/hyperqueue/src/client/commands/submit/jobfile.rs
spirali: Or we can have separate SmallVec with timestamps, but `SmallVec<[(Time, JobSubmitDescription); N]` is not nice but not that bad.

Kobzol: Ok, change the storage to keep the submission date on the server.

spirali:
It seems good now, I am just afraid about last thing in reconstruction:

First it adds "worker lost event" and after that "task finished/failed event", because of the stable sort, if both events have exactly same time then worker lost is before task finished despite it had to happened in the opposite way. I am not sure if it is may hurt dashboard. Easy fix should be, put "worker lost" events after processing messages.

Kobzol:
Ok, changed.

spirali (approved):

Issue #873, opened 2025-04-06
HQ crashes


lahwaacz:
`hq server` just crashed on me:

```
[2025-04-06T18:04:54.261Z DEBUG tako::internal::server::task] New task 30064772574 ResourceRequestVariants { variants: [ResourceRequest { n_nodes: 0, resources: [ResourceRequestEntry { resource_id: ResourceId(0), request: Compact(8) }], min_time: 84600s }] }
[2025-04-06T18:04:54.261Z DEBUG tako::internal::server::task] New task 30064772575 ResourceRequestVariants { variants: [ResourceRequest { n_nodes: 0, resources: [ResourceRequestEntry { resource_id: ResourceId(0), request: Compact(8) }], min_time: 84600s }] }
[2025-04-06T18:04:54.261Z DEBUG tako::internal::server::task] New task 30064772576 ResourceRequestVariants { variants: [ResourceRequest { n_nodes: 0, resources: [ResourceRequestEntry { resource_id: ResourceId(0), request: Compact(8) }], min_time: 84600s }] }
[2025-04-06T18:04:54.261Z DEBUG tako::internal::server::task] New task 30064772577 ResourceRequestVariants { variants: [ResourceRequest { n_nodes: 0, resources: [ResourceRequestEntry { resource_id: ResourceId(0), request: Compact(8) }], min_time: 84600s }] }
[2025-04-06T18:04:54.261Z DEBUG tako::internal::server::task] New task 30064772578 ResourceRequestVariants { variants: [ResourceRequest { n_nodes: 0, resources: [ResourceRequestEntry { resource_id: ResourceId(0), request: Compact(8) }], min_time: 84600s }] }
[2025-04-06T18:04:54.261Z DEBUG tako::internal::server::task] New task 30064772579 ResourceRequestVariants { variants: [ResourceRequest { n_nodes: 0, resources: [ResourceRequestEntry { resource_id: ResourceId(0), request: Compact(8) }], min_time: 84600s }] }
[2025-04-06T18:04:54.261Z DEBUG tako::internal::server::task] New task 30064772580 ResourceRequestVariants { variants: [ResourceRequest { n_nodes: 0, resources: [ResourceRequestEntry { resource_id: ResourceId(0), request: Compact(8) }], min_time: 84600s }] }
[...]

PR #879, opened 2025-04-07
Fix restoration tasks with dependencies

spirali:
This fixes journal loading problem in #873

### DIFF
--- a/tests/test_journal.py
+++ b/tests/test_journal.py
@@ -6,7 +6,7 @@ from .autoalloc.mock.mock import MockJobManager
 from .autoalloc.flavor import all_flavors, ManagerFlavor
 from .autoalloc.utils import add_queue, remove_queue
 from .conftest import HqEnv
-from .utils import wait_for_job_state
+from .utils import wait_for_job_state, wait_for_task_state
 import os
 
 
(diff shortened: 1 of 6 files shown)


tests/test_journal.py
```diff
@@ -342,3 +342,31 @@ def collect_ids():
...
+    hq_env.start_worker()
+    tmp_path.joinpath("job.toml").write_text(
+        """
+[[task]]
+id = 1
+command = ["sleep", "0"]
+
+[[task]]
+id = 2
+command = ["sleep", "2"]
```
Kobzol, line 359: Let's use some custom task requirements that will only be provided by the second worker in the test, to avoid race conditions (`wait_for_task_state` polls in some intervals, and in theory the task could be finished before we stop the server). Also we could check that the task was actually executed (e.g. by writing something to disk). The job could be marked as finished after restoration without the tasks actually running.

tests/utils/wait.py
```diff
@@ -65,6 +65,24 @@ def wait_for_worker_state(env, ids: Union[int, List[int]], target_states: Union[
...
+    if isinstance(states, str):
+        states = [states]
+    assert len(task_ids) == len(states)
+
+    ids = ",".join(str(t) for t in task_ids)
+    states = [s.lower() for s in states]
+
+    def check():
+        result = env.command(["--output-mode=json", "task", "info", str(job_id), ids], as_json=True)
+        return [r["state"] for r in result] == states
```
Kobzol, line 81: Hmm, I wonder if we promise to return the JSON response in the same order as were the input arguments :sweat_smile:

Kobzol (commented):
Maybe event sourcing wasn't such a bad idea fater all :sweat_smile:

tests/test_journal.py
spirali: Fixed

tests/test_journal.py
```diff
@@ -342,3 +342,70 @@ def collect_ids():
...
+    tmp_path.joinpath("sleep_time").write_text("100")
+    tmp_path.joinpath("job.toml").write_text(
+        """
+[[task]]
+id = 1
+command = ["touch", "a"]
+
+[[task]]
+id = 2
+command = ["bash", "-c", "sleep `cat sleep_time` && touch b"]
```
Kobzol, line 360: Clever!

Kobzol (approved):

PR #885, opened 2025-04-09
Fixed invalid semantics for finished tasks

spirali:
This fixes problem in #873. 

The root cause was the unsound semantics when task is removed from Tako. 
When t2 depends on t1 then t2.deps == [t1] and t1.consumers == [t2]. 
It caused crash when a new submit occurs, because `t.deps` could sometimes points to removed tasks even delete mechanism tries to prevent this.

The current version now works as follows:
1) if a `task_id` is in `t.consumers` then `core.tasks` have contains task_id (this is not changed)
2) if a `task_id` is in `t.deps`, then task_id may NOT be in `core.tasks`, if it happens, it should be considered be as a resolved dependency and ignored. (this is changed)

This PR also fixes a forgotten heuristic from RSDS, that makes some sense for data transfers, but has no meaning for pure data dependencies.

### DIFF
--- a/crates/tako/src/internal/scheduler/metrics.rs
+++ b/crates/tako/src/internal/scheduler/metrics.rs
@@ -1,7 +1,7 @@
+use crate::TaskId;
 use crate::internal::common::{Map, Set};
 use crate::internal::server::task::Task;
 use crate::internal::server::taskmap::TaskMap;
-use crate::TaskId;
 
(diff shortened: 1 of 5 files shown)


spirali:
I would be happy to do a white board exercise when you arrive in the office:)

lahwaacz:
I don't understand anything from the internals, but it's awesome that you managed to find and solve the issue :tada:

PR #887, opened 2025-04-11
Disable worker overview by default and make it temporarily overridable when streaming events

Kobzol:
Fixes: https://github.com/It4innovations/hyperqueue/issues/864

### DIFF
--- a/crates/hyperqueue/src/client/commands/journal/mod.rs
+++ b/crates/hyperqueue/src/client/commands/journal/mod.rs
@@ -64,6 +64,7 @@ async fn stream_json(gsettings: &GlobalSettings, live_events: bool) -> anyhow::R
         .connection()
         .send(FromClientMessage::StreamEvents(StreamEvents {
             live_events,
+            enable_worker_overviews: false,
         }))
         .await?;
     let stdout = std::io::stdout();
--- a/crates/hyperqueue/src/client/commands/worker.rs
+++ b/crates/hyperqueue/src/client/commands/worker.rs
@@ -294,10 +294,10 @@ fn gather_configuration(opts: WorkerStartOpts) -> anyhow::Result<WorkerConfigura
             Some(duration)
         }
     });
-    let overview_configuration = send_overview_interval.map(|interval| OverviewConfiguration {
-        send_interval: interval,
+    let overview_configuration = OverviewConfiguration {
+        send_interval: send_overview_interval,
         gpu_families,
-    });
+    };
 
     Ok(WorkerConfiguration {
         resources,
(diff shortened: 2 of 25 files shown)


spirali (commented):
Looks good!

I just do not like that when dashboard is running, journal starts to grow. So you have random overview events in journal, but only for times when dashboard was enabled.

# Proposal

I propose the following CLI:

Start journaling, no overview events are stored no matter if dashboard is enabled or not:

`hq server start --journal x` 

Start journaling, journal gathers overview events every 10s:

`hq server start --journal x --overview-interval 10s`

Opens dashboard and starts to gather overviews every 2s (if `hq server start --journal x --overview-interval XX` was used, then interval in the worker is temporarily increased to 2s if XX was bigger than 2s):

`hq dashboard` 

With this, we can completely remove `--overview-interval` as an argument on worker. It will be completely controlled by server.

# Conclusion

I am aware that it is non trivial extension. I am quite ok to takes this PR as it is, if we can somehow prevent writing overview into journal just because dashboard is open, even it would mean to temporarily disable writing this event into journal at all.

Kobzol:
Adding a default `--overview-interval` to the server would be useful (same as for the idle timeout), but I think that it should still be overridable per worker, e.g. you can have a GPU node where the HW detection is broken and keeps spamming the log, so you want to disable it there, or vice versa you might want to gather HW stats only on GPU nodes, as an example.

Regarding the rest, I would really like to make `hq server start` + `hq worker start` + `hq dashboard` "just work" and show worker overviews. Requiring a flag for this to work is not very discoverable, and also it can be quite difficult for "just-in-time" monitoring. If you start the server and then after a few hours/days you want to debug something or monitor the state of the cluster, you probably don't want to restart the server just to enable worker overviews.

I would propose an alternative. Each worker knows if the overview interval was enabled explicitly or if it's only temporarily overridden because of dashboard. We could examine this property, and avoid storing overview events into the journal if they were only generated because of a temporary streaming/dashboard request. In that case we won't store these temporary events into the journal.

But I'm not sure if this is very good UX, because it means that even if you use the journal, then by default if you start the dashboard, quit it, and restart it again, you will not see HW overview history that you saw in the first instance of the dashboard.

spirali:
I think that we do not understand each other. I definitely want that "hq server start" + "hq worker start" + "hq dashboard" works without any flags ("works" here means that user see overviews in dashboard, no overview is stored in the journal). My point was that storing overview events in journal and showing them in dashboard should be completely independent in terms of configuration (including with independent intervals if they are both enabled together).

Re UX: I think that is it easy to explain that if the journal is off, no history is stored. Showing random parts of the history when dashboard was running in past seems to me much more confusing.

Re worker interval: You are right, I agree that worker may override interval. But setting `worker start --overview-interwal` should serve as a kind of lowerbound that can be asked from server. If no one wants overview (journal nor dashboard) than worker should not collect overviews just becase it was started with a flag configuring this.

Kobzol:
Added a check to avoid storing "stream-only" worker overviews into the journal. It seems to work well.

Issue #862, opened 2025-03-28
Naming jobs from Python API and showing task names in the dashboard


lahwaacz:
It seems that `hq submit` allows to set a name for the submitted jobs which is then shown in the dashboard:

```
$ hq submit --help
Submit a job to HyperQueue

Usage: hq submit [OPTIONS] <COMMANDS>...

Arguments:
  <COMMANDS>...
          Command that should be executed by each task

Options:
      --name <NAME>
          Name of the job
...
```

However, the Python API does not allow this: there is no `name` parameter in [`Job.__init__()`](https://it4innovations.github.io/hyperqueue/stable/python/apidoc/hyperqueue.job.Job.html#f___init__).

On the other hand, the `Job.function` and `Job.program` functions have a `name` parameter, but this names individual tasks rather than the whole job. Also, task names are not shown in the dashboard (AFAICT).

Kobzol:
Task names are indeed a "Python API-only" concept, they are stored purely for debugging reasons on the client and the server has no such concept. I'll expose job names in the Python API.

PR #888, opened 2025-04-11
Allow configuring job name in Python API

Kobzol:
Fixes: https://github.com/It4innovations/hyperqueue/issues/862


spirali (approved):

spirali (commented):
Review for another PR was placed here. I am accepting the change without comments.

PR #894, opened 2025-04-16
Missing serde_bytes annotation

spirali:


Kobzol (approved):
Crap, there should be some lint for this.

spirali:
This one is quite innocent because it is just few bytes. But I have hit the wall in data transfers, because it takes tens of seconds to transfer 300MB without the annotation (in debug mode). So I rather checked all other structures :)

PR #895, opened 2025-04-25
Refactor autoalloc tests

Kobzol:
I can't even remember how many times have I already refactored this :man_facepalming: Hopefully this time it will last a bit longer.
There's now a single class that handles the manager logic, and it uses another class to handle the manager-specific logic (parsing of input, formatting of output and starting of workers).


spirali (approved):

spirali:
> I can't even remember how many times have I already refactored this 🤦‍♂️ Hopefully this time it will last a bit longer. There's now a single class that handles the manager logic, and it uses another class to handle the manager-specific logic (parsing of input, formatting of output and starting of workers).

Learning from mistakes:)

PR #898, opened 2025-04-29
TaskId unification

spirali:
Historically we had JobId/JobTaskId in HQ and TaskId in Tako. Last versions made simple bijection between them to avoid doing lookups. This PR unifies it completely, to avoid code duplication and better and unified outputs.

- JobId and JobTaskId is moved to Tako
- TaskId is now pair JobId/JobTaskId. For Tako it is still functionally opaque type, but it can provide better messages because it prints real task ID and not some huge number.
- In HQ, when both job_id/job_task_id was used, it is replaced by direct usage of TaskId. JobId or JobTaskId is used in places when we need to know only job id or job task id, but not both.
- Reimports Tako types in HQ was removed, HQ now uses Tako types directly; mainly TaskTakoId was removed and HQ directly uses tako::TaskId.


spirali:
Note: It actually removes more lines than it adds, it seems large only because it contains unmerged data layer

crates/tako/src/lib.rs
```diff
@@ -10,16 +10,20 @@ pub mod hwstats;
...
 pub use crate::internal::common::index::{AsIdVec, ItemId};
 pub use crate::internal::common::taskgroup::TaskGroup;
 pub use crate::internal::common::utils::format_comma_delimited;
+pub use crate::internal::common::WrappedRcRefCell;
 pub use crate::internal::common::{Map, Set};
 
+define_id_type!(JobId, u32);
+define_id_type!(JobTaskId, u32);
 define_id_type!(WorkerId, u32);
 define_id_type!(TaskId, u64);
```
Kobzol: Is the u64 `TaskId` still required for something?

crates/hyperqueue/src/client/commands/journal/output.rs
```diff
@@ -76,32 +76,26 @@ fn format_payload(event: EventPayload) -> serde_json::Value {
                 "allocation-id": allocation_id,
             })
         }
-        EventPayload::TaskStarted {
-            job_id, task_id, ..
-        } => json!({
+        EventPayload::TaskStarted { task_id, .. } => json!({
```
Kobzol, line 79: Maybe we should rename `task_id` fields here to just `id`?

crates/tako/src/internal/common/ids.rs
```diff
@@ -0,0 +1,46 @@
...
+
+#[derive(Default, Copy, Clone, Hash, PartialOrd, Ord, PartialEq, Eq, Serialize, Deserialize)]
+pub struct TaskId {
+    job_id: JobId,
+    job_task_id: JobTaskId,
+}
+
+impl Display for TaskId {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}@{}", self.job_id, self.job_task_id)
```
Kobzol, line 18: I would use `/` as the separator. We already used it on a couple of places before, and `@` evokes "location" to me, so `X@N` could be interpreted as task X on worker N.

Kobzol (approved):
Nice :)

crates/tako/src/internal/common/ids.rs
spirali: It was my first implementation, but because they are number, the result looks like a fraction: "1/2", so "/" is bad idea.

Kobzol: - `1-2`: looks like a range or a subtraction :)
- `[1, 2]`: looks like an array
- `1>2` ?
- `1#2`

crates/hyperqueue/src/client/commands/journal/output.rs
spirali: I would prefer, to have "task_id", because in many contexts we are having several ids at once, worker_id, job_id, task_id etc, and having explicit "task_id" seems to more clear while it is still a short name.

crates/tako/src/internal/common/ids.rs
spirali: You are following my steps:). Actually I was reserving "#" for data objects, 1@2#3 could be a name of data object in the future. Other reasonable options are: 1`2 or 1$2, but 1@2 seems better in my eyes.

crates/hyperqueue/src/client/commands/journal/output.rs
Kobzol: Ok, fair enough.

crates/tako/src/internal/common/ids.rs
Kobzol: `1$2#3@4` - Job 1, task 2, data object 3 at worker 4 :laughing: Ok, let's leave `@`.

PR #899, opened 2025-05-06
New scheduler priority assignment

spirali:
This PR changes changes how scheduler priority is calculated:
* b-level computaiton was changed to t-level computation
* It adds a small priority to tasks from older jobs (determined by job_id, not submission time).

# Effects

* This solves the problem discussed in #883.
* Because tasks are already sorted by topological order wrt dependencies, it is much cheaper to compute t-level than b-level.
* For usages without dependencies, HQ will slightly prioritize tasks from older jobs
* For usage with dependencies, when a component of tasks is started, HQ will prioritize its completion.
* I am not aware of any negative effect on pipelines that I am aware of, except of artificial adversary pipelines crafted specially for this case.
* The only thing I am afraid is some interference with "witness computation" in the worker; however I was not able to construct any counter example. Moreover, if there is such that, then there it can be probably used to create adversary example for the original solution.


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ### New features
 
+* The server scheduler now slightly prioritizes tasks from older jobs and finishing partially-computed task graph
```
Kobzol, line 5:
```suggestion
* The server scheduler now slightly prioritizes tasks from older jobs and finishing partially-computed task graphs.
```

crates/tako/src/internal/server/reactor.rs
```diff
@@ -169,17 +169,20 @@ pub(crate) fn on_new_tasks(core: &mut Core, comm: &mut impl Comm, new_tasks: Vec
     assert!(!new_tasks.is_empty());
     for mut task in new_tasks.into_iter() {
         let mut count = 0;
+        let mut priority = -(task.id.job_id().as_num() as i32);
         task.task_deps.retain(|t| {
             if let Some(task_dep) = core.find_task_mut(*t) {
                 task_dep.add_consumer(task.id);
                 if !task_dep.is_finished() {
+                    priority = std::cmp::max(priority, task_dep.scheduler_priority + 256);
```
Kobzol, line 177: Maybe provide some explanation/ideas behind 256 in a comment? :)

Kobzol (approved):
Kind of sad to see the venerable B-level go, but if it doesn't pull its weight, then it's a good thing. It doesn't seem to show anything on microbenchmarks.

Kobzol:
Thanks, LGTM :)

Issue #886, opened 2025-04-10
Limits on job mix


svatosFZU:
Hi,
I would like to ask if it is possible to put in a limit on one job kind. Until now, I was running only one kind of jobs (i.e. all jobs have the same requirements on CPU and memory). Now, I would like to try to send a mix of jobs. For example, to 128-core nodes of Karolina, I would like to send mix of 32-core jobs and single-core jobs. Leaving the HQ to just mix them without constraints will not work as too many of those single-core jobs could kill the worker node. So, is there a way to tell the HQ to mix the jobs to fill the node in such a way that there is only configurable number (32 or 64 would make sense in this case) of those single-core jobs?

spirali:
You can create an artificial resource that will limit the single core nodes.

Example: 

Create resources on worker (32 artifical resources "foo")

`hq worker start --resource "foo=sum(32)"`

If you submit task like:

`hq submit --resource foo=1 -- sleep 1`

Then it will need 1 cpu and 1 foo. So it will not run more than 32 of these tasks on the worker because it has only 32 foo resources.

spirali:
Yes, this is the right usage.

Btw: If your 1 cpu tasks are limited by the memory, you can also use:

`hq submit --resource mem=100` # This task needs 100MB of RAM.

And it will not start the task if the sum of the resource "mem" of all running tasks would exceed the RAM. So you do not need an artificial resource, because "mem" resource is detected automatically. But artificial resources are nothing bad. It is up to you.

svatosFZU:
Thanks, that is what I needed to know.
closing

svatosFZU:
Hi,
I would have one follow-up question related to this. I am testing it on Barbora (36-core WNs). The situation is simpler there as I do not need mixed jobs, i.e. I need to submit either one 36-core job or 36 single-core jobs. But I have difficulties to do the latter one. It seems that the HyperQueue is trying to submit the jobs even though it does not have enough to fill the WN. As result, I have HQ workers which ran only few single-core jobs. That is quite inefficient. Is there a way for me to tell the HyperQueue to submit allocation only if it has enough work to fill it completely?

Kobzol:
Could you please clarify what is WN? Just so that I don't miss anything important :)

[...]

PR #904, opened 2025-05-22
Extended query

spirali:
This PR updates scheduler query:
 * adds "min_utilization" to query (see #886)
 * Fix problems when query was performed on server with no worker.

### DIFF
--- a/crates/tako/src/internal/scheduler/query.rs
+++ b/crates/tako/src/internal/scheduler/query.rs
@@ -12,10 +12,9 @@ struct WorkerTypeState {
 
 /* Read the documentation of NewWorkerQuery in gateway.rs */
 pub(crate) fn compute_new_worker_query(
-    core: &Core,
+    core: &mut Core,
     queries: &[WorkerTypeQuery],
 ) -> NewWorkerAllocationResponse {
-    assert!(core.sn_ready_to_assign().is_empty()); // If there are read_to_assign tasks, first we have to call scheduling
     log::debug!("Compute new worker query: query = {:?}", queries);
 
     let add_task = |new_loads: &mut [WorkerTypeState], task: &Task| {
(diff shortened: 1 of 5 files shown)


crates/tako/src/internal/scheduler/query.rs
```diff
@@ -1,17 +1,60 @@
...
 use crate::internal::server::workerload::{WorkerLoad, WorkerResources};
 
-/* Read the documentation of NewWorkerQuery in gateway.rs */
+struct WorkerTypeState {
+    loads: Vec<WorkerLoad>,
+    w_resources: WorkerResources,
+    max: u32,
+}
+
+/* Read the documentation of WorkerTypeQuery in gateway.rs */
```
Kobzol, line 13: It's not `gateway.rs` anymore :)

crates/tako/src/internal/server/workerload.rs
```diff
@@ -266,6 +266,14 @@ impl WorkerLoad {
         }
         result
     }
+
+    pub(crate) fn utilization(&self, wr: &WorkerResources) -> f32 {
+        let mut utilization = 0.0f32;
+        for (n, w) in self.n_resources.iter().zip(wr.n_resources.iter()) {
+            utilization = utilization.max(n.as_f32() / w.as_f32());
```
Kobzol, line 273: What about this?
```rust
self.n_resources.iter().zip(wr.n_resources.iter())
            .map(|(n, w)| n.as_f32() / w.as_f32())
            .max()
            .unwrap_or(0.0)
```

Also, are we sure that `w` won't ever be zero? :laughing:

Kobzol (approved):
The interface looks reasonable, I'll try to use it in autoalloc. Left two nits.

Kobzol (approved):

PR #905, opened 2025-05-23
Batch canceling event

spirali:
Tasks are usually canceled in batches (e.g. whole job is canceled or failed tasks cancel all their successors). 
This PR modifies journal event for canceling: it now takes array of tasks instead of one task.
This ensures that task canceling is now stored in journal atomically and restoring crashed server will not load only partial cancellation.

### DIFF
--- a/crates/hyperqueue/src/client/commands/journal/output.rs
+++ b/crates/hyperqueue/src/client/commands/journal/output.rs
@@ -87,7 +87,7 @@ fn format_payload(event: EventPayload) -> serde_json::Value {
             "job": task_id.job_id(),
             "task": task_id.job_task_id(),
         }),
-        EventPayload::TaskCanceled { task_ids } => json!({
+        EventPayload::TasksCanceled { task_ids } => json!({
             "type": "task-canceled",
             "tasks": task_ids,
         }),
--- a/crates/hyperqueue/src/client/commands/journal/output.rs
+++ b/crates/hyperqueue/src/client/commands/journal/output.rs
@@ -87,10 +87,9 @@ fn format_payload(event: EventPayload) -> serde_json::Value {
             "job": task_id.job_id(),
             "task": task_id.job_task_id(),
         }),
-        EventPayload::TaskCanceled { task_id } => json!({
+        EventPayload::TaskCanceled { task_ids } => json!({
             "type": "task-canceled",
-            "job": task_id.job_id(),
-            "task": task_id.job_task_id(),
+            "tasks": task_ids,
         }),
         EventPayload::TaskFailed { task_id, error } => json!({
             "type": "task-failed",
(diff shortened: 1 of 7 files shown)


crates/hyperqueue/src/client/commands/journal/output.rs
```diff
@@ -87,10 +87,9 @@ fn format_payload(event: EventPayload) -> serde_json::Value {
             "job": task_id.job_id(),
             "task": task_id.job_task_id(),
         }),
-        EventPayload::TaskCanceled { task_id } => json!({
+        EventPayload::TaskCanceled { task_ids } => json!({
```
Kobzol, line 90:
```suggestion
        EventPayload::TasksCanceled { task_ids } => json!({
```

crates/hyperqueue/src/server/client/mod.rs
```diff
@@ -353,12 +353,22 @@ fn reconstruct_historical_events(
                     ));
                 }
                 JobTaskState::Canceled { cancelled_date, .. } => {
-                    events.push(Event::at(
-                        *cancelled_date,
-                        EventPayload::TaskCanceled {
-                            task_id: TaskId::new(job.job_id, *id),
-                        },
-                    ));
+                    if let Some(task_ids) = events.last_mut().and_then(|e| {
```
Kobzol, line 356: Why is this needed if the server should now generate cancels in batches?

crates/hyperqueue/src/server/event/payload.rs
```diff
@@ -48,7 +48,7 @@ pub enum EventPayload {
     },
     /// Task has been canceled
```
Kobzol, line 49:
```suggestion
    /// Tasks have been canceled. For performance and correctness reasons, this event is batched.
```

crates/hyperqueue/src/server/restore.rs
```diff
@@ -290,36 +290,39 @@ impl StateRestorer {
...
-                            task.state =
-                                match std::mem::replace(&mut task.state, JobTaskState::Waiting) {
-                                    JobTaskState::Running { started_data } => {
-                                        JobTaskState::Canceled {
-                                            started_data: Some(started_data),
-                                            cancelled_date: event.time,
+                EventPayload::TaskCanceled { task_ids } => {
+                    log::debug!("Replaying: TaskCanceled {task_ids:?}");
+                    for task_id in task_ids {
+                        if let Some(job) = self.jobs.get_mut(&task_id.job_id()) {
```
Kobzol, line 296: As a micro-optimization, we could store the last resolved job and check if the current `task_id` has the same job ID (since this will usually be the case). But cancelling shouldn't be performance critical, so it probably isn't worth it.

crates/hyperqueue/src/server/client/mod.rs
Kobzol: I wasn't sure why we needed to do the batching here, but now I understand, thanks.

Kobzol (approved):

Kobzol (approved):

PR #907, opened 2025-05-23
Task array data sharing

spirali:
Each task in Tako contains serialized payload of task data.
Previously, each task had its own payload even they came from the task array.

This PR changes it. Payload no longer contains task id and entry, so they can be shared across the task array.
It is first step of larger scheduler changes, but if @Kobzol has time, I would be curious if it has measurable performance implications.

### DIFF
--- a/crates/hyperqueue/src/transfer/messages.rs
+++ b/crates/hyperqueue/src/transfer/messages.rs
@@ -12,11 +12,12 @@ use crate::server::event::Event;
 use crate::server::job::{JobTaskCounters, JobTaskInfo, SubmittedJobDescription};
 use std::path::PathBuf;
 use std::time::Duration;
-use tako::gateway::{LostWorkerReason, ResourceRequestVariants, TaskDataFlags, WorkerRuntimeInfo};
+use tako::gateway::{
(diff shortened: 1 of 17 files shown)


crates/hyperqueue/src/transfer/messages.rs
```diff
@@ -138,7 +137,7 @@ pub enum JobTaskDescription {
     /// Either a single-task job or a task array usually submitted through the CLI.
     Array {
         ids: IntArray,
-        entries: Option<Vec<BString>>,
+        entries: Option<Vec<ThinVec<u8>>>,
```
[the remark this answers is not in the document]
Kobzol: Oh, sorry, I didn't realize `ThinVec` is just a single word. Nevermind!

crates/tako/src/internal/messages/worker.rs
```diff
@@ -40,6 +40,8 @@ pub struct ComputeTaskMsg {
 
     #[serde(with = "serde_bytes")]
     pub body: Box<[u8]>,
```
Kobzol, line 42: Do we still need the body here?

spirali: We need it because task sharing works now only in the server. So for workers we are sending the body for each task.

Issue #906, opened 2025-05-23
Overly zealous autoalloc


njspix:
Hello,

Thanks for your work on HQ, it's a really nice tool! 

I've noticed that the auto-allocator will happily spin up allocations that it can never use (it only checks if there are jobs waiting, not their requirements). For allocs with a short timeout, this results in incessant alloc thrashing (create worker - no jobs allocated - worker dies due to inactivity - repeat). 

Would it be feasible to add a simple check to see if there are any jobs waiting *which have resources that would fit into the available allocs*? 

Thanks!
v0.22

Kobzol:
Hi, I'm planning to work on this very soon, so I hope to make improvements in this area. The `which have resources that would fit into the available allocs` part is not so easy though, because HQ doesn't know at start what resources are available in an allocation; it's quite difficult and non-portable to ask this of PBS/Slurm, and until the first allocation appears, we don't actually know how will the workers "look like" in terms of HW resources. There are some heuristics that we'll try to apply there though.

njspix:
Oh - that makes sense. I only use pre-specified resources (e.g. 
`hq alloc add slurm --time-limit 2w  --max-worker-count 4 --name big     --cpus  64 --resource  "mem=sum(486000)" --no-detect-resources --idle-timeout 1s -- -p big    -c  64  --mem=500G`). Would you consider adding the check in that case as a first, more simple, step?

(N.B. I use this instead of autodetect because my institution uses Docker to partition physical nodes into multiple virtual nodes, which are what Slurm sees. using autodetect sees the CPUs and memory of the host physical node, not the the virtual slurm node).

Issue #857, opened 2025-03-23
Autoalloc workers are started when some task is waiting for a dependency


lahwaacz:
I have a very simple task graph to reproduce this issue with Python, but it takes some compute time to demonstrate the effects:

```python
import time
import datetime

from hyperqueue import Job, Client
from hyperqueue.ffi.protocol import ResourceRequest


def work(index):
    if index == 0:
        print(f"{index = } is going to sleep")
        time.sleep(300)
        print(f"{index = } is done")
    else:
        print(f"{index = } is done")


client = Client()
job = Job()
resource_request = ResourceRequest(cpus=1, min_time=datetime.timedelta(minutes=10))

# Add a task graph
tasks = {}
for i in range(10):
    # every task depends on the previous one
    deps = []
    if i > 0:
        deps.append(tasks[i - 1])

    tasks[i] = job.function(
        work,
        args=(i,),
        deps=deps,
        resources=resource_request,
    )

submitted = client.submit(job)
client.wait_for_jobs([submitted])
```

To reproduce, first start the HyperQueue server in one terminal:

```
$ hq server start
```

Then in another terminal:

```
$ hq alloc add slurm --idle-timeout 30s --max-worker-count 10 --backlog 10 --time-limit 1h
$ python hq-bug.py  # run the repro script above
```

You can observe the progress in a third terminal (notice that there are always 10 workers running and going idle after 30 seconds, until the first task finishes after 5 minutes):

```
$ hq task list last
+---------+---------+--------+----------+-------+
| Task ID | State   | Worker | Makespan | Error |
+---------+---------+--------+----------+-------+
|       0 | RUNNING | gp11   | 1m 13s   |       |
|       1 | WAITING |        |          |       |
|       2 | WAITING |        |          |       |
|       3 | WAITING |        |          |       |
|       4 | WAITING |        |          |       |
|       5 | WAITING |        |          |       |
|       6 | WAITING |        |          |       |
|       7 | WAITING |        |          |       |
|       8 | WAITING |        |          |       |
|       9 | WAITING |        |          |       |
+---------+---------+--------+----------+-------+

$ hq worker list --all
+----+--------------+----------+--------------------------------------+---------+----------------+
| ID | State        | Hostname | Resources                            | Manager | Manager Job ID |
+----+--------------+----------+--------------------------------------+---------+----------------+
[...]

Issue #243, opened 2021-12-11
Support multi-node tasks


giovannipizzi:
(Note: maybe I'm completely misunderstanding how multiple allocs are supposed to be used - so please tell me if this can be solved in a different way!)

I have two usecases and I didn't understand if they are already supported (or they work already by in some way I don't know)

1. on some of our supercomputers at the Swiss supercomputing centre CSCS, we have two partitions, one with CPU only and one with GPUs. Therefore, I'd like to start two allocs, one for each (e.g. `-C mc` vs. `-C gpu`), give to each of them a name, and when I submit a job to `hq`, specify to which of the two they should run. Is this possible already? ([The docs state that the name is only used for debug reasons](https://it4innovations.github.io/hyperqueue/stable/deployment/allocation/#parameters))

2. If I understand correctly, when I start the alloc, I specify all parameters, including how many nodes each worker should have. Is there a way to start automatically workers with different number of nodes? E.g. if I have a machine with nodes with 128 cores, I might want to have jobs with e.g. 16, 32, 64 or 128 cores run in workers that use 1 node; but I might also have e.g. jobs that I want to run on 2 or 4 nodes, and for these I'd like to also have some workers run on 2 or 4 nodes. At the same time, I don't want the jobs with <= 128 cores to run on these workers but only on the 1-node workers. Can this be already achieved by creating 3 allocs with 1, 2, 4 nodes respectively? I tried to do this but I saw some strange result that I couldn't debug properly or test extensively yet (I think I had only the 1-node alloc first, I submitted a 2-node job to `hq` and I *think* that `hq` started submitting workers, when they entered the SLURM queue it realized they weren't OK for the job and it started submitting more; I then created also a 2-node alloc but it didn't enter the queue for quite some time because of the priorities of the machine, and in the meantime I kept having 1-node workers being submitted, starting, and dying after the 5-minutes idle timeout, and new ones getting launched again).
   For this usecase, I would either have something like the point 1 above (I can decide to send a job only to one specific alloc and avoid that the ones that are not compatible keep starting, timing out and restart), or some automatic way for `hq` to understand how many nodes a job requires, and send to the correct alloc

Kobzol:
1) This is definitely a valid use case and we intend to support it. Currently, this can be achieved in a semi-manual way.

[...]

PR #909, opened 2025-05-26
Rewrite automatic allocator

Kobzol:
This PR rewrites the automatic allocator, so that it leverages scheduler worker queries. It now works across allocation queues, handles multi-node tasks, and primarily makes use of information from the scheduler to determine which allocations to create. This takes both current computational load into demand, worker/task resources and also task dependencies.

Future work includes:
- More accurately represent the resources of each queue (currently, each queue is assumed to have one CPU core)

Fixes: https://github.com/It4innovations/hyperqueue/issues/243
Fixes: https://github.com/It4innovations/hyperqueue/issues/857
Fixes: https://github.com/It4innovations/hyperqueue/issues/906

### DIFF
--- a/tests/autoalloc/test_autoalloc.py
+++ b/tests/autoalloc/test_autoalloc.py
@@ -154,7 +154,7 @@ def test_pbs_multinode_allocation(hq_env: HqEnv):
         start_server_with_quick_refresh(hq_env)
         prepare_tasks(hq_env)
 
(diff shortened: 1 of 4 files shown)


spirali:
LGTM, but as far as I understand, it does not reproduce old behavior in the case of multi-node tasks. If there is only multi-node tasks then it will not spawn any allocation, that is not what old version does (?).

PR #913, opened 2025-05-28
"explain" command

spirali:
This PR adds `hq task explain` that explains why a task is not running on a worker.

It may help with debugging situation like: https://github.com/It4innovations/hyperqueue/issues/784

### DIFF
--- a/crates/hyperqueue/src/client/output/cli.rs
+++ b/crates/hyperqueue/src/client/output/cli.rs
@@ -18,7 +18,7 @@ use crate::transfer::messages::{
     ServerInfo, TaskDescription, TaskKind, TaskKindProgram, WaitForJobsResponse, WorkerExitInfo,
     WorkerInfo,
 };
-use tako::{JobId, JobTaskCount, JobTaskId, WorkerId};
+use tako::{JobId, JobTaskCount, JobTaskId, TaskId, WorkerId};
 
 use chrono::{DateTime, Local, SubsecRound, Utc};
 use core::time::Duration;
@@ -48,6 +48,7 @@ use tako::gateway::{
     LostWorkerReason, ResourceRequest, ResourceRequestEntry, ResourceRequestVariants,
     WorkerRuntimeInfo,
 };
+use tako::server::{TaskExplainItem, TaskExplanation};
 use tako::{Map, format_comma_delimited};
 
 pub const TASK_COLOR_CANCELED: Colorization = Colorization::Magenta;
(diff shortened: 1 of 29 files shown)


crates/hyperqueue/src/client/task.rs
```diff
@@ -53,6 +55,19 @@ pub struct TaskInfoOpts {
     pub verbosity: VerbosityFlag,
 }
 
+#[derive(clap::Parser)]
+pub struct TaskExplainOpts {
+    /// Select specific job
+    #[arg(value_parser = parse_last_single_id)]
+    pub job_selector: SingleIdSelector,
+
+    /// Select specific task(s)
```
Kobzol, line 64: Can you actually select multiple tasks with this?

crates/hyperqueue/src/client/output/cli.rs
```diff
@@ -1044,6 +1045,99 @@ impl Output for CliOutput {
...
     }
+
+    fn print_explanation(
+        &self,
+        task_id: TaskId,
+        worker_id: WorkerId,
+        explanation: &TaskExplanation,
+    ) {
+        let all = explanation.variants.len();
+        let count: usize = explanation
```
Kobzol, line 1056: `count` -> `runnable_variants` (it was hard for me to understand how this works)

crates/hyperqueue/src/client/output/cli.rs
```diff
@@ -1044,6 +1045,99 @@ impl Output for CliOutput {
...
+            .sum();
+        if count > 0 {
+            if all == 1 {
+                println!(
+                    "Task {task_id} {} on worker {worker_id}.",
+                    "can run".color(colored::Color::Green)
+                );
+            } else {
+                println!(
+                    "Task {task_id} {} on worker {worker_id}, ({count}/{all} variants)",
```
Kobzol, line 1069: resource variants

crates/hyperqueue/src/client/output/cli.rs
```diff
@@ -1044,6 +1045,99 @@ impl Output for CliOutput {
...
+                );
+            }
+        } else if all == 1 {
+            println!(
+                "Task {task_id} {} on worker {worker_id}.",
+                "cannot run".color(colored::Color::Red)
+            );
+        } else {
+            println!(
+                "Task {task_id} {} on worker {worker_id}, none of {all} variants.",
```
Kobzol, line 1094: resource variants

Kobzol (commented):
This is incredible, thank you! I thought about how would users use this, and I think that we could make it even better with the following:
- Allow selecting multiple workers, and default to choosing all workers. The query doesn't look very expensive at the moment, so perf. wise it should hopefully be fine. When you have a task that doesn't run, and you have multiple workers, it probably won't be very exciting to go through the workers one by one and trying to figure out why it doesn't run on each one of them. We could present some aggregated statistics in the output (can run on workers 1, 3, 5, cannot run on workers 2, 4, etc.), and then maybe show the resource explanations in a table? One row per worker, one column per type (min time, multi-node, resources, etc.), and for resources do a multi-row cell, or something like that.
- Could we also take into account the dynamic utilization of each worker? The output could say "task 1 can run on worker 2, but not right now, because worker 2 has all GPUs exhausted", or something like that.

Anyway, both could be done in a follow-up PR, this is pretty nice on its own.

crates/hyperqueue/src/client/task.rs
spirali: fixed

Kobzol:
Nice! After this line: `Task 1@0 is ready to run because it has no dependencies.`, it would be also useful if it printed something like "Task can run on workers 1-5,7,9". Because now there is a nice summary at the top about dependencies, but then only a table is printed, that kind requires the user to interpret the situation from the table, without having any summary provided by HQ.

Kobzol:
Looks incredible, thank you! I tried to use `explain` on a task that was already finished, and it returned "Task not found", even though the task does exist in HQ (but probably not anymore in Tasko).

PR #915, opened 2025-05-29
Idle timeout checked in server

spirali:
This PR moves idle check from worker to server.
* It simplifies the code, because reservation does not need to be announced to worker.
* Makes it more robust, because broken worker cannot ignore idle timeout


Kobzol (commented):
I was wondering whether the worker can now stay open indefinitely. In theory, if the worker stops receiving communication from the server, but the TCP/IP connection doesn't close, this could happen? Maybe it would be worth it to keep some sanity check on the worker, i.e. if no message was received in 30 minutes and nothing is being computed, just shut down? But I guess that would break the fact that reservations are only on the server for now. So probably not worth it.

Nice simplification!

Kobzol (approved):
Right, it would be weird if the worker could send stuff to the server, but not receive anything, it should at least got TCP ACKs back.

Issue #919, opened 2025-06-05
test_max_fails_many failing on main branch


lahwaacz:
I've just built [hyperqueue-git](https://aur.archlinux.org/packages/hyperqueue-git) on Arch Linux from https://github.com/It4innovations/hyperqueue/commit/8bba90c4c6c7926c5918345a1ae0afc69361bd67 and the following test consistently fails:

```
$ python -m pytest tests -- tests/test_job.py::test_max_fails_many
========================================================= test session starts ==========================================================
platform linux -- Python 3.13.3, pytest-8.4.0, pluggy-1.6.0
rootdir: /build/hyperqueue-git/src/hyperqueue/tests
configfile: pytest.ini
plugins: xdist-3.6.1, inline-snapshot-0.23.2
collected 1 item

tests/test_job.py FE                                                                                                             [100%]

==================================== ERRORS ====================================
___________________ ERROR at teardown of test_max_fails_many ___________________

tmp_path = PosixPath('/tmp/pytest-of-builduser/pytest-3/test_max_fails_many0')

    @pytest.fixture(autouse=False, scope="function")
    def hq_env(tmp_path):
>       with run_hq_env(tmp_path) as env:
             ^^^^^^^^^^^^^^^^^^^^

/build/hyperqueue-git/src/hyperqueue/tests/conftest.py:334:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
/usr/lib/python3.13/contextlib.py:148: in __exit__
    next(self.gen)
/build/hyperqueue-git/src/hyperqueue/tests/conftest.py:349: in run_hq_env
    env.check_running_processes()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <tests.conftest.HqEnv object at 0x7748c5ce9160>

    def check_running_processes(self):
        """Checks that everything is still running"""
        for p in self.processes:
            if p.final_check and p.process.poll() is not None:
>               raise Exception("Process {0} crashed (log in {1}/{0}.out)".format(p.name, self.work_path))
E               Exception: Process server crashed (log in /tmp/pytest-of-builduser/pytest-3/test_max_fails_many0/server.out)

[...]

PR #920, opened 2025-06-06
Fix task cancellation on max fails

Kobzol:
Found this while investigating https://github.com/It4innovations/hyperqueue/issues/919. I can't reproduce that locally, so I'm not sure if this solves the issue, but the `is_empty()` check was quite suspicious.

An alternative would be to handle `Canceled` state in `set_failed_state`, but I guess that would just hide potential bugs?


Kobzol:
Seems like this is exposing more bugs :cold_face:

spirali (approved):

PR #927, opened 2025-06-13
Expose `min utilization` for autoalloc

Kobzol:
Fixes: https://github.com/It4innovations/hyperqueue/issues/886

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,10 @@ The automatic allocator has been finally reimplemented, and is now much better:
 * It now uses information from the scheduler to determine how many allocations to spawn, and thus it can react to the
   current computational load much more accurately. It should also be less "eager".
 * It properly supports multi-node tasks.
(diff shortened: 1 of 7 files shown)


spirali:
Can we have an end-to-end test that checks that the functionality is working? I made some tests but only on the scheduler level.

Kobzol:
I thought about it, but I think that I first need to implement proper resources for queues. Now I tell the scheduler that every queue has 1 CPU core and nothing else. Not sure if min utilization will work with that? :laughing:

spirali:
Good point.

Btw: The fact that it says that it has 1 cpus will cause the problem: if all tasks needs >1cpu or some other resource then autoalloc will never allocate. So it is not compatible with the previous version in this respect.

spirali (approved):

Kobzol:
Yeah, it's definitely something we need to resolve before publishing a next release. But the previous PR was already quite big, so I'll do it in a follow-up next week.

PR #930, opened 2025-06-13
drop atty crate

xosnrdev:
Figured it's no longer maintained.


Kobzol (approved):
Thanks! I noticed this a few weeks ago, but then forgot about it.

PR #933, opened 2025-06-20
Leftovers update

spirali:
* Leftovers is now array of ResourceRequests. You can decide if it makes sense to make a probe for a queue for these requests (e.g. min_time > time_limit so probe is unnecessary, or n_nodes > max_worker_per_alloc, or if we already know the number of cpus, etc).

* Leftovers is optional (because it is not free to compute now), so if there is no "unknown queues", you can skip computing leftovers in query.

### DIFF
--- a/crates/tako/src/control.rs
+++ b/crates/tako/src/control.rs
@@ -8,7 +8,7 @@ use tokio::net::TcpListener;
 use tokio::sync::Notify;
 
 use crate::events::EventProcessor;
-use crate::gateway::{MultiNodeAllocationResponse, TaskSubmit, WorkerRuntimeInfo};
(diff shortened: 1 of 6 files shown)


spirali:
This version of leftovers already consider multinode tasks.

crates/tako/src/control.rs
```diff
@@ -45,7 +45,7 @@ pub struct NewWorkerAllocationResponse {
     /// be spawned for the given query
     pub single_node_workers_per_query: Vec<usize>,
     /// True iff not all tasks could be executed on resources provided in the query
```
Kobzol, line 47: Please update the comment :)

spirali: Thanks, fixed:)

PR #934, opened 2025-06-20
Estimate resources of autoalloc queues

Kobzol:
Estimate resources of autoallocation queues.

### DIFF
--- a/crates/hyperqueue/src/server/autoalloc/process.rs
+++ b/crates/hyperqueue/src/server/autoalloc/process.rs
@@ -26,9 +26,7 @@ use std::path::PathBuf;
 use std::time::{Instant, SystemTime};
 use tako::WorkerId;
 use tako::control::{ServerRef, WorkerTypeQuery};
-use tako::resources::{
-    CPU_RESOURCE_NAME, ResourceDescriptor, ResourceDescriptorItem, ResourceDescriptorKind,
-};
+use tako::resources::ResourceDescriptor;
 use tako::{Map, Set};
 use tempfile::TempDir;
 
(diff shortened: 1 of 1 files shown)


crates/hyperqueue/src/server/autoalloc/process.rs
```diff
@@ -372,12 +387,68 @@ async fn perform_submits(
...
+            (worker_config.resources.clone(), false)
+        } else if let Some(resources) = queue.info().cli_resource_descriptor() {
+            // If not, try to at least estimate partial resources based from the CLI arguments
+            (resources.clone(), true)
+        } else {
+            // Otherwise just assume that the worker will have at least a single CPU
+            (
+                ResourceDescriptor::new(vec![ResourceDescriptorItem {
+                    name: CPU_RESOURCE_NAME.to_string(),
+                    kind: ResourceDescriptorKind::regular_sockets(1, 1),
```
spirali, line 424: This is wrong logic. It says, "we do not know all resources about worker, but we know that it has 1 cpu" so it will be ignored by tasks that needs more cpus. 
If we really do no know nothing, then it should be an empty vec.
I other words, the partial resources are interpreted as lowebound on what we know about the worker, not lowerbound on resources.

crates/hyperqueue/src/client/commands/autoalloc.rs
```diff
@@ -306,7 +309,12 @@ wasted allocation duration."
             "--cpus".to_string(),
             format!("\"{}\"", cpus.into_original_input()),
         ]);
+    } else {
+        log::warn!(
+            "Creating an autoalloc queue without specifying worker core count. Consider specifying the core count of workers in this queue to improve the efficiency of automatic allocation."
```
spirali, line 314: I would formulate it rather something like:
"No resources are specified for the queue. Autoallocator is running in a probing mode until a first worker
occur and we get the information about worker resources"

spirali (changes requested):

crates/hyperqueue/src/server/autoalloc/process.rs
Kobzol: Good catch! I implemented this while still thinking in the "old regime" where we didn't have the lower bound.

However, it looks like worker query computation doesn't handle an empty vec currently:
```
called `Result::unwrap()` on an `Err` value: cannot compute worker query

Caused by:
    Resource 'cpus' is missing
```
It validates all the resource descriptors at start, and validation doesn't allow missing CPUs.

spirali (approved):

PR #938, opened 2025-06-27
Add support for partial resources to worker query

spirali:
Improved API instead of just collecting leftovers.

### DIFF
--- a/crates/tako/src/control.rs
+++ b/crates/tako/src/control.rs
@@ -8,7 +8,9 @@ use tokio::net::TcpListener;
 use tokio::sync::Notify;
 
 use crate::events::EventProcessor;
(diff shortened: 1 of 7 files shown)


spirali:
Added "partial" to WorkerTypeQuery

crates/tako/src/control.rs
```diff
@@ -39,17 +42,19 @@ pub struct WorkerTypeQuery {
     pub min_utilization: f32,
 }
 
+#[derive(Debug)]
+pub struct WorkerLowerBoundQuery {
```
Kobzol, line 46: Leftover? :)

Kobzol (approved):

PR #939, opened 2025-06-27
Fixed clippy warnings for rust 1.88.0

spirali:


Kobzol (approved):
Heh, finally enabled by default after a year. Needs reformatting, otherwise LGTM.

Issue #937, opened 2025-06-25
HyperQueue not submitting allocations


svatosFZU:
I am using hq-nightly-2025-06-23-3020ab5c8bca3548b2afbc04623bc191c33550bc-linux-x64 to test https://github.com/It4innovations/hyperqueue/issues/886

Since I started testing it yesterday, I have noticed that the HyperQueue is not submitting any batch jobs. It submits one to test when a new allocation queue is made but nothing more, even though there is enough work to fill multiple workers. I tried allocation queue with and without --min-utilization=1 but it makes no difference. 

I see no errors in the debug log: https://www.fzu.cz/~svatosm/hq-debug-output.log

svatosFZU:
Thanks for the info but that does not seem to be the cause. The log above is from Karolina. I have the same situation on Barbora, where I have only single-core jobs now.
debug log: https://www.fzu.cz/~svatosm/hq-debug-output-barbora.log

Kobzol:
Hi, could you please try again now that https://github.com/It4innovations/hyperqueue/pull/934 has been merged? We have extended the logging to make it easier to debug what's going on, and now the queues should estimate resources in a smarter way.

Kobzol:
I see. Yes, the worker has ended because of idle timeout, and it wasn't re-created again, because the automatic allocator figured out that this worked is unable to execute the submitted tasks.

The problem is that the connected worker doesn't have enough memory (`{ name: "mem", kind: Sum { size: 191851.8476 } `), because the task asks for `192000`. Try lowering the `mem` request of the tasks to e.g. `191000`.

svatosFZU:
Thanks, I will try that. So, I was going to cancel these jobs asking with a lot of memory. When I tried that, the HyperQueue crashed. So, this is just FYI:
https://www.fzu.cz/~svatosm/20250701131502-hq-debug-output.log

spirali:
> Thanks, I will try that. So, I was going to cancel these jobs asking with a lot of memory. When I tried that, the HyperQueue crashed. So, this is just FYI: https://www.fzu.cz/~svatosm/20250701131502-hq-debug-output.log

Thank you for the report. #949 should fix this crash.

svatosFZU:
Sure, I think this is a good thing to do. By the way, should I try to use min_allocation again?

Kobzol:
The fixes were merged yesterday, so today's nightly should contain them. You're welcome to try :)

Kobzol:
[...]

PR #949, opened 2025-07-01
Fix crash when query + canceled sleeping tasks

spirali:
This should fix the crash mention in #937.

### DIFF
--- a/crates/tako/src/internal/server/core.rs
+++ b/crates/tako/src/internal/server/core.rs
@@ -194,11 +194,13 @@ impl Core {
     pub fn new_worker(&mut self, worker: Worker) {
         /* Wake up sleeping tasks */
         let mut sleeping_sn_tasks = self.take_sleeping_tasks();
-        self.single_node_ready_to_assign
-            .append(&mut sleeping_sn_tasks);
-
+        if self.single_node_ready_to_assign.is_empty() {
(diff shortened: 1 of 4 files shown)


crates/tako/src/internal/server/core.rs
```diff
@@ -194,11 +194,13 @@ impl Core {
     pub fn new_worker(&mut self, worker: Worker) {
         /* Wake up sleeping tasks */
         let mut sleeping_sn_tasks = self.take_sleeping_tasks();
-        self.single_node_ready_to_assign
-            .append(&mut sleeping_sn_tasks);
-
+        if self.single_node_ready_to_assign.is_empty() {
```
Kobzol, line 197: Is this always better? If `single_node_ready_to_assign` is long running (but often empty), it could accumulate capacity, while here we would reset the capacity every time.

spirali: `single_node_ready_to_assign` is not accumulating capacity because scheduler calls `take_single_node_ready_to_assign` and consumes whole vec. I will look it, but the idea was to free potentially large vec that is not not needed for most of the lifetime of the server (usually only at the beginning).

Kobzol: Ah, I didn't realize that scheduler takes the vec. Ok, then it should be fine.

PR #956, opened 2025-07-03
Improve documentation

Kobzol:
- Document `hq doc`
- Document `hq generate-completion`
- Fix a few typos
- Improve, rewrite and unify a few sections of the documentation
- Replace the `mkdocs-nedoc-plugin` with a Python script hook, change the link prefix to be `pyapi:`
- Add a postprocessing hook script to process `cli:` prefixed Markdown links so that they point to documentation generated by `cli_doc`
- Add CLI links to all commands mentioned in the documentation (!)


crates/hyperqueue/src/common/cli.rs
```diff
@@ -191,7 +191,7 @@ pub enum SubCommand {
     Job(JobOpts),
     /// Commands for tasks
     Task(TaskOpts),
-    /// Submits a new job
+    /// Submit a new job
```
spirali, line 194: We have all cli doc verbs in "s" form.

scripts/docs/postprocess_nedoc.py
```diff
@@ -0,0 +1,109 @@
...
+It also allows you to use links to API documentation of a Python package documented by Nedoc.
+
+Use a link that starts with `pyapi:`, followed by a key contained in `map.json`.
+`map.json` is generated by Nedoc if you use the `create_map_json = True` configuration in `nedoc.conf`.
+
+Links cannot span a line! Keep them on a single line.
+Links that start with # will not be considered.
+
+Example:
+The [Box](pyapi:elsie.box.Box) class is awesome.
```
spirali, line 12: Does this the same think as: https://github.com/spirali/mkdocs-pdoc-plugin?:)

crates/hyperqueue/src/common/cli.rs
Kobzol: [Not all](https://github.com/it4innovations/hyperqueue/blob/146463615121481aa6d5eab5a26c4899e454a989/crates/hyperqueue/src/common/cli.rs#L376) :laughing: But it looks like I updated it in the wrong "direction", thanks.

spirali (approved):
Good job!

PR #968, opened 2025-07-22
Fix downloading test on machines with long domain resolution

### LINKED ISSUE #963: download_invalid_hostname test fails
When building the [Arch Linux package](https://gitlab.archlinux.org/archlinux/packaging/packages/hyperqueue) for 0.23.0, there is one test that consistently fails:

```
failures:

---- internal::datasrv::tests::download_invalid_hostname stdout ----

thread 'internal::datasrv::tests::download_invalid_hostname' panicked at crates/tako/src/internal/datasrv/test_utils.rs:81:9:
assertion `left == right` failed
  left: 0
 right: 1


failures:
    internal::datasrv::tests::download_invalid_hostname

test result: FAILED. 201 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 7.00s

error: test failed, to rerun pass `-p tako --lib`
```

This is from `cargo test --frozen --release --all-features`. Any idea what is wrong?

spirali:
Fixes #963

Instead of resolving invalid hostname, we will try to connect to port 1 on localhost that should also fails. 
In Tako code, this error has the same code path.

### DIFF
--- a/crates/tako/src/internal/datasrv/tests.rs
+++ b/crates/tako/src/internal/datasrv/tests.rs
@@ -47,16 +47,16 @@ fn download_update_priorities() {
 }
 
 #[tokio::test]
-async fn download_invalid_hostname() {
+async fn download_invalid_host() {
     let set = tokio::task::LocalSet::new();
     set.run_until(async move {
         let (dm_ref, interface) = test_download_manager();
 
         let data_id1 = DataObjectId::new(1.into(), 5.into());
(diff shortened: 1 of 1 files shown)


spirali:
> Maybe 2 would be better? :D https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers In any case, feel free to merge.

Thanks. I have changed it to port 0.

PR #973, opened 2025-07-23
Added missing typos.sh

spirali:


lahwaacz:
You can configure what to lint or not to lint in `_typos.toml`, see https://github.com/crate-ci/typos/blob/master/docs/reference.md

And if you add the `venv` etc. directories to `.gitignore`, typos should also ignore it by default.

PR #977, opened 2025-07-24
Restore known worker resources for allocation queues

Kobzol:
So that they are not forgotten when the server restarts.

Closes: https://github.com/It4innovations/hyperqueue/issues/958


spirali (commented):
Looks good.

I would be nice that allocation can print some info what it assumes that resources of workers are (or if it in a probing mode). Then we can have an integration test that we have loaded this info from journal.

Kobzol:
Created an issue for that: https://github.com/It4innovations/hyperqueue/issues/984.

Issue #984, opened 2025-07-25
Output known worker configuration for autoalloc queues


Kobzol:
https://github.com/It4innovations/hyperqueue/pull/977#pullrequestreview-3054879261

PR #986, opened 2025-08-01
Show known worker resources in alloc list output

Kobzol:
Closes: https://github.com/It4innovations/hyperqueue/issues/984


spirali:
Now we can have a better test if resource queue was loaded from journal :)

spirali (approved):

PR #1014, opened 2025-08-29
Advanced coupling

spirali:
This PR introduces advance coupling that allows to set individual weights between groups.

Important notes:
* New dependency on HiGHS (MLP solver), that is written in C++. It needs cmake & gcc to be installed for build (however everything is still buildable via cargo, no manual installation of HiGHS is needed). The final library is then statically linked to HQ. 
   - In case of build problems, it it quite easy to switch to minilp (pure Rust solver) but the performance differences for more complex scenarios is much worse, so for now, the implementation has only HiGHS support.
* All computations related to resource groups are now solved by the MLP solver so even if the user is not using coupling, this PR may have different performance then the previous code. As we automatically detect CPU groups, probably every run on a bigger cluster (with NUMA compute nodes) will be hit by this change. So it is a good idea to benchmark this change on real machines. 
* In case of build problems on small machines like Raspberry PI, we migrate all "group" support behind a feature flag. It should be ok as all Raspberry PIs is probably without NUMA.
* This is PR not planned for v0.24.0.
* Cargo fix moves some codes to the newest Rust ...

### DIFF
--- a/crates/hyperqueue/src/client/commands/worker.rs
+++ b/crates/hyperqueue/src/client/commands/worker.rs
@@ -105,8 +105,15 @@ pub struct SharedWorkerStartOpts {
 
     /// Resource coupling
     ///
+    /// Coupling is a way to specify that some resource groups have an affinity
+    /// It is used for `compact` and `tight` allocation policy
+    ///
+    /// Syntax: Comma separated list of: `<RESOURCE1>[<GROUP1>]:<RESOURCE2>[GROUP2]` or
+    /// `<RESOURCE1>[<GROUP1>]:<RESOURCE2>[GROUP2]=<WEIGHT>`.
+    ///
     /// Examples:{n}
-    /// - `--coupling TODO"
+    /// - `--coupling=cpus[0]:foo[0],cpus[1]:foo[1]`
+    /// - `--coupling=cpus[0]:foo[0]=128`
     #[arg(long)]
     pub coupling: Option<String>,
 


crates/hyperqueue/src/client/commands/worker.rs
```diff
@@ -107,9 +106,9 @@ pub struct SharedWorkerStartOpts {
     /// Resource coupling
     ///
     /// Examples:{n}
-    /// - `--coupling cpus,gpus"
-    #[arg(long, value_parser = passthrough_parser(parse_resource_coupling))]
-    pub coupling: Option<PassThroughArgument<ResourceDescriptorCoupling>>,
+    /// - `--coupling TODO"
```
Kobzol, line 109: The TODO is intended to be here?

crates/hyperqueue/src/worker/parser.rs
```diff
@@ -208,12 +206,61 @@ pub fn parse_resource_definition(input: &str) -> anyhow::Result<ResourceDescript
...
-    let names = input.split(",").map(|s| s.trim().to_string()).collect_vec();
-    if names.len() < 2 {
-        bail!("Resource coupling needs at least two resource names separated by commas");
-    }
-    Ok(ResourceDescriptorCoupling { names })
+fn parse_resource_coupling_group(
+    input: &str,
+    resources: &[ResourceDescriptorItem],
+) -> anyhow::Result<(u8, ResourceGroupIdx)> {
+    let input = input.trim();
```
Kobzol, line 213: You gave up on the parser combinators? 😆 It's easier to have a high quality error message when parsing "manually", that's for sure...

spirali: I am not giving up completely, but if I remember correctly, in this case, it was quite ugly code and manual parser was easy.

PR #1024, opened 2025-09-22
Waiting mechanism refactor + user notification mechanism.

spirali:
This PR contains two things:

## Rewritten system on task waiting
  * It unifies plain waiting for a job and waiting for a job with progress bar.
  * Previously there was server side call for plain waiting job and wait with progress bar was periodically checked the state of the server (with period of 1s).
  * The new system utilize event system. It adds option to filter messages on type and job id. So waiting client should subscribe for relevant events (in plain waiting just top-level job events; in the case of progressbar waiting also task events). The new system does  not need any periodic queries into the server.
  * Old plain job waiting system was completely removed.
 
## Notification

  * Task are now allowed to sent notifications. Notifications can be consumed by client. Notifications is never persisted. If there is no listening end, they are just discarded.
  * On "task" level it utilizes "local connection" that is already there for data layer (and will be used for subworkers).
  * On received part, it just utilize event system.


Kobzol:
> Previously it returned 1 when at least one waited task failed or was canceled.

This was actually on purpose, so that you can wait for a job to end, and you immediately get an information whether the job failed or not. (Not sure if it properly failed if all tasks finished *after* the wait was started were successful, but there was a previous task failure though)

crates/hyperqueue/src/client/commands/journal/mod.rs
```diff
@@ -60,20 +60,27 @@ struct ExportOpts {
 pub async fn command_journal(gsettings: &GlobalSettings, opts: JournalOpts) -> anyhow::Result<()> {
     match opts.command {
         JournalCommand::Export(opts) => export_json(opts),
-        JournalCommand::Replay => stream_json(gsettings, false).await,
-        JournalCommand::Stream => stream_json(gsettings, true).await,
+        JournalCommand::Replay => stream_json(gsettings, true, false).await,
+        JournalCommand::Stream => stream_json(gsettings, true, true).await,
```
Kobzol, line 65: What happens if you select live events false and past events false? This would be better as an enum maybe?

crates/hyperqueue/src/server/client/mod.rs
```diff
@@ -477,44 +499,53 @@ async fn handle_wait_for_jobs_message(
...
-
-    for result in results {
-        match result {
-            Ok(job_id) => {
-                match state.get_job(job_id) {
-                    Some(job) => update_counters(&mut response, &job.counters),
-                    None => continue,
-                };
+    if !remaining.is_empty() {
+        let mut task_set = Set::from_iter(remaining.into_iter());
```
Kobzol: This is job IDs, why is this called `task_set`? That's confusing :)

crates/hyperqueue/src/server/client/mod.rs
```diff
@@ -477,44 +499,53 @@ async fn handle_wait_for_jobs_message(
...
+        let filter = EventFilter::new(Some(task_set.clone()), false);
+        let (sender, mut receiver) = mpsc::unbounded_channel();
+        senders.events.register_listener(filter, sender);
+        loop {
```
Kobzol: `while !task_set.is_empty()`

crates/hyperqueue/src/server/client/mod.rs
```diff
@@ -212,8 +217,21 @@ pub async fn client_rpc_loop<
         match message_result {
             Ok(message) => {
                 let response = match message {
-                    FromClientMessage::Submit(msg) => {
-                        submit::handle_submit(&state_ref, senders, msg)
+                    FromClientMessage::Submit(msg, stream_opts) => {
+                        let response = submit::handle_submit(&state_ref, senders, msg);
+                        if let Some(stream_opts) = stream_opts {
```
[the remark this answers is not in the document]
spirali: We can lost an arbitrary number of events between processing two client messages, it would for example break the "hq submit --wait / --progress".

crates/hyperqueue/src/common/cli.rs
```diff
@@ -400,6 +400,10 @@ pub struct JobWaitOpts {
     /// Waits until all tasks are completed, even if the job is still open
     #[clap(long, action)]
     pub without_close: bool,
+
+    /// Call the program in the argument that is executed whenever a task produce a notification
+    #[clap(long, action)]
```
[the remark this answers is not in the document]
spirali: This is a forgotten leftover that is not used. I am removing it.

crates/hyperqueue/src/server/client/mod.rs
spirali: Code is removed in the final version.

crates/hyperqueue/src/server/client/mod.rs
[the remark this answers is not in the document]
Kobzol: Right, why does it matter if we lose them? Shouldn't `submit + sleep 1 + wait` have the same outcome as `submit --wait`? Or is this purely because of task notifications that could be lost?

PR #1026, opened 2025-09-26
Initial implementation of HTML report from journal

spirali:
This PR adds `hq journal report <journal> <output-filename>` that exports journal statistics as follows:

image

* This implementation now takes the journal file as the input. In the case of need it is possible to create a report from a live server or JSON export.
* This version is *not* rebased on the newest main, because #1024 adds new events and v0.23.0-v0.24.0 journal would be not loadable. I am leaving it as it is for now so potential testers may use their current journals. (btw: It it non conflicting with the current main)
* I have just implemented basic charts and basic time filtering. There are a lots of option what to implement but I would let it be user use-case driven.
* Known bug: It does not work when resource variants are used (needs to update journal to implement this; it will be in the next PR).

### DIFF
--- a/crates/hyperqueue/src/client/commands/journal/report.rs
+++ b/crates/hyperqueue/src/client/commands/journal/report.rs
@@ -14,7 +14,7 @@ use std::path::PathBuf;
 use tako::gateway::{ResourceRequest, ResourceRequestVariants};
 use tako::resources::ResourceAmount;
 use tako::worker::WorkerConfiguration;
-use tako::{JobId, JobTaskId, TaskId, WorkerId};
+use tako::{JobId, JobTaskId, ResourceVariantId, TaskId, WorkerId};
 
 #[derive(Parser)]
 pub(crate) struct JournalReportOpts {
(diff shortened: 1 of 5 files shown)


Kobzol:
> This version is not rebased on the newest main

CI will always rebase on `main`, so you won't be able to share the CI artifacts with the main that this PR is based on.

crates/hyperqueue/src/client/commands/journal/report.rs
```diff
@@ -0,0 +1,832 @@
...
+    ///
+    /// Counted from the start of the server (e.g. "1h")
+    #[arg(long)]
+    start_time: Option<String>,
+
+    /// Show only statistics upto a given time
+    ///
+    /// Counted from the start of the server (e.g. "1h")
+    #[arg(long)]
+    end_time: Option<String>,
```
Kobzol, line 39: We can use this:
```rust
#[arg(
        long,
        value_parser = parse_hms_or_human_time,
        help = duration_doc!("Duration after which will an idle worker automatically stop")
    )]
    pub idle_timeout: Option<Duration>,
```
for parsing human durations.

crates/hyperqueue/src/client/commands/journal/report.rs
```diff
@@ -0,0 +1,832 @@
...
+                        jstats.running_tasks.remove(&task_id)
+                    {
+                        jstats.task_start_stop(
+                            time,
+                            task_id,
+                            rv_id,
+                            &workers,
+                            TaskStartStop::Stop {
+                                start_time: task_start_time,
+                                fail: false,
```
Kobzol, line 279: Shouldn't this be `fail: true`?

spirali: Fixed. Good catch:)

crates/hyperqueue/src/client/commands/journal/report.rs
spirali: Fixed

PR #1055, opened 2026-01-28
Tests refactoring

spirali:
Refactoring of Tako tests.

* Tests unified over TestEnv, with few execeptions that test lowlevel stuff, all tests should construct core via TestEnv.
* Removed some pre-TestEnv helper functions for populating core
* TaskBuilder does not take ID; ID is attached when task is inserted into core. It allows to reuse task definition.
* WorkerBuilder introduced; ad-hoc worker configuration creation replaced by this builder

### DIFF
--- a/crates/tako/src/internal/server/explain.rs
+++ b/crates/tako/src/internal/server/explain.rs
@@ -144,79 +144,72 @@ pub fn task_explain_for_worker(
 
 #[cfg(test)]
 mod tests {
-    use crate::internal::common::resources::map::GlobalResourceMapping;
+
     use crate::internal::server::explain::{TaskExplainItem, task_explain_for_worker};
-    use crate::internal::server::worker::Worker;
+
     use crate::internal::server::workergroup::WorkerGroup;
-    use crate::internal::tests::utils::schedule::create_test_worker_config;
     use crate::internal::tests::utils::task::TaskBuilder;
-    use crate::resources::{
-        ResourceAmount, ResourceDescriptor, ResourceDescriptorItem, ResourceIdMap,
-    };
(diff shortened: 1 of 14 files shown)


crates/tako/src/internal/server/explain.rs
```diff
@@ -228,11 +228,11 @@ mod tests {
         ));
         assert_eq!(r.n_enabled_variants(), 0);
 
-        let task = TaskBuilder::new(task_id)
+        let task = TaskBuilder::new()
```
Kobzol: It would be better if there was some test context from which you create the task builder, and it assigns the task id automatically, rather than choosing the ID manually in tests :)

crates/tako/src/internal/server/task.rs
```diff
@@ -521,29 +522,27 @@ mod tests {
...
-        let b = task_with_deps(1, &[&a], rmap);
-        let c = task_with_deps(2, &[&b], rmap);
-        let d = task_with_deps(3, &[&b], rmap);
-        let e = task_with_deps(4, &[&c, &d], rmap);
-
-        let expected_ids = vec![b.id, c.id, d.id, e.id];
-        submit_test_tasks(&mut core, vec![a, b, c, d, e]);
-
+        let mut rt = TestEnv::new();
+        let a = rt.new_task_default(0);
```
Kobzol: 😆 I feel like you were reading my slides.

crates/tako/src/internal/server/explain.rs
spirali: I plan this but in the next iteration during integrating of scheduler-v2, because lots of current scheduler tests will be written differently.

PR #1088, opened 2026-04-02
In heartbeat handle, proactively check for missed min_time

spirali:
Optimization where worker proactively retracts prefilled tasks that missed min_time.
This speeds up to returning the task non-prefilled state and makes the scheduling more deterministic.


Kobzol:
Why is this done in the heartbeat handle? Seems a bit weird. Shouldn't this be a separate periodic background check?

spirali:
> Why is this done in the heartbeat handle? Seems a bit weird. Shouldn't this be a separate periodic background check?

It should be, I just did not want to introduce a new config option :see_no_evil:.

Kobzol:
This is an internal optimization, I don't think we have to make it configurable.

spirali:
It should be configurable for tests; otherwise, tests would be too long & dependent on an internal constant.

spirali:
Removed from heartbeat; now it has a stand-alone loop & configuration.

Kobzol (approved):

Issue #1001, opened 2025-08-15
Add cancellation reason to `hq job cancel`


Kobzol:
> Is it currently possible to add a reason when cancelling a job? Something like hq job cancel 42 --reason="Accidentally used wrong binary" or hq job cancel 43 --reason="submitted to wrong alloc"?
If not, I think it would be a great addition, as it would allow one to retrospectively do hq job list --all and be reminded of why one cancelled these jobs and resubmitted them.

Asked on [Zulip](https://hyperqueue.zulipchat.com/#narrow/channel/325544-general/topic/Give.20reason.20for.20cancelling.20a.20job/near/534657253) by @StHagel.

It think it would be a nice addition, WDYT @spirali?

spirali:
I agree. We already have a state with an attached message (error state) so it is not a difficult to extend this.

spirali:
This dichotomy comes from the fact that cancel has now two origins:
1) Cancel from user; it always cancel the whole job, user cannot cancel just an individual task.
2) When a task fails, the server cancel all tasks that depends on this task (so just a subset of job tasks may be canceled).

In long term a would like to rename the second situation as "abort"; so "cancel" always come from the user and "abort" come from the reaction on a situation.

So in this case, I think that we should use your solution No. 2, so "cancel" as a "whole job action"; Later (when abort is introduced) we rename TaskCancel event to TaskAbort. But for now we can live with having both JobCancel and TaskCancel.

PR #1089, opened 2026-04-13
Add User Cancel Event to Journal, Server cancel replaced with Abort

Dym03:
### I would like to hear your opinion if the colors suggested are okay with you 
``` create_dataset(&aborted, "Aborted", Color::LightCyan) ``` and 
``` DashboardTaskState::Aborted => Color::LightCyan ```
Example below: 
image

### Sorry for not dividing it into a smaller chunk, like only the addition of the JobCancel, but while I was at it, I just took a shot at doing it at once :P.

### DIFF
--- a/docs/jobs/jobs.md
+++ b/docs/jobs/jobs.md
@@ -231,20 +231,19 @@ $ hq job info <job-id>
 
 ### Task state
 
-Each task starts in the `Waiting` state and can end up in one of the terminal states: `Finished`, `Failed`
-or `Canceled`.
+Each task starts in the `Waiting` state and can end up in one of the terminal states: `Finished`, `Failed`, `Canceled` or `Aborted`.
 
 ```
-Waiting-----------------\
-   | ^                  |
-   | |                  |
-   v |                  |
-Running-----------------|
-   | |                  |
-   | \--------\         |
-   |          |         |
-   v          v         v
-Finished    Failed   Canceled
+Waiting----------------------\
+   | ^                       |
+   | |                       |
+   v |                       |
+Running----------------------|
+   | |                       |
+   | \--------\              |
+   |          |              |
+   v          v              v
+Finished    Failed   Canceled/Aborted
 ```
 
 - **Waiting** The task was submitted and is now waiting to be executed.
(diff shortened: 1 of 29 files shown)

### CODE REVIEW

**Comment by Dym03** [comment]
Updated test so it matches the new aborted state.

**Comment by spirali** [comment]
LGTM. 
Random thoughs:
* Maybe color should be more similar to "Failed" as it something happen from runtime not from a user command.
* We can think about (not in this PR) if not to use Abort also after reaching crash counter (it now sets a Failed state, but technically it is not that the fail of the task itself), but it would be nice to distinguish between "abort-after-dep" and "abort-after-crash".

**Comment by Dym03** [comment]
Okey, will look at the colorway. And update the documentation. Also there are some minor bugs of Abort, that will be fixed.
About the crash counter, that as you said could be changed in another PR.

**Comment by Dym03** [comment]
Actually finding all the places that this would affect took me a while. But god bless tests, that forced me to do it. Hopefully I have managed to discover all the places.

Example of the colorway in dashboard:
image

File: docs/jobs/jobs.md [hunk 1]
```diff
@@ -231,30 +231,30 @@ $ hq job info <job-id>
...
+   | ^                  |           |
+   | |                  |           |
+   v |                  |           |
+Running-----------------|-----------|
+   | |                  |           |
+   | \--------\         |           |
+   |          |         |           |
+   v          v         v           v
+Finished    Failed   Canceled    Aborted
 ```
```
spirali, line 247: Change just to:

```
Waiting--------------------\
   | ^                     |
   | |                     |
   v |                     |
Running--------------------|
   | |                     |
   | \--------\            |
   |          |            |
   v          v            v
Finished    Failed   Canceled/Aborted
```

spirali (commented):
I like the final colors.

I have just put one change request to the documentation; otherwise it is ready to merge.

Dym03:
Documentation updated, I think with this merged the #1001 can be closed. And maybe new Issue can be opened to capture the thought about the distuguish of the Abort due to Dependency/ Abort due to crash of worker | due to max fails. I have found 2 more things that needs an update, so I will update them later today.

Dym03:
Updated documentation about the addition of the Aborted state also in the Job, to keep the original idea of propagating the task state to the Job state. But it is also mentioned in the documentation that this shouldn't happen as aborted state is closely connected to the failure of the task, which would imply that this Job State shouldn't be reached, in the current state.
But I think that it should be added, as if there ever in the future there could be abortion of the task without the need of a failure of a task, I think this should be ready to go :).

spirali:
Thanks! 

I have just updated changelog (mainly because it should be moved into the "unreleased" section)

PR #1105, opened 2026-05-16
Add scrolling to cpu util table

Dym03:
For bigger nodes with higher number of cpus, there might an overflow in the cpu util table. Scrolling through it handles it. 
Current implementation has 2 flaws, but i think that handling them is not worth it, but let me know, the first one can be handled easily, second one could, but it wouldn't be clean I think.
Flaws:
- 1 Wrapping of the scroll is handeled only in the way upward. Wrapping of downward selection would require storing the number of visible_cpus or rewriting the table to use the ```StatefulTable``` but i wasn't fond of this solution by the first glance. 
- 2 Right now scrolling is enabled even if it isn't needed. Solving this would require calculating how many cpus can be viewed on the screen.


spirali (approved):
LGTM. Both are just minor things. Thank you!

PR #1111, opened 2026-05-25
Fixed Cargo.lock

### LINKED ISSUE #1110: Cargo.lock is out of sync in release 0.26.1
`Cargo.lock` in release 0.26.1 still has `version = "0.26.0"`. Hence, using the `--locked` flag fails:

```
$ cargo fetch --locked --target host-tuple
    Updating crates.io index
error: cannot update the lock file /build/hyperqueue/src/hyperqueue-0.26.1/Cargo.lock because --locked was passed to prevent this
help: to generate the lock file without accessing the network, remove the --locked flag and use --offline instead.
```

spirali:
Fixes #1110

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -60,7 +60,7 @@ jobs:
       - name: Compile
         env:
           HQ_BUILD_VERSION: ${{ needs.set-env.outputs.version }}
-        run: cargo build --profile dist
+        run: cargo build --profile dist --locked
 
       - name: Compress debuginfo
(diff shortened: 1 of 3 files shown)


Kobzol:
CI should be using `cargo build --locked` somewhere to ensure that this doesn't happen again.

spirali:
@Kobzol Can you please check that I have put `--locked` in the right places?

Kobzol (approved):
Having it in one place would be enough 😆 But this is fine, thanks.

PR #1117, opened 2026-07-03
worker: expose --heartbeat through hq alloc add

MauriceDHanisch:
## Summary
- `hq worker start` already accepted `--heartbeat`, but `hq alloc add slurm`/`hq alloc add pbs` had no way to set it — the field lived on `WorkerStartOpts`, not the `SharedWorkerStartOpts` that autoalloc's generated worker command reuses.
- Moves `heartbeat` onto `SharedWorkerStartOpts` (`Option<Duration>`, no default so we can detect "not set") and forwards it into the generated `hq worker start ...` invocation in `autoalloc.rs` only when explicitly provided, preserving the existing 8s default otherwise.

## Motivation
On a busy server (e.g. mid MILP scheduler solve, or handling a burst of other workers registering), the single-threaded executor can go long enough without polling that an already-connected worker's heartbeat looks lost even though nothing is actually wrong. Autoalloc-managed workers had no way to raise their heartbeat interval to ride this out, since the flag was never wired through `hq alloc add`. Plain `hq worker start` was unaffected — this only closes the gap for the autoalloc path.

## Test plan
- [x] `cargo check -p hyperqueue -p tako`
- [x] `cargo test -p tako` (no regressions)
- [x] Manually verified `hq alloc add slurm --heartbeat 45s` renders `--heartbeat "45s"` in the generated sbatch worker command

### DIFF
--- a/crates/hyperqueue/src/client/commands/worker.rs
+++ b/crates/hyperqueue/src/client/commands/worker.rs
@@ -49,6 +49,10 @@ use crate::worker::parser::{
 use crate::{DEFAULT_WORKER_GROUP_NAME, rpc_call};
 use tako::WorkerId;
 
+/// Default worker heartbeat interval, applied when `--heartbeat` is not passed
+/// (both for `hq worker start` and for `hq alloc add` pilot workers).
+const DEFAULT_HEARTBEAT_INTERVAL: Duration = Duration::from_secs(8);
+
 #[derive(clap::ValueEnum, Clone)]
 pub enum WorkerFilter {
     Running,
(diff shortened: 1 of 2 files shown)


crates/hyperqueue/src/client/commands/worker.rs
```diff
@@ -137,6 +141,13 @@ pub struct SharedWorkerStartOpts {
     )]
     pub idle_timeout: Option<Duration>,
 
+    #[arg(
+        long,
+        value_parser = parse_hms_or_human_time,
+        help = duration_doc!("How often heartbeats are sent\n\nHeartbeats are used to detect worker's liveness. If the worker does not send a heartbeat for given time, then the worker is considered as lost.\n\nDefaults to 8s.")
+    )]
+    pub heartbeat: Option<Duration>,
```
spirali, line 149: why is there Option here? Why not `pub heartbeat: Duration` (+ default_value = "8s") as it was before?

MauriceDHanisch: It is mirroring idle_timeout above. Probably minor, but this way the default is not baked in at queue creation time instead of letting the worker apply it's own default at start time

spirali: Ok, I just do not like that default is written explicitly in the help description. It is quite non obvious that when you change the constant, you need to change also the help.

PR #1118, opened 2026-07-03
worker: retry the connect+auth+registration handshake on transient stalls

MauriceDHanisch:
## Summary
- Extracts the worker's connect -> authenticate -> send registration -> await response sequence into `connect_and_register`, and wraps it in `connect_and_register_with_retry` (8 attempts, 10s backoff between attempts).
- Adds `DsError::AuthenticationRejected`, a distinct, non-retryable error variant for deterministic auth rejections (wrong secret key, incompatible protocol/role, bad challenge, etc.), and updates `finish_authentication`'s rejection sites in `auth.rs` to construct it explicitly instead of falling through the generic `GenericError` catch-all. The retry loop treats this variant as immediately fatal; every other error is retried.
- Along the way, replaces two hard `panic!`s in the old inline registration-response-wait code with proper `Err(...)` returns.

## Motivation
The raw TCP connect step already retried (20x/2s via `connect_to_server`), but nothing downstream of it did. On a busy server — mid MILP scheduler solve on its single-threaded executor, or handling a burst of other workers registering concurrently — any leg of the auth handshake or the registration round-trip could miss its own hardcoded 15s timeout simply because the server didn't get scheduled in time to answer, not because anything was actually wrong. That killed the whole `hq worker start` process outright. On an autoalloc-managed SLURM pilot, that means the entire allocation is wasted, not just one delayed worker.

Reproduced live with a combined registration-burst + real scheduler-load stress test: previously a second batch of workers was left permanently unable to register; with this change all of them (tested up to 30 workers) connect and register cleanly.

Client-side CLI RPCs (`Connection::init` in `connection.rs`) are a separate caller of `do_authentication` and are intentionally untouched — those should keep failing fast rather than retrying.

## Test plan
- [x] `cargo check -p hyperqueue -p tako`
- [x] `cargo test -p tako` — 205 passed, 0 failed, ~5s (confirmed the three deterministic-auth-rejection tests in `test_secret.rs` stay fast — they hit the new non-retryable `AuthenticationRejected` path instead of looping through all 8 retry attempts)
- [x] Live stress repro: registration burst + real scheduler load, up to 30 workers, zero registration failures

### DIFF
--- a/crates/tako/src/internal/common/error.rs
+++ b/crates/tako/src/internal/common/error.rs
@@ -11,6 +11,12 @@ pub enum DsError {
     SchedulerError(String),
     #[error("{0}")]
     GenericError(String),
+    /// The peer deterministically rejected the authentication handshake (wrong
+    /// secret key, incompatible protocol/role). Unlike a timed-out or dropped
+    /// connection, retrying reproduces the exact same rejection every time, so
+    /// callers should treat this as immediately fatal rather than retryable.
+    #[error("{0}")]
+    AuthenticationRejected(String),
 }
 
 impl From<serde_json::error::Error> for DsError {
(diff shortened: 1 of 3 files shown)


spirali:
Thank you.

